PR #3, opened 2016-05-15
rates in dps for iNav > 1.1.0, see iNavFlight#204

DzikuVx:
PID Tuning shows rates in dps in compatible versions of iNav
See iNavFlight#204

### DIFF
--- a/js/fc.js
+++ b/js/fc.js
@@ -37,6 +37,13 @@ var FAILSAFE_CONFIG;
 var RXFAIL_CONFIG;
 
 var FC = {
+    isRatesInDps: function () {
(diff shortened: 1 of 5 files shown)


js/fc.js
```diff
@@ -37,6 +37,13 @@ var FAILSAFE_CONFIG;
 var RXFAIL_CONFIG;
 
 var FC = {
+    isRatesInDps: function () {
+        if (typeof CONFIG != "undefined" && CONFIG.flightControllerIdentifier == "INAV" && semver.gt(CONFIG.flightControllerVersion, "1.1.0")) {
```
digitalentity: No need to check for INAV really, Configurator won't connect to anything but INAV so we can safely remove most of compatibility stuff from Cleanflight.

PR #5, opened 2016-05-17
change firmware location from cleanflight to inavflight

straga:
I changed the location for the firmware files. Now you can flash directly in the configurator.


DzikuVx:
Super nice, thanks

Issue #6, opened 2016-05-19
Add 25ms delay after each line when restoring a dump in cli


theArchLadder:
Multiple users have reported issues with restoring a dump in the CLI, flight modes are sometimes not restored, or only partially. Random settings are not set, and so on. See this issue:
https://github.com/iNavFlight/inav/issues/229

@stronnag wrote

> This is a well known feature of all CF firmware variants, the firmware cannot process mode changes fast enough, the buffer overflows, data is lost. This is why the mwptools cf-cfi (cli backup / restore tool) uses a baud rate and context aware inter-line delay. Slower than pasting into a terminal, but reliable.

He then suggested that a 25ms delay after each new line could help.
To all javascript gurus: How can this be implement so we can try it and see if that solves the problem?

theArchLadder:
Well, that was simple

https://github.com/iNavFlight/inav-configurator/blob/master/tabs/cli.js#L102
Looks like changing the 5 to whatever milliseconds you want does the trick
At least it looks slow in the CLI, so i assume it works. I guess we just need to try and restore dumps now!

@oleost @stronnag Play around with it and see if it works for you guys, i will do the same tomorrow.
I think you have to go into Chromium->settings menu->more tools->extensions and then click "reload" on iNav configurator after the change has been made...

oleost:
100ms is added. This could be closed.

PR #7, opened 2016-05-20
increased timeout for CLI commands

DzikuVx:
As @theArchLadder found and suggested, this would fix iNavFlight/inav#229 and #6 

@oleost give it a try


theArchLadder:
I think this needs some testing, perhaps 10ms is enough?

Looks like a dump is almost 400 lines long, at 25ms that would take 10 seconds to restore, that sound okay with me, in fact, i think it would be better to be a little slow because then one would have a chance to see if there was any errors with any settings...

DzikuVx:
Let's confirm 25 works then we can go lower. I can try test it next week perhaps

theArchLadder:
I will also try and test this when i have some time over. As you said, there is not guarantee that even 25ms will solve the problem.

oleost:
I just tested and confirm it worked.

theArchLadder:
@oleost Did you only test one time? Did it feel annoyingly slow?

oleost:
No, I have 10s to wait when restoring entire FC when I update. Even if it would take a minute I still have that time to being certain that everything is being restored correctly.

And when just updating a single value I dont even notice its slower.

theArchLadder:
I just did a few tests, 70 lines(settings, aux etc) was changed from the defaults.
3 restores with 5ms delay, 3 (not sure why this one went so well), 44 and 40 wrong lines.
2 restores with 10ms delay, first had 21 lines wrong, second had 28 lines wrong.
2 restores with 20ms delay, 0 and 20 lines wrong. Perhaps 25ms is not good enough?
2 restores with 40ms delay, 0 and 7 lines wrong... Why does the first attempt work but not the second?!
3 restores at 20ms delay with reconnecting between attempts, 0, 0 and 1 (aux) line wrong, dammit.
2 restores at 20ms delay with restarting configurator between attempts, 0 and 9 lines wrong, dammit.
3 restores at 100ms delay, 0 lines wrong three times in a row, yay! Need to do at least 10 test in a row but i'm way to annoyed to do that now, this was boring! I must automate this so i don't go nuts...

oleost:
@theArchLadder Still good results with 100ms delay?

If so maybe @DzikuVx could change to 100ms delay and then get it merged?

theArchLadder:
@oleost I have not done any more testing, it's so boring! Hopefully i will get around to it some day, but feel free to help test it!

oleost:
Tried 5 in a row with 100ms delay. Not a single line wrong.

DzikuVx:
I will test Today too

theArchLadder:
Feels like the proper solution would be some kind of confirmation message from the FC for each setting that has changed... Anyone know how to do something like that?

DzikuVx:
That probably would require a change in MSP protocol implementation and as far as I know, break any compatibilty. But maybe I'm wrong

skaman82:
is this helpful? https://github.com/cleanflight/cleanflight-configurator/pull/354

theArchLadder:
@skaman82 Yes, i think so! Thanks.

@DzikuVx How about adding that extra delay for profile commands?
https://github.com/cleanflight/cleanflight-configurator/pull/354/files

theArchLadder:
Hmm, wait a minute, betaflight dumps all profiles, right? Does cleanflight do that? Maybe it's only useful in that case, because inav doesn't.

theArchLadder:
Done some more testing.
100ms has always worked for me so far.
The code skaman82 linked to did not seem to make it possible to reduce that 100ms.

But i did find something interesting:
When restoring the dump, the CLI prints out everything that gets restored, but when it misses something it does not get printed! This looks to be true for both settings and aux, i'm sure someone clever could use this to get a confirmation after each restored line... Thoughts?

oleost:
Also tried skaman82 code 549166f9bfcf0866ce7077f1c92e51bf917cd785

Modes worked, but still got lost lines.

oleost:
Please merged this? I need to change it manually everytime and I can think of any reason not merging it?

PR #8, opened 2016-05-24
GUI rework

skaman82:
bildschirmfoto 2016-05-24 um 17 41 42

bildschirmfoto 2016-05-24 um 20 11 51


skaman82:
I also noticed that there ist a "calibration" menu entry but the tab-files for that are missing... Is someone working on that?

theArchLadder:
Wow, beautiful! :smiley: 

Is there any green left? For example, in the blackbox tab there is green when a button is pressed, and the progress bar for downloading the log is green.

Do you have an automated way to search and replace colors in all files? I could perhaps come up with something if you don't.

DzikuVx:
Yay

digitalentity:
@skaman82 cool, I absolutely like it! Please merge at will!

I've started working on Calibration tab, but apparently forgot to commit the new files. Unfortunately they are on my home PC which is inaccessible at the moment.

skaman82:
The are some areas untested due to missing equipment:
• Onboard SD Card interface
• Race Transponder Tab

@theArchLadder I can't find where the green color is coming from in this dialog:
bildschirmfoto 2016-05-24 um 19 55 42. Everything else should be fine

skaman82:
Merging now, will do a small follow-up in another PR, if necessary

theArchLadder:
Just tried it, looks fantastic, amazing work! :+1:

Issue #18, opened 2016-06-18
Create calibration tab file


cnHeider:
Calibration tab, makes other tabs unusable until disconnect.
![2016-06-18-102339_1366x768_scrot](https://cloud.githubusercontent.com/assets/509350/16169887/3f33a158-353f-11e6-934b-6ce70684b77c.png)

DzikuVx:
This tab is not working yet. Unfortunately we know

DzikuVx:
"Calibration" tab progress should be part of #20

skaman82:
@digitalentity Any progress on this? I can help create the GUI for this if I know how you like to implement the process.

digitalentity:
@skaman82 Blank calibration page is part of #20 

I want to have 3 calibration options on this tab:
**1. Accelerometer calibration**
Here I'd like to have one button "Calibrate" and 6 pictures of UAV in different positions (like in the wiki). By default all of them would be grey. FC will tell the Configurator which position it is calibrating and Configurator should light that picture by some color when it's working and if it's successful it should change color to green.
Also I want to have 6 boxes to show or type in current calibration values for accelerometer - accelerometer zero and accelerometer gain.

**2. Level calibration**
The button "LEVEL CALIBRATION" will simply take current tilt angles and adjust board alignment in such way that angles are zeroed. Maybe duplicate pitch-roll board alignment values here as well.

**3. Compass calibration**
Just a button and 3 input boxes to show or type in Mag zero values.

TIA for your help!

skaman82:
Will prepare the HTML/CSS for this in the next days

skaman82:
@digitalentity my part is ready
<img width="1551" alt="bildschirmfoto 2017-01-11 um 19 51 28" src="https://cloud.githubusercontent.com/assets/11768024/21861996/af55f71a-d837-11e6-957f-6c6ea37a35d9.png">

Issue #13, opened 2016-06-08
cannot open menu Calibration


risnandar:
just install Inav Configurator 1.0.0 with iNav 1.1
when i try to access menu 'Calibration' it just showing 'Waiting for data'.. don't have clue why

DzikuVx:
@risnandar this option is not implemented yet, do not open it

PR #19, opened 2016-06-18
calibration tab removed

DzikuVx:
This PR removes unfinished "Calibration" tab. There were multiple request about this tab: #18 #13 and I personally had to answer at least two emails on this topic.  

"Calibration" tab will be added as separate PR


DzikuVx:
@digitalentity how about merging it?

PR #26, opened 2016-07-14
Correct icon for airplane mixer.

oleost:
Output 1 Motor/ESC

Output 2 Motor/ESC

Output 3 Elevator

Output 4 Aileron

Output 5 Aileron

Output 6 Rudder


oleost:
Unsure why the file is so much bigger, I only dragged around the numbers and changed colour to make it correct.

Maybe because its a newer version of Inkscape, or they remove uneccesary stuff in the original file?

PR #35, opened 2016-08-09
Fixed to be up to date with latest changes in iNav

martinbudden:
@digitalentity , can you merge this ASAP. This catches up with latest iNav changes.

### DIFF
--- a/js/msp.js
+++ b/js/msp.js
@@ -854,6 +854,7 @@ var MSP = {
                 RX_CONFIG.rx_max_usec = data.getUint16(offset, 1);
                 offset += 2;
                 if (semver.gte(CONFIG.apiVersion, "1.21.0")) {
+                    offset += 4; // 4 null bytes for betaflight compatibility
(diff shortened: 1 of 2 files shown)


js/msp.js
```diff
@@ -1387,6 +1388,10 @@ MSP.crunch = function (code) {
             buffer.push(lowByte(RX_CONFIG.rx_max_usec));
             buffer.push(highByte(RX_CONFIG.rx_max_usec));
             if (semver.gte(CONFIG.apiVersion, "1.21.0")) {
+                buffer.push(0); // 4 null bytes for betaflight compatibility
+                buffer.push(0);
+                buffer.push(0);
+                buffer.push(0);
```
digitalentity: what are these nulls for?

martinbudden: The 4 nulls are to reserve alignment with betaflight, so that when I add NRF24 to betaflight the MSP serialisation will be the same.

Issue #34, opened 2016-07-28
Drop PID scaling


DzikuVx:
Drop PID scaling in configurator, use CLI values where possible

DzikuVx:
@digitalentity any thought here?

theArchLadder:
I agree with this, having different scaling in the GUI and CLI is just confusing on unit-less values.

digitalentity:
I agree. No need to scale anything, just use same values as in CLI to avoid confusion.

Ralf-W:
BTW "set roll_rate =20" is displayed as 200 deg/s in configurator. Please change GUI to 20 deg/s or add a decimal Point 20.0 deg/s.

digitalentity:
@Ralf-W FYI CLI displayes rate values in 10*deg/s, so `roll_rate=20` **is** 200 deg/s. It's also documented in wiki - https://github.com/iNavFlight/inav/wiki/Rate-conversion-for-INAV-1.2

Ralf-W:
@digitalentity 
Uups. Sorry, my fault.

PR #38, opened 2016-08-29
PID values scaling removed

DzikuVx:
This PR tries to solve #34.
Not yet ready, do not merge
@digitalentity could you setup labels like `Do not merge` and `In progress` ?


digitalentity:
Good enough? :smile:

DzikuVx:
As well as removing PID values scaling it optimizes JS on PID tab making it few dozen lines shorter

PR #43, opened 2016-09-13
sponsors section removed

DzikuVx:
@digitalentity do we have sponsors?


digitalentity:
@DzikuVx don't remove it completely, just keep multiwiicopter there - they were the very first who donated hardware and keept INAV going. More will come hopefully.

DzikuVx:
@digitalentity done

PR #60, opened 2016-11-03
Gyro configuration

DzikuVx:
- [x] MSP_INAV_PID frame
- [x] MSP_SET_INAV_PID frame
- [ ] Async modes GYRO and ALL


DzikuVx:
Basic support. Requires iNavFlight/inav#709

PR #71, opened 2016-11-22
Port of Betaflight OSD support

DzikuVx:


DzikuVx:
@digitalentity @martinbudden @oleost @stronnag PR is done
There were quite massive changes in MSP layer so please click through whole configurator.
Besides that, initial OSD support is ready, we will not go further in 1.5 that that

digitalentity:
I don't see any ill effects - everything seems to function properly

DzikuVx:
@jflyper found some minor issues I will investigate

digitalentity:
Found a bug. Non-detected (failing) sensors are flashing red. Probably Configurator is processing sensor status (and updated sensor iconds) from both `MSP_STATUS` and `MSP_SENSOR_STATUS` messages while in 1.5 only `MSP_SENSOR_STAUS` should be used for displaying sensor icons.

DzikuVx:
@digitalentity I will work on it while implementing sensor enabling

PR #75, opened 2016-11-30
Async processing GUI support

DzikuVx:


DzikuVx:
@digitalentity @oleost @martinbudden @sambas  UI to set async modes done. Probably. It would be nice it this is somehow tested before merge.

oleost:
Also if I set example way to high for my board(Spracingf3), like 8khz my CPU is at 600% usage and I'm struggling to even set it back. So some room for improvements regarding warning user. Also I just did try some varius things and one time I had to reflash my board since it would not connect at all. 

Also some explanation, or just a link to a docs page https://github.com/iNavFlight/inav/blob/master/docs/Looptime.md would probably be benifitial.

Like "Gyroscope task frequency" lists Hz, but changes a denom number (gyro_sync_denom)

Also "Flight controller loop time" Lists in Hz, while most people are often used to Cycle time.

I understand its different. But for me this is confusing. So we probably need clear instruction on how to use. And some explanation.

On the positive side, I did test alot, and it seemed like the correct values got changes when I verified in cli.

DzikuVx:
@oleost thanks for testing

1. Yes, additional explanations are required. I will add them to tooltips, hope they will help
1. Why everything is in **Hz** even when something else is set in the end? To simplify user interface. I see it this way: as a regular user I do not want to know that with gyro_lpf=188Hz gyro_denom = 1 gives me 1kHz while with gyro_lpf=256HZ gyro denom 1 gives 8kHz. Users should be taken away from all those technicalities as far as possible. This is why everything has been presented using the same unit: Hz. After all, Hz are used to measure frequencies, right?
1. Betaflight also presents those values using Hz in configurator. 
1. I think right now it is a "standard" to use frequency for looptime like in "i'm using 4kHz mode"
1. The problem with values that consumes all CPU values is that Configurator has no idea about max. capabilities of a board. This is purely hardware depending not only on CPU, but also bus used. SPI will be faster than I2C and so on. Maybe docs will help to understand those things

PR #81, opened 2016-12-17
Sensor config

DzikuVx:
@digitalentity @oleost Few days faster than I promised, please check


DzikuVx:
@digitalentity you might test it and merge

digitalentity:
Already on it

PR #92, opened 2017-01-04
Configurator presets

DzikuVx:


DzikuVx:
OK, This is done on functional level

@digitalentity @oleost @andre92 @stronnag you might want to take a look a defined profiles in `tabs/profiles.js` file and add new or modify existing. 

@skaman82 If you want you might take a look at it from UI perspective. ATM it's rather not pretty

Any all, I'm merging it into master since quite a few things changed and the faster someone will start to use it, the better

PR #110, opened 2017-01-16
Profile gui

DzikuVx:
@skaman82 so, shall we merge?


skaman82:
@DzikuVx sure. But we need to put some more text in for "Applies following settings:" and rearrange some content if you are fine with the solution for the 10" preset.

DzikuVx:
@skaman82 looks good. I will prepare some additional element classes to allow for better text formatting in list and then we can merge it back

skaman82:
Great

PR #121, opened 2017-01-29
Report raw baro altitude in sensors tab

digitalentity:
This side of https://github.com/iNavFlight/inav/pull/1186


digitalentity:
@DzikuVx I would appreciace if you have a look at this.

Issue #122, opened 2017-01-29
TPA threshold set to 0 after applying Preset 600mm Flying Wing


DzikuVx:
@oleost I remember correctly/

oleost:
Yes

 ```
    name: "600mm Flying Wing",
        description: "Small flying wing on multirotor racer parts<br>" +
            "<span>300g-500g weight, 3S-4S battery</span>",
        features: [
            "Adjusted gyro filtering",
            "Adjusted PIDs",
            "Adjusted rates"
        ],
        applyDefaults: ["PIDs", "INAV_PID_CONFIG", "ADVANCED_CONFIG", "RC_tuning", "PID_ADVANCED", "FILTER_CONFIG", "FC_CONFIG"],
        settings: [
            presets.elementHelper("BF_CONFIG", "mixerConfiguration", 8),
            presets.elementHelper("PIDs", 0, [15, 30, 15]),  //ROLL PIDs
            presets.elementHelper("PIDs", 1, [15, 40, 15]),  //PITCH PIDs
            presets.elementHelper("RC_tuning", "roll_rate", 400),
            presets.elementHelper("RC_tuning", "pitch_rate", 150),
            presets.elementHelper("ADVANCED_CONFIG", "gyroSync", 1),
            presets.elementHelper("INAV_PID_CONFIG", "gyroscopeLpf", 1)
        ],
        type: 'flyingwing'
    }
];
```

Looking at that preset I'm guessing it will happend to all preset which does not have an defined TPA_

```
presets.defaultValues = {
    PIDs: [
        [40, 30, 23],
        [40, 30, 23],
        [85, 45, 0],
        [50, 0, 0],
        [65, 120, 10],
        [180, 15, 100],
        [10, 5, 8],
        [20, 15, 75],
        [60, 0, 0],
        [100, 50, 10]
    ],
    INAV_PID_CONFIG: {"asynchronousMode": "0", "accelerometerTaskFrequency": 500, "attitudeTaskFrequency": 250, "magHoldRateLimit": 90, "magHoldErrorLpfFrequency": 2, "yawJumpPreventionLimit": 200, "gyroscopeLpf": "3", "accSoftLpfHz": 15},
    ADVANCED_CONFIG: {"gyroSyncDenominator": 2, "pidProcessDenom": 1, "useUnsyncedPwm": 1, "motorPwmProtocol": 0, "motorPwmRate": 400, "servoPwmRate": 50, "gyroSync": 0},
[...]

PR #124, opened 2017-01-29
Fix profiles tpa

DzikuVx:
@oleost it fixes some problems from #122 (TPA Threshold and rates)
Some have to wait for iNavFlight/inav#1187

You might want to test it


digitalentity:
#1187 is (or at least should be) fixed by #1189

DzikuVx:
Another issue: different sets of FW pids will be required for Presets tab

PR #148, opened 2017-02-19
Add new preset Wing win Z84

oleost:
Need help setting:

set tpa_rate = 33
set tpa_breakpoint = 1300
set gyro_hardware_lpf = 20HZ
set fw_p_level = 10

### DIFF
--- a/tabs/profiles.js
+++ b/tabs/profiles.js
@@ -232,6 +232,27 @@ presets.presets = [
         ],
         type: 'flyingwing'
     }
+        {
+        name: "Flying wing Z84",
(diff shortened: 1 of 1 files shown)


DzikuVx:
@oleost I will merge and add requrested settings in a few minutes.

oleost:
@DzikuVx great, would be nice if you provide some info where to find the info required.

Think being able for hardcore users to create new presets is beneficial.

I'm planning to try and write some pointers.

If you could dump some link or additional valuable info here : https://github.com/iNavFlight/inav/wiki/Request-form-new-PRESET

PR #160, opened 2017-03-09
OpenLayers and OpenStreetMap instead of Google Maps

DzikuVx:
@digitalentity you might want to check and deploy as soon as possible.
Today we exceeded the quote on GM and it was required to clean it up.
So.... OpenStreetMap, here we come... Rest of planned changes will come later


stronnag:
You appear to have lat,lon the wrong way around. I am NOT in the Indian Ocean!
screenshot-20170304215940-1253x500

stronnag:
Confirmed, tested 'live'. Looks great.

digitalentity:
Tested, works:
image

DzikuVx:
In that case merging back

Issue #134, opened 2017-01-31
Failsafe changes in 1.6+


digitalentity:
In 1.6 failsafe is going to be changed a lot:

 1. There's no Stage 1 anymore
 1. Failsafe is always enabled (feature FAILSAFE is deprecated)
 1. There is a new "do nothing" failsafe procedure which will keep enabled modes and inputs as they are.

![2017-01-31 13 37 24](https://cloud.githubusercontent.com/assets/11059099/22451479/88812e38-e7ba-11e6-9dd8-1fc01c3c030c.png)

oleost:
Also is shows "Stage 2" as disabled. even when enabled.

This must for sure confuse alot of users and needs updating.

DzikuVx:
Implementation in place, closing

PR #164, opened 2017-03-10
failsafe changes for 1.6

DzikuVx:
Failsafe changes from #134

@digitalentity @oleost please take a look

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -1656,16 +1656,16 @@
         "message": "<strong>Hold</strong> means maintain the last good value received. <strong>Set</strong> means the value given here will be used"
     },
     "failsafeStageTwoSettingsTitle": {
-        "message": "Stage 2 - Settings"
+        "message": "Settings"
     },
     "failsafeFeatureItem": {
-        "message": "Failsafe Stage 2 enabled"
+        "message": "Enabled"
     },
     "failsafeFeatureHelp": {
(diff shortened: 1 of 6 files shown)


oleost:
Looks good.

image

Maybe we could add an small description to "RTH" ?

Espescially what confuses many people are the "throttle value used while landing". Maybe somehow make it clear that its different from RTH.

oleost:
Found a bug.

oleost:
When changing stuff in "configurator" tab it doesnt reconnect after save and reboot, it only gives me "You need to upgrade your firmware before you can use the configuration tab"

It does not happend on chrome version 1.6.2. Im also running latest inav 1.6.1 release.

image

( I can access the configuration tab when reconnect manually )

DzikuVx:
@oleost This is general "reconnect" problem issue, not Failsafe page specific. I will really have to fix how Configurator handles reconnect and solve it once and for all

oleost:
Are you sure? Newer seen that error message before. And it's fine on 1.6.2 release.

DzikuVx:
Will take a look in the evening. Which board were you using?

oleost:
spracingf3, default settings.

oleost:
Never mind. Wrong board.

On airbotf4 I can replicate it.

https://www.youtube.com/watch?v=ztWGW1pkCvI

digitalentity:
@DzikuVx looks good

DzikuVx:
@oleost Just tested reconnect in this tab on AirbotF4 and SPracingF3. In both cases works like expected. Airbot requires some pateince not to click "connect" but both boards gave me 10/10

DzikuVx:
In case of furter reconnect issues create general "reconnect" ticket

PR #166, opened 2017-03-11
some rationalization in PID table generation and PID names no longer …

DzikuVx:
…taken directly from FC


digitalentity:
Looks good!

PR #188, opened 2017-04-24
Mention that they change pwm refresh for servos.

oleost:
Added in description for tricopter that they apply faster pwm for servos.


DzikuVx:
I also rebuilded JS files

oleost:
@DzikuVx ?

DzikuVx:
```
For local development, node.js build system is used.

1. Install node.js
1. From project folder run npm install
1. To rebuild JS and CSS use ./node_modules/gulp/bin/gulp.js
```

Issue #199, opened 2017-06-02
F7 flashing


sambas:
F7 flashing would need this patch:
https://github.com/betaflight/betaflight-configurator/commit/e4794221b95c8a682b790e8f23148984479f87aa

DzikuVx:
@sambas I was able to flash F7 with INAV configurator without any problems

sambas:
It's the 722 that has issues

DzikuVx:
Implementation in place, @sambas test please

PR #200, opened 2017-06-04
F72x flashing fix

DzikuVx:
Adresses #199 

@sambas could you please test if it solves the problem with F72x ?


DzikuVx:
OK, merging and let's see what will happen :)

PR #215, opened 2017-07-04
New GPS module type MTK

marbalon:


DzikuVx:
@marbalon setup a link to corresponding firmware pull request so we do not forget to merge when needed

marbalon:
Related with this PR
https://github.com/iNavFlight/inav/pull/1819

js/fc.js
```diff
@@ -570,6 +570,10 @@ var FC = {
             data.push('UBLOX7')
         }
 
+        if (semver.gte(CONFIG.flightControllerVersion, "1.7.2")) {
```
fiam: It seems like the PR for INAV implementing this feature (https://github.com/iNavFlight/inav/pull/1819) was merged after 1.7.2 was released, so this should be changed to `semver.gt()`. Otherwise, it looks fine to me.

digitalentity: Or stay with `semver.gte` but bump the version up to `1.7.3`

PR #221, opened 2017-08-06
Add rcsplit support to inav

azolyoung:
Hi there,

Here is azol, I'm adding rcsplit control to iNav FC, this pull request is about the configurator that support rcsplit.

### DIFF
--- a/tabs/ports.html
+++ b/tabs/ports.html
@@ -21,6 +21,7 @@
                         <td>Telemetry</td>
                         <td>RX</td>
                         <td>GPS</td>
+                        <td class='peripherls-column'>Peripherals</td>
                     </tr>
                 </thead>
                 <tbody>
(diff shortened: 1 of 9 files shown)


tabs/ports.js
```diff
@@ -4,6 +4,7 @@ TABS.ports = {};
 
 TABS.ports.initialize = function (callback, scrollPosition) {
     var board_definition = {};
+    var isSupportPeripherals = semver.gte(CONFIG.apiVersion, "1.27.0");
```
digitalentity: Actual MSP API didn't change. I think it's better to check for firmware version, not for apiVersion. @DzikuVx what do you think?

azolyoung: @digitalentity 
I've update it, and I'm post a  wrong comment of the last commit in this PR, in my code, I'm using FC version to detect which is support peripherals configuration or not.
But actually, the result is same.

digitalentity: Great! I need to test this

tabs/ports.html
```diff
@@ -62,6 +63,9 @@
                 <td class="functionsCell-gps"><select class="gps_baudrate">
                         <!-- list generated here -->
                 </select></td>
+                <td class="functionsCell-peripherals"><select class="blackbox_baudrate">
```
digitalentity: Do I understand correctly that it would be using the same baud rate as selected for blackbox?

PR #225, opened 2017-08-20
Set MOTOR_STOP and AIRMODE features in all FW presets

fiam:
AIRMODE will only be set if INAV version is > 1.7.2, since it
does nothing in previous versions even when the feature is
advertised.

### DIFF
--- a/build/script.js
+++ b/build/script.js
@@ -21444,6 +21444,15 @@ presets.model = (function () {
 
             }
         }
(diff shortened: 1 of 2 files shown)


fiam:
Should we get this one in before 1.7.3 or wait for 1.8?

digitalentity:
Yes, this should go in before 1.7.3

PR #226, opened 2017-08-24
Split OSD character maps into images

fiam:
- Use images to store the character sets and compile `.mcm` files from the images themselves
- Delete the ßF logo from the character set
- Icon cleanups (add missing chars to non-default fonts, combine 2-char images)

Related to https://github.com/iNavFlight/inav/issues/1933 and https://github.com/iNavFlight/inav/issues/1756


DzikuVx:
Unfortunately I have one very strong objection against this change: few hundred files are a no-no.
The reason is simple: when chrome will stop supporting packed aps, we will have to switch to NW.js. And it hates big number of files because compression issues: build for windows can start for minutes if number of files in project if large. 

How about rendering one PNG with all chars and generating CSS sprites as well?

fiam:
@DzikuVx The small PNGs don't need to be shipped with the app. The .mcm files are kept up to date and the configurator is parsing the font files. You don't need a big PNG nor the CSS sprites for anything and the small ones are so much convenient to edit and manage.

DzikuVx:
So hmm.. I'm not following what is what. If PNG does not have to be shipped, why are they there? Also, are those fonts from BF or INAV? We have slightly different fonts with symbols that does not appear in BF. @marbalon knows all the details here

fiam:
@DzikuVx The PNGs are there for making the charmaps easier to maintain, not for shipping them. The idea is that anyone can add/remove/modify PNG files with their favourite editor and then just run the tool to update the charset. This severely reduces the maintenance burden. I thought excluding files from the Chrome app was a non-issue, but we can put the source PNGs somewhere else if that's a problem.

Regarding the current charmaps, they are missing some nice to have symbols (e.g. mp/h) and still have the ßF logo on it. It looks like they were just pulled from ßF and they had no INAV specific modifications and it seemed to me that one was properly maintaining them. However, I used the ones from the master branch on this repo as a starting point and I've kept all the existing characters at the same location.

I do plan to add some more characters in the near future, in order to display the navigation modes, failsafe and other info in a better way. Unfortunately, neither ßF nor MWOSD have characters for those, so we either implement it [poorly] using pure text or we need to create and maintain specific charsets for INAV.

digitalentity:
@fiam can we generate fonts on the fly so we won't have to build fonts?

fiam:
@digitalentity That's definitely a very good idea. We can generate the font file on the fly in the configurator, so we don't need to keep the opaque `.mcm` files around.

What's your opinion regarding the small icons in the repo and the packed app? That seems to be the strongest objection by @DzikuVx to merge something like this.

digitalentity:
@fiam note that INAV has different heading arrow orientation than Betaflight. There are also other differences in fonts which need to be adressed. 

I don't think @DzikuVx will object against keeping only the PNGs - it's the large number of files he's concerned about. If we can have one PNG per font (all glyphs in one file) - all will be perfect.

fiam:
Just to clear and settle the confusion. I mentioned that the ßF logo was present in the font files in the current master in this repo, but I didn't use the .mcm files from ßF. I did use the ones from INAV, so all the original symbols are there in their original locations.

@digitalentity Let me understand something first. What's the current workflow for working with the character set for the OSD?  I ask because editing the `.mcm` with that Java tool or modifying a big PNG is a total pain, so I must be missing something.

Regarding the nw.js issue, I've just checking the documentation and excluding files from the NW build is definitely non-issue. From https://github.com/nwjs-community/nw-builder/blob/master/README.md:

```
options.files Required

Type: String
Default value: null

The path to your node webkit app. It supports simple-glob so you can do stuff like ['foo/*.js', '!foo/bar.js', 'foo/bar.js'].
```

Maybe it was't the case the last time @DzikuVx looked into it, since the package.json in the configurator requires nwjs 0.18 and latest release is 0.24.4, but right now excluding files is just a matter of listing the ones you don't want in your package. I've also tried to run the configurator in NW (just in case the objection was regarding the development workflow) and it starts instantly, despite the big number of small PNGs.

With these new findings, I'd ask you to reconsider your position regarding the small PNG files. They're the true source form for this data as well as the most convenient format to edit them, so it's only logical that they're in the repo. We can derive any other representations we need from them either at commit time or at build time or, most conveniently and as @digitalentity  suggested, at runtime.

DzikuVx:
@fiam @digitalentity here are my points:

1. I like the idea that working with `.mcm` is simpler
1. We can not just switch to BF files since some of our symbols are just different and not compatible
1. I deed I did not checked NW.js files exclusion for some time, maybe this indeed has been improved
1. To be honest, I would prefer this font building feature to be pulled out from Configurator and moved to small separate project. I think this would make a nice concern separation. In a small new project we can keep files however we want. The only "problem" would be copying `.mcm` files as PR into Configurator project

How about that?

fiam:
@DzikuVx Let's get to the bottom of this, because I must be missing something. How are you currently adding/editing icons? I find working with the `.mcm` directly a total pain, since the only tools I found for managing them were pixel based editors (this was like 8 months ago, so something might have changed). This makes painting anything way more difficult for me, but I guess it would feel natural for someone used to make pixel art.

Note that I want to add the small PNGs to the repo because I want to make it easier for others to edit the character map. I already do have the images split and the tools to merge them at my disposal, so having them on the repo doesn't save me any additional work. So, if you already have a convenient method of editing the character maps that works for you, I guess that you don't see the advantage of introducing another simpler but different method.

Regarding your point 2, I HAVEN'T AND I WON'T PURSUE SWITCHING TO ßF's CHARACTER SET (caps for emphasis, don't take me wrong). I did comment that the current character sets still had the ßF logo on it which INAV is not using and that's what made me think that maintaining those files was too painful since someone would have already removed that logo otherwise. For reference, here is the character set on master (1st) and the one on my branch (2nd):

old


default

fiam:
@DzikuVx Ok, I think I understood all the points you mentioned.

If I'm not wrong, the biggest problem we have with this PR is just a misunderstanding (English not being the 1st language for any of us certainly doesn't help in this department). I'm keeping the `.mcm` files in the repository. I only added the PNGs to make it easier for others to edit the character maps. As you can see in this [README](https://github.com/fiam/inav-configurator/tree/osd_images/resources/osd) for the `resources/osd` folder I've added a Makefile using a 3rd party open source tool to update the `.mcm` files before commiting the changes to the repo (that 3rd party happens to be me, but we could use another tool if you find it more convenient). In fact, in that same folder you can see that `default.mcm`, `bold.mcm` and `large.mcm` are still there. 

The only difference is that now the PNGs are right there in the repo for anyone to modify or add new ones and there are also clear instructions on how to do so. Note that the PNGs are not used at runtime. They're available to easily regenerate the `.mcm`,  just like you would e.g. commit the `.y`, `.l` and the generated `.c` files to the repo when using yacc and bison. It's a common pattern when you use autogenerated files but you don't want to require every developer to install those tools first.

Now, regarding packacking a NW.js app, that's something I hadn't thought of. I can, however, add some rules to automatically package an NW.js app excluding all the PNGs as well as all the `.css` and `.js` files which have been compiled into `build/script.js`, `build/style.css`, etc... something which isn't currently done at the moment (again, support for it might be recent, I hadn't touched NW.js before). All you'll need to do is run `./node_modules/nw-builder/bin/nwbuild` and it will generate all the package apps for the supported platforms.

Would you agree to merge these changes if I also take care of the packaging of the NW.js app and you make sure it's not negatively affecting the performance of the final apps?

DzikuVx:
@fiam now that we have a understanding, let's do the following:

1. Let's merge this
1. I will make a ticket to prepare NW.js configuration without OSD chars PNG's, `*.js` from `js` folder and other `*.css` besides build ones
1. I will add a ticket to add `nwbuilder` to dependencies, it is currently missing

Deal?

fiam:
@DzikuVx Deal, thank you! I'll submit a PR for the NW issue later today, since I have some of the work done already.

PR #227, opened 2017-08-26
Implement MSPv1 JUMBO frames and MSPv2

fiam:
While debugging some other issues, I found it annoying that the configurator was hanging up from time to time due to JUMBO frames, so I went ahead and implemented support for them as well as for MSPv2.

### DIFF
--- a/js/msp.js
+++ b/js/msp.js
@@ -32,21 +32,27 @@ var MSP = {
         TO_MWC: '<'.charCodeAt(0),
         UNSUPPORTED: '!'.charCodeAt(0),
     },
+    constants:                  {
+        JUMBO_FRAME_MIN_SIZE:       255,
+    },
     decoder_states:             {
-        IDLE:                   0,
-        PROTO_IDENTIFIER:       1,
-        DIRECTION_V1:           2,
-        DIRECTION_V2:           3,
-        PAYLOAD_LENGTH_V1:      4,
-        PAYLOAD_LENGTH_V2_LOW:  5,
-        PAYLOAD_LENGTH_V2_HIGH: 6,
-        CODE_V1:                7,
-        CODE_V2_LOW:            8,
-        CODE_V2_HIGH:           9,
-        PAYLOAD_V1:             10,
-        PAYLOAD_V2:             11,
-        CHECKSUM_V1:            12,
-        CHECKSUM_V2:            13,
+        IDLE:                       0,
+        PROTO_IDENTIFIER:           1,
+        DIRECTION_V1:               2,
+        DIRECTION_V2:               3,
+        PAYLOAD_LENGTH_V1:          4,
+        PAYLOAD_LENGTH_JUMBO_LOW:   5,
+        PAYLOAD_LENGTH_JUMBO_HIGH:  6,
+        PAYLOAD_LENGTH_V2_LOW:      7,
+        PAYLOAD_LENGTH_V2_HIGH:     8,
+        CODE_V1:                    9,
+        CODE_JUMBO_V1:              10,
+        CODE_V2_LOW:                11,
+        CODE_V2_HIGH:               12,
+        PAYLOAD_V1:                 13,
+        PAYLOAD_V2:                 14,
+        CHECKSUM_V1:                15,
+        CHECKSUM_V2:                16,
     },
     state:                      0, // this.decoder_states.IDLE
     message_direction:          1,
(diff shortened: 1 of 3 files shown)


js/serial_backend.js
```diff
@@ -246,8 +246,13 @@ function onOpen(openInfo) {
 
         FC.resetState();
 
-        // request configuration data
+        // request configuration data. Start with MSPv1 and
+        // upgrade to MSPv2 if possible.
+        MSP.protocolVersion = MSP.constants.PROTOCOL_V1;
         MSP.send_message(MSPCodes.MSP_API_VERSION, false, false, function () {
+            if (CONFIG.apiVersion && semver.gte(CONFIG.apiVersion, "1.27.0")) {
```
digitalentity: We didn't bump API version up when merging MSPv2. I think we should increase version major to "2.0.0" and do the same change in firmware.

fiam: That make sense. Increase it in INAV and I'll update the PR.

BTW, what's the long term plan for the MSPv1 protocol? Will it be removed or kept around for the foreseeable future?

digitalentity: MSPv1 in the firmware will be kept as long as there are tools that use it (like mwosd, EzGUI etc). As for Configurator - we'll probably remove it after a while (probably a long while).

fiam: It seems we'll be stuck with supporting MSPv1 for a long time then. I was asking because if MSPv1 was dropped from the FC side, the MSPv2 detection logic would eventually need to be changed (e.g. send MSPv2 1st, fallback to MSPv1 on timeout), but it seems that's not gonna be the case for a long time.

digitalentity: MSPv1 doesn't put much maintenance burden on FW side. We don't deprecate old messages, MSPv2 is merely more frexible transport layer, so there's no point in deprecating MSPv1 on firmware side. Configurator is another story - once major part of users will be MSPv2 aware we can drop v1 support in Configurator.

digitalentity:
@DzikuVx see https://github.com/iNavFlight/inav/pull/2042 for MSP API version bump. As soon as we have this merged I'll use bigger blackbox payloads when using MSPv2 (no need for jumbo frames).

digitalentity:
@fiam firmware part ready and merged. Please update API version to 2.0.0 here so we can merge this

digitalentity:
Found a bug. This PR breaks `MSP_DATAFLASH_READ`. 
```
var chunkAddress = response.data.getUint32(0, 1);
```
seems to always return 0 regardless of requested address. Forcing `protocolVersion = 1` doesn't show this behavior. Everything else works.

digitalentity:
`MSP.send_message` appears to send zeros instead of actual payload.

js/msp.js
```diff
@@ -160,48 +274,60 @@ var MSP = {
...
+                view[0] = this.symbols.BEGIN;
+                view[1] = this.symbols.PROTO_V2;
+                view[2] = this.symbols.TO_MWC;
+                view[3] = 0; // flag: reserved, set to 0
+                view[4] = code & 0xFF;  // code lower byte
+                view[5] = (code & 0xFF00) >> 8; // code upper byte
+                view[6] = payloadLength & 0xFF; // payloadLength lower byte
+                view[7] = (payloadLength & 0xFF00) >> 8; // payloadLength upper byte
+                for (var ii = 0; ii < payloadLength; ii++) {
+                    view[7+ii] = data[ii];
```
digitalentity: This overwrites payloadLength upper byte. Should be 
```
view[8+ii] = data[ii];
```

fiam: @digitalentity Nice catch! I'll submit the requested changes in a few minutes. There are also a few style issues I've noticed since I submitted the PR that I plan to address. It was late when I finished writing that code, don't judge me :stuck_out_tongue_winking_eye:

digitalentity: No worries. I suspect there is another bug with MSP payload serialization. Debugging.

digitalentity:
Yay! Tested a hacky implementation of large payloads. `MSP_DATAFLASH_READ` is blazing fast when using 4K payload size!

js/msp.js
fiam: @digitalentity Just pushed the changes for version detection and those style fixes. I don't have time to run any exhaustive tests right now, so please let me know if you find any other problems and I'll look into it.

PR #231, opened 2017-08-28
NW.js updated to latest stable

DzikuVx:


package.json
```diff
@@ -21,7 +21,7 @@
   "author": "iNavFlight",
   "license": "GPL-3.0",
   "dependencies": {
-    "nw": "^0.18.0",
+    "nw": "^0.24.4",
```
fiam: Is there an easy way to use `0.24.4` for release builds and `0.24.4-sdk ` for development? Versions without `-sdk` suffix don't support the inspector view.

fiam: OK, seems I should have checked all available `nw-build` arguments before commenting. `nw-build` uses pre-made builds for NW.js, so it doesn't matter if we use an `-sdk` version for development, the release build can easily use non `-sdk` versions. So I'd suggest to use `0.24.4-sdk` instead.

PR #251, opened 2017-09-16
Update STM32 USB DFU driver

digitalentity:
Fixes the hang-up when flashing DFU devices with "Full flash erase" option enabled.


DzikuVx:
tested on F7, flashing works, full chip erase seems to be working as well. Merging

PR #253, opened 2017-09-18
Add preview support for OSD vertical elements

fiam:
Also, add preview for OSD_VARIO

### DIFF
--- a/tabs/osd.js
+++ b/tabs/osd.js
@@ -29,7 +29,8 @@ SYM.DIR_TO_HOME = 0x60;
 SYM.DIST_TO_HOME = 0xA0;
 SYM.HEADING1 = 0xA9;
 SYM.HEADING2 = 0xA8;
-SYM.VARIO = 0x9F;
+SYM.VARIO_UP_2A = 0xA2;
+SYM.M_S = 0x9F;
 SYM.CLOCK = 0xBC;
 SYM.LAST_CHAR = 188;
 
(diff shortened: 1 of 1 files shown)


tabs/osd.js
```diff
@@ -667,8 +672,18 @@ OSD.GUI.preview = {
         //noinspection JSUnresolvedVariable
         var field_id = parseInt(ev.dataTransfer.getData('text'));
         var display_item = OSD.data.display_items[field_id];
-        var overflows_line = FONT.constants.SIZES.LINE - ((position % FONT.constants.SIZES.LINE) + display_item.preview.length);
-
+        var line_width = 0;
+        var item_width = 0;
+        for (var ii = 0; ii < display_item.preview.length; ii++) {
+            if (display_item.preview[ii] == '\n'.charCodeAt(0)) {
+                item_width = Math.max(item_width, item_width);
```
DzikuVx: This doesn't makes sense

tabs/osd.js
```diff
@@ -667,8 +672,18 @@ OSD.GUI.preview = {
...
+                line_width = 0;
+                continue;
+            }
+            line_width++;
+        }
+        item_width = Math.max(item_width, item_width);
```
DzikuVx: The same?

fiam:
@DzikuVx Nice catch! I have a bunch of uncommitted stuff in my working copy, so I manually edited the patch and I ended up messing up that line, which then I copied onto the other one. I'll amend the commit right now.

DzikuVx:
@fiam I think your amend did not amended anything `Math.max` is still there. Just keep it simple and make another commit :) `You should not lie to GIT` :)

fiam:
@DzikuVx I stashed all my changes, applied this patch and made the changes over it to avoid making the same error (not that it would surprise me if I still made the same error 😄 ). `Math.max()` is there to set the element width to the maximum width calculated for each of its lines. Note that I changed the second argument from `item_width` to `line_width`, so the line actually does something now.

DzikuVx:
I trust you know what you are doing :)

PR #257, opened 2017-09-24
Organize OSD elements into groups

fiam:
Implement groups which can be declared in any order including
any item in the order deemed more useful. This way we no longer
need to put elements ordered by their osd_items_e value.

Also, elements and even groups can be disabled when the features
or hardware required for drawing them is not present or not
enabled.


fiam:
Screenshot of OSD tab with this PR applied:

screen shot 2017-09-24 at 11 04 33

Issue #162, opened 2017-03-10
Problem: Calibrate magnetometer button greyed out


avirotstein:
Calibrate magnetometer button in the inav configurator (1.6.2) is greyed out unless first clicking sensors tab and then clicking back to the setup tab

AustinSaintAubin:
Problem remains in Configurator 1.7.0 w/ firmware ( INAV/SPRACINGF3MINI 1.7.0 Apr 19 2017 / 09:28:25 (65cbad0) )

dragon1020:
Problem remains in Configurator 1.7.1. :(

Ogocology:
So sorry to post on an old issue.
I have this problem at the moment and I cant figure out exactly how to make it go away.

Any help please?

cdsisf:
![problema gps](https://user-images.githubusercontent.com/37956843/38163610-dbc37876-34ee-11e8-8fcb-d7f062e7d83e.jpg)
the same problem here, here are my connections

Ogocology:
Hello,For the calibration grey out issue, the temporary fix is to click on any other tab ( like configuration first) then go back to calibration. That solved it for me.
Your GPS showing red possibly means you got the connection wrong. Switch the TX and RX connections and check.
Br,Ogocology
    On Saturday, March 31, 2018, 2:22:03 PM GMT+1, cdsisf <notifications@github.com> wrote:  
 
 

the same problem here, here are my connections

cdsisf:
Changing Tab does not solve the problem. You really are right, by doing this I will activate the button but after finishing the calibration time I can not record the calibration. Maybe it's the MAG that has something wrong. The MAG appears in blue but if turned upside down it turns red. Although it seems that the MAG (HMC5983) is recognized, in the Tab of the sensors shows no sign of life.

PR #262, opened 2017-09-29
Fix disabled calibrate mag button

### LINKED ISSUE #245: Calibrate magnetometer button greyed out
I have the Omnibus F4 V3. Holybro Ublox NEO-M8N GPS Module with HMC5883L digital compass. the Holybro does turn green and start to flash the blue led(?) after start time on the unit. When i enable the Mag to this unit INAV causes the ready light on the Omnibus FC to flash rapidly. I assume this is because the compass has not been calibrated. IF i turn off the MAG the flashing stops. Anyway i do not  know if this is a Hardware problem or INAV. So, Can not fly. NO Calibrate Compass button.

Please advise

Thanks

digitalentity:
Looks like we have a possible race condition in Setup tab code between `periodicStatusUpdater` and `TABS.setup.initialize`. 

On connect `CONFIG.activeSensors` is zero. If `TABS.setup.initialize` calls `process_html` before `periodicStatusUpdater` polls for sensor status "Calibrate magnetometer" button will be disabled.

This PR queries for `MSP_STATUS_EX` message explicitly.

Should resolve #245, #162

### DIFF
--- a/js/msp/MSPHelper.js
+++ b/js/msp/MSPHelper.js
@@ -1957,6 +1957,10 @@ var mspHelper = (function (gui) {
         MSP.send_message(MSPCodes.MSP_BF_CONFIG, false, false, callback);
     };
 
+    self.quaryFcStatus = function (callback) {
+        MSP.send_message(MSPCodes.MSP_STATUS_EX, false, false, callback);
+    };
+
     self.loadMisc = function (callback) {
         MSP.send_message(MSPCodes.MSP_MISC, false, false, callback);
     };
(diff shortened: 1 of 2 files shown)


digitalentity:
@DzikuVx @fiam please review

js/msp/MSPHelper.js
```diff
@@ -1957,6 +1957,10 @@ var mspHelper = (function (gui) {
         MSP.send_message(MSPCodes.MSP_BF_CONFIG, false, false, callback);
     };
 
+    self.quaryFcStatus = function (callback) {
```
DzikuVx: `queryFcStatus` not `quaryFcStatus`

digitalentity: Fixed!

PR #267, opened 2017-10-05
Add support for the RX changes in MSP

fiam:
Implement a combobox selector for the receiver type, showing
just the options for the specific type (e.g. selecting serial
receiver reveals the combobox for the serial protocol).

On INAV > 1.7.3 read and write receiver type to RX_CONFIG rather
than to feature bits. Note that this will break the configurator
in builds with version = 1.7.4 before https://github.com/iNavFlight/inav/pull/1596/files
is applied.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -464,7 +464,7 @@
     "RX_SERIAL": {
         "message": "Serial-based receiver (SPEKSAT, SBUS, SUMD)"
     },
-    "RX_PARALLEL_PWM": {
+    "RX_PWM": {
         "message": "PWM RX input (one wire per channel)"
     },
     "RX_MSP": {
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -458,17 +458,30 @@
         "message": "EEPROM <span style=\"color: #37a8db\">saved</span>"
     },
 
-    "featureRX_PPM": {
+    "RX_PPM": {
         "message": "PPM RX input"
     },
+    "RX_SERIAL": {
+        "message": "Serial-based receiver (SPEKSAT, SBUS, SUMD)"
+    },
+    "RX_PARALLEL_PWM": {
+        "message": "PWM RX input (one wire per channel)"
+    },
+    "RX_MSP": {
+        "message": "MSP RX input (control via MSP port)"
+    },
+    "RX_NRF24": {
+        "message": "RX SPI based receiver (NRF24L01, RFM22)"
+    },
+    "RX_NONE": {
+        "message": "No receiver"
+    },
+
     "featureVBAT": {
         "message": "Battery voltage monitoring"
     },
     "featureINFLIGHT_ACC_CAL": {
         "message": "In-flight level calibration"
-    },
-    "featureRX_SERIAL": {
-        "message": "Serial-based receiver (SPEKSAT, SBUS, SUMD)"
     },
     "featureMOTOR_STOP": {
         "message": "Don't spin the motors when armed"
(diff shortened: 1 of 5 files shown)


digitalentity:
Will test ASAP!

digitalentity:
Doens't work correctly with https://github.com/iNavFlight/inav/pull/1596

Configurator:
image

CLI:
image

digitalentity:
Looks like Configurator still uses feature bits even if version is 1.7.4 :thinking:

js/msp/MSPHelper.js
```diff
@@ -593,8 +593,12 @@ var mspHelper = (function (gui) {
                     offset++;
                     RX_CONFIG.nrf24rx_id = data.getUint32(offset, true);
                     offset += 4;
+                    RX_CONFIG.nrf24rx_channel_count = data.getUint8(offset);
+                    offset += 1;
                 }
                 if (semver.gt(CONFIG.flightControllerVersion, "1.7.3")) {
+                    // unused byte for fpvCamAngleDegrees, for compatiblity with betaflight
```
DzikuVx: @fiam in RX_CONFIG frame?

fiam: Yep, that's where ßF sends that byte. I guess their reasoning is that it affects input mixing.

js/fc.js
```diff
@@ -604,6 +600,68 @@ var FC = {
...
+        var rxTypes = [
+            {
+                name: 'RX_SERIAL',
+                bit: 3,
+                value: 3,
+            },
+            {
+                name: 'RX_PPM',
+                bit: 0,
+                value: 1,
```
digitalentity: PWM and PPM are mixed up:
```
typedef enum {
    RX_TYPE_NONE        = 0,
    RX_TYPE_PWM         = 1,
    RX_TYPE_PPM         = 2,
    RX_TYPE_SERIAL      = 3,
    RX_TYPE_MSP         = 4,
    RX_TYPE_SPI         = 5,
} rxReceiverType_e;
```

_locales/en/messages.json
```diff
@@ -458,18 +458,31 @@
...
+    "RX_SERIAL": {
+        "message": "Serial-based receiver (SPEKSAT, SBUS, SUMD)"
+    },
+    "RX_PARALLEL_PWM": {
+        "message": "PWM RX input (one wire per channel)"
+    },
+    "RX_MSP": {
+        "message": "MSP RX input (control via MSP port)"
+    },
+    "RX_NRF24": {
```
digitalentity: Shouldn't we rename `RX_NRF24` to `RX_SPI` for consistency?

js/fc.js
fiam: I think those are incorrectly named in the configurator code. There's `RX_PPM` and `RX_PARALLEL_PWM` and those are the same thing! Shouldn't it be `RX_PPM` and `RX_PWM`?

_locales/en/messages.json
fiam: We can take the opportunity to do so. On which order should they appear in the combobox? I guess `RX_SERIAL` should be the first one, not sure about the rest though.

js/fc.js
digitalentity: Agree, let's rename them to RX_PPM and RX_PWM

_locales/en/messages.json
digitalentity: Let's make it SERIAL, PPM, PWM, SPI, MSP, NONE

fiam:
@digitalentity Should be ready for prime time now.

fiam:
Merging this one, since the INAV part is already in.

PR #269, opened 2017-10-06
Disable "Load Firmware [Online]" button while downloading it

fiam:
Provides feedback to the user to let he know that the configurator
didn't hang up

screen shot 2017-10-06 at 15 46 47

### DIFF
--- a/tabs/firmware_flasher.js
+++ b/tabs/firmware_flasher.js
@@ -338,7 +338,7 @@ TABS.firmware_flasher.initialize = function (callback) {
 
             var summary = $('select[name="firmware_version"] option:selected').data('summary');
             if (summary) { // undefined while list is loading or while running offline
-                $("a.load_remote_file").text(chrome.i18n.getMessage('firmwareFlasherButtonLoading')).addClass('disabled');
+                $(".load_remote_file").text(chrome.i18n.getMessage('firmwareFlasherButtonLoading')).addClass('disabled');
                 $.get(summary.url, function (data) {
                     enable_load_online_button();
                     process_hex(data, summary);
--- a/tabs/firmware_flasher.js
+++ b/tabs/firmware_flasher.js
@@ -27,6 +27,10 @@ TABS.firmware_flasher.initialize = function (callback) {
         // translate to user-selected language
         localize();
 
+        function enable_load_online_button() {
+            $("a.load_remote_file").text(chrome.i18n.getMessage('firmwareFlasherButtonLoadOnline')).removeClass('disabled');
+        }
+
         function parse_hex(str, callback) {
             // parsing hex in different thread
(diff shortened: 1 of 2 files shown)


tabs/firmware_flasher.js
```diff
@@ -27,6 +27,10 @@ TABS.firmware_flasher.initialize = function (callback) {
         // translate to user-selected language
         localize();
 
+        function enable_load_online_button() {
+            $("a.load_remote_file").text(chrome.i18n.getMessage('firmwareFlasherButtonLoadOnline')).removeClass('disabled');
```
DzikuVx: element type selector is redundant, there is no reason to keep `a` at all in here, it only makes selector slower

tabs/firmware_flasher.js
```diff
@@ -329,11 +333,14 @@ TABS.firmware_flasher.initialize = function (callback) {
             function failed_to_load() {
                 $('span.progressLabel').text(chrome.i18n.getMessage('firmwareFlasherFailedToLoadOnlineFirmware'));
                 $('a.flash_firmware').addClass('disabled');
+                enable_load_online_button();
             }
 
             var summary = $('select[name="firmware_version"] option:selected').data('summary');
             if (summary) { // undefined while list is loading or while running offline
+                $("a.load_remote_file").text(chrome.i18n.getMessage('firmwareFlasherButtonLoading')).addClass('disabled');
```
DzikuVx: if we have separate function to enable, why not for disabling?

fiam: There are 2 disabled states. One when there's no firmware selected in the top comboboxes and another one while the firmware is being downloaded. The button will show different texts. There's only one enabled state, so it extracted that to a function.

DzikuVx: OK, understand

tabs/firmware_flasher.js
fiam: Just copied the existing selector already used by the code. I don't think performance will be a problem in this case and since we're selecting in class rather than id, I think the `a` at the front makes it less likely to screw up in the future.

DzikuVx: class and id has now almost the same performance. Could you just make me this pleasure and remove this ugly `a` please?

fiam: Sure, no problem!

fiam: Just to be clear... That `a` or all the occurrences of `a.load_remote_file`? There are 4 or 5 already in the file.

DzikuVx: The ones you touched and I will pretend I did not saw the rest :)

fiam:
I hope you don't mind my "creative" commit messages. Bought that stuff you recommended me and is making me feel funny.

PR #272, opened 2017-10-08
Minimum failsafe distance

StewLG:
This adds support for the Minimum Failsafe Distance patch in the Configurator:

https://github.com/iNavFlight/inav/pull/2244

I believe everything functions OK, but I could use some guidance about how to properly deal with changing MSP and/or FC version numbers. Specifically, I'd like to know exactly what to do at Line 1200 in MSPHelper.js and line 113 in Failsafe.html.

minimumfailsafedistanceconfigurator


js/fc.js
```diff
@@ -464,6 +471,14 @@ var FC = {
             );
         }
 
+        // Unsure which version these will actually deploy in, guidance welcome. Using 1.7.3 
+        // so it works for the moment. And I'm really not sure how it works anyhow
+        if (semver.gte(CONFIG.flightControllerVersion, '1.7.3')) {
+            features.push(
+                {bit: 30, group: 'rxFailsafeExtended', name: 'FAILSAFE_EXTENDED', haveTip: false, showNameInTip: false}
```
digitalentity: Why changing the features? Relevant firmware change is not using features.

js/msp/MSPHelper.js
```diff
@@ -1194,6 +1208,33 @@ var mspHelper = (function (gui) {
                 buffer.push(lowByte(FAILSAFE_CONFIG.failsafe_throttle_low_delay));
                 buffer.push(highByte(FAILSAFE_CONFIG.failsafe_throttle_low_delay));
                 buffer.push(FAILSAFE_CONFIG.failsafe_procedure);
+                buffer.push(FAILSAFE_CONFIG.failsafe_recovery_delay);
+
+                // API version check needed here?
+                //if (semver.gte(CONFIG.apiVersion, "1.21.0")) {
+                // Or
+                // if (semver.gte(CONFIG.flightControllerVersion, "1.7.3")) {
```
digitalentity: Checking for `flightControllerVersion` is better

StewLG:
* Changed from meters to centimeters, updated relevant text
* Removed question about feature in code
* Settled on checking for version FlightController version 1.7.3. If it should be a different version, let me know, or change at will.

digitalentity:
@DzikuVx @fiam do you think this can be merged?

fiam:
On the phone, so make sure I didn’t read it incorrectly. It seems the version check is gte 1.7.3. Should be **gt** 1.7.3. Everything else seems fine. Will take a look from the computer later.

StewLG:
Thanks @fiam, made the change to gt 1.7.3.

I had not realized developer branch was version 1.7.4; I had started this PR on master which which was on 1.7.3. Which explains my initial confusion about what to put here.

js/msp/MSPHelper.js
```diff
@@ -1210,7 +1210,7 @@ var mspHelper = (function (gui) {
                 buffer.push(FAILSAFE_CONFIG.failsafe_procedure);
                 buffer.push(FAILSAFE_CONFIG.failsafe_recovery_delay);
 
-                if (semver.gte(CONFIG.flightControllerVersion, "1.7.3")) {
+                if (semver.gt(CONFIG.flightControllerVersion, "1.7.3")) {
```
digitalentity: Wrong `gt`. `failsafe_fw_roll_angle` etc ARE present in 1.7.3

js/msp/MSPHelper.js
```diff
@@ -1194,6 +1208,27 @@ var mspHelper = (function (gui) {
...
+                    buffer.push(lowByte(FAILSAFE_CONFIG.failsafe_fw_pitch_angle));
+                    buffer.push(highByte(FAILSAFE_CONFIG.failsafe_fw_pitch_angle));
+
+                    buffer.push(lowByte(FAILSAFE_CONFIG.failsafe_fw_yaw_rate));
+                    buffer.push(highByte(FAILSAFE_CONFIG.failsafe_fw_yaw_rate));
+
+                    buffer.push(lowByte(FAILSAFE_CONFIG.failsafe_stick_motion_threshold));
+                    buffer.push(highByte(FAILSAFE_CONFIG.failsafe_stick_motion_threshold));
+
+                    buffer.push(lowByte(FAILSAFE_CONFIG.failsafe_min_distance));
```
digitalentity: You should protect only `failsafe_min_distance` and `failsafe_min_distance_procedure` with a check for 1.7.4+

StewLG:
Made change so that only two new variables are protected for 1.7.4 or greater.

Should we also protect the other existing variables that weren't implemented in the configurator until this PR? failsafe_recovery_delay appears here https://github.com/iNavFlight/inav/commit/836ce5e835614af91feb2599a52911e03af35f00, for example. Are we worried about that?

digitalentity:
>Should we also protect the other existing variables that weren't implemented in the configurator until this PR? failsafe_recovery_delay appears here iNavFlight/inav@836ce5e, for example. Are we worried about that?

Now you mentioned that, I think we should... Those settings are only available in 1.7.3. Can you please add a guard for them as well?

StewLG:
1.7.3 settings guarded.

js/msp/MSPHelper.js
```diff
@@ -617,6 +617,20 @@ var mspHelper = (function (gui) {
                 offset += 2;
                 FAILSAFE_CONFIG.failsafe_procedure = data.getUint8(offset);
                 offset++;
+                FAILSAFE_CONFIG.failsafe_recovery_delay = data.getUint8(offset);
```
DzikuVx: Those settings were introduced only around February, older versions of firmware will not send them. We should find a release that intrudced them and use this data segment only is firmware version allows for it

tabs/failsafe.js
```diff
@@ -225,6 +231,7 @@ TABS.failsafe.initialize = function (callback, scrollPosition) {
         $('input[name="failsafe_off_delay"]').val(FAILSAFE_CONFIG.failsafe_off_delay);
         $('input[name="failsafe_throttle_low_delay"]').val(FAILSAFE_CONFIG.failsafe_throttle_low_delay);
         $('input[name="failsafe_delay"]').val(FAILSAFE_CONFIG.failsafe_delay);
+        $('input[name="failsafe_min_distance"]').val(FAILSAFE_CONFIG.failsafe_min_distance);
```
DzikuVx: Like I wrote before, older versions of firmeare (1.6) might not sent this data. Firmware version should be checked. I suggest allow for it only starting from gte 1.7.4

StewLG:
@DzikuVx, I've made the changes you pointed out this needed. It now should shows/hides the controls properly when talking to 1.7.3/1.7.4, and it only should be manipulating 1.7.4 data structures when actually present. Let me know if I've missed anything else.

Again, I appreciate all the help very much.

digitalentity:
@StewLG thanks! Let's wait for  @DzikuVx approval and merge this along with it's firmware counterpart https://github.com/iNavFlight/inav/pull/2244

DzikuVx:
@digitalentity @StewLG approved

PR #277, opened 2017-10-10
Support more than 32-bits of modes using MSP_ACTIVEBOXES API

digitalentity:
Now we can support up to 52 box modes (20 more than now)

### DIFF
--- a/tabs/modes.js
+++ b/tabs/modes.js
@@ -11,6 +11,10 @@ TABS.modes.initialize = function (callback) {
         googleAnalytics.sendAppView('Modes');
     }
 
+    function get_active_box_data() {
+        MSP.send_message(MSPCodes.MSP_ACTIVEBOXES, false, false, get_box_ids);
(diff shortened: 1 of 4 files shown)


tabs/modes.js
```diff
@@ -11,10 +11,6 @@ TABS.modes.initialize = function (callback) {
         googleAnalytics.sendAppView('Modes');
     }
 
-    function get_box_data() {
```
DzikuVx: To be honest I would leave this message in place. periodicStatusUpdater can be postponed by serial port balancing and it's not guaranteed that active modes will be loaded when TAB is entered

digitalentity: MSP_BOX was not implemented at all - it was reused as MSP_ACTIVEBOXES.

digitalentity: I'll add it back

digitalentity:
Should be ready now

PR #289, opened 2017-10-18
Don't test for RX_NONE on INAV <= 1.7.3

### LINKED ISSUE #287: 1.8.0 not able to set Receiver Mode of 1.7.3 target AIR3 ; backward compatibility issue
If i use 1.8.0 to change the receiver mode of a board coded 1.7.3 ; for example to Multiwire - on reboot it shows "no receiver" ; or other settings,  similar - whatever is chosen / saved - after beboot is corrupted to another    = bug

if i use 1.7.3 to change the receiver mode of a board coded 1.7.3 ; for example to Multiwire - on reboot it shows multiwire - correctly saving and showing the new mode.

fiam:
Since in RX_NONE bit is undefined, bit_check() will always return
true. Fix is just removing RX_NONE from the available RX types
on versions using feature bits for RX, since they won't allow
not having a RX configured (FC will fall back to RX_PPM).

Fixes #287

### DIFF
--- a/js/fc.js
+++ b/js/fc.js
@@ -642,10 +642,14 @@ var FC = {
             value: 4,
         });
 
-        rxTypes.push({
-            name: 'RX_NONE',
-            value: 0,
-        });
+        // Versions using feature bits don't allow not having an
+        // RX and fallback to RX_PPM.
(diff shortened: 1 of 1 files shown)


fiam:
@digitalentity That was fast!

digitalentity:
:sunglasses:

PR #298, opened 2017-10-30
Change the OSD config to show ACRO instead of the defunct STAB flight mode placeholder

teckel12:


teckel12:
@digitalentity This is to remove STAB from the flight mode preview and make it ACRO instead to match INAV v1.8.0

PR #302, opened 2017-11-18
Add support for changing the craft name

fiam:
Uses MSP_NAME/MSP_SET_NAME to get and set it. A new section was
added to the Configuration tab which includes an input for the name.
Also, some CSS changes to allow wider fields for string inputs.

iNav PR at https://github.com/iNavFlight/inav/pull/2488


fiam:
@DzikuVx Just fixed the conflicts on this one. Can we get it merged?

PR #305, opened 2017-11-24
Fix toggle color in OSD Tab

skaman82:
Removed the yellow color from toggle elements and replaced it with the proper blue color.


skaman82:
bildschirmfoto 2017-11-24 um 13 47 34

Issue #234, opened 2017-09-04
Pending issues to release the configurator as an NW.js app


fiam:
- [x] Fix drag and drop in the OSD tab. It fails on NW.js 0.24.4 (fixed in #237)
- [x] Add an update mechanism. Can be as simple as showing a dialog to the user when a new version of the configurator is released.
- [x] The app always launches with a 1024px window, which doesn't show tab names. Consider making the app bigger by default or, at least, save the last used size/position. We need to take into account devices with screens 1024px wide, so we can't just make the app require 1200px.
- [x] Build easily redistributable package for Windows (fixed in #235)
- [x] Build easily redistributable package for macOS (fixed in #235)
- [x] Build easily redistributable package for Linux

fiam:
I think this can be closed. The remaining points have been fixed by @Afinogen .

PR #314, opened 2017-12-26
Save the last used size/position

Afinogen:
Save the last used size/position for issue #234
Need to check whether my window closing handler has affected on function createdWindow.onClosed
I did not understand how it works (createdWindow.onClosed).

### DIFF
--- a/main.js
+++ b/main.js
@@ -38,30 +38,34 @@ $(document).ready(function () {
             break;
     }
 
-    // Load native UI library
-    var gui = require('nw.gui');
-    var win = gui.Window.get();
-    //Get saved size and position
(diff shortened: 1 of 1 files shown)


Afinogen:
I found the problem, if I run from the application from the browser, then does not find the JS required (
It is necessary to make a test of existence or add to the gulp

digitalentity:
We need to test that this code works for stand-alone win/macos applications.
---
Нужно протестировать код и убедиться что он корректно работает в автономных приложениях для win/macos

Afinogen:
@digitalentity , I can not check for macOs - I do not have such a system
Мака у меня нет, даже у знакомых...
На винде без проблем, сегодня проверю

PR #323, opened 2017-12-31
Fix old firmware

Afinogen:


digitalentity:
Good. Now Calibration works correctly for both 1.8.0 and 1.8.1

PR #324, opened 2018-01-02
Mission planer

Afinogen:
The first version, I accept offers and wishes.
Download and save does not work yet


DzikuVx:
wow… great beginning!

Afinogen:
Need to fix the css styles

DzikuVx:
@Afinogen is this any usable?

Afinogen:
@DzikuVx I did not understand the question
If it's about styles, then need to make adaptability

DzikuVx:
I mean if i can plan a mission and upload. Basic functionality and all of tha

Afinogen:
@DzikuVx  Yes, it work. Checked in MP for android

Afinogen:
Corrected styles, now looks better

digitalentity:
Seems working. Not perfect, but we'll improve it on the way

PR #337, opened 2018-02-07
Add FW_LAUNCH feature to configurator

giacomo892:
Adds configurator support to https://github.com/iNavFlight/inav/pull/2731

Is there a way I can check for "Fixed Wing" and display the option only if "Fixed Wing" is set on the FC?


DzikuVx:
@giacomo892 yes, you have to check for mixer type. It will change in a future, but for now this is the way

PR #339, opened 2018-02-12
Add support for viewing trace output via MSP from the sensors tab

fiam:
Use an additional window to show the debug trace, so the configurator
can be used at the same time the user is viewing the output.


digitalentity:
I think we can merge this PR already

PR #353, opened 2018-02-19
Fix typo in MSPV2_INAV_MISC decoding

shellixyz:
Not a big issue the only consequence is that after you select mWh as battery capacity unit the drop down menu will show a blank item selected. But it needs to be selected again before saving again otherwise the setting will go back to mAh.


shellixyz:
@digitalentity Please merge this for 1.9 it is fixing a bug in the GUI.

digitalentity:
@shellixyz thanks

PR #358, opened 2018-02-22
Add support for displaying vertical speed in the baro graph

fiam:
Should we also rename this graph to Altitude rather than Barometer?


digitalentity:
Aren't 3 graphs a bit too much?

PR #359, opened 2018-02-24
Add feature for enabling/disabling profile selection by TX stick command

shellixyz:
Matching configurator update for https://github.com/iNavFlight/inav/pull/2830


digitalentity:
I'm not sure if this should go into 1.9 or 2.0. Change seems small.

PR #363, opened 2018-03-01
Set capacity thresholds as percentage of battery capacity

shellixyz:
As discussed in https://github.com/iNavFlight/inav/issues/2855


digitalentity:
Cool! No possibility for confusion now :+1:

Issue #364, opened 2018-03-01
Mission Config: lat and lon labels transposed


stronnag:
The mission lat and lon labels are transposed. The data is correct wrt to the geo display and other ground stations (e.g. when downloaded from FC into mwp). 

lat should be the Y axis, lon the X axis, not vice-versa.

PR #367, opened 2018-03-01
Mission planer v1.1

Afinogen:
New appearance of the interface, need testing
Add RTH, landing and speed


shellixyz:
Looking good, nice improvement

tabs/mission_control.js
```diff
@@ -307,6 +310,8 @@ TABS.mission_control.initialize = function (callback) {
                 $('#pointLon').val(coord[1]);
```
shellixyz: The Latitude and Longitude are transposed see #364 and #366.

Lines 306 and 307 corrected:
```
$('#pointLon').val(coord[0]);
$('#pointLat').val(coord[1]);
```

PR #369, opened 2018-03-02
Reword text and tooltip for auto_disarm_delay

fiam:
Make it clear that this setting disarms after THR has been low
for the given number of seconds and it's only used with stick
arming.


fiam:
This is how it looks now

screen shot 2018-03-02 at 14 58 01
screen shot 2018-03-02 at 14 58 06

PR #392, opened 2018-03-21
Add support for configuring multiple OSD layouts

fiam:
Don't merge this yet, the INAV part is still in progress so I might need to make some changes.

INAV part at https://github.com/iNavFlight/inav/pull/2955

### DIFF
--- a/tabs/osd.js
+++ b/tabs/osd.js
@@ -259,14 +259,14 @@ var OSD = OSD || {};
 
 // common functions for altitude and negative altitude alarms
 function altitude_alarm_unit(osd_data) {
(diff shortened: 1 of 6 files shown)


tabs/osd.js
```diff
@@ -768,8 +793,51 @@ OSD.get_item_preview = function(item) {
     return preview;
 }
 
+OSD.use_191_api = function() {
+    return semver.gte(CONFIG.flightControllerVersion, '1.9.1');
```
digitalentity: This should be replaced to 2.0.0 😀

PR #401, opened 2018-04-05
Add MSP2_INAV_OUTPUT_MAPPING decoding

shellixyz:
Related to https://github.com/iNavFlight/inav/pull/3008

The mapping is stored in OUTPUT_MAPPING. It is an array, one entry by output. Each entry is an array of possible usages for the output: MC_MOTOR, MC_SERVO, FW_MOTOR, FW_SERVO

### DIFF
--- a/js/msp/MSPHelper.js
+++ b/js/msp/MSPHelper.js
@@ -1330,21 +1330,8 @@ var mspHelper = (function (gui) {
                 break;
             case MSPCodes.MSPV2_INAV_OUTPUT_MAPPING:
                 OUTPUT_MAPPING = [];
-                output_usage_values = [
-                    { usage: 'MC_MOTOR', value: (1<<2) },
-                    { usage: 'MC_SERVO', value: (1<<3) },
-                    { usage: 'FW_MOTOR', value: (1<<5) },
-                    { usage: 'FW_SERVO', value: (1<<6) }
-                ];
-                for (i = 0; i < data.byteLength; ++i) {
-                    output_usage_value = data.getUint8(i);
-                    output_usage = [];
-                    for (usage_value_index = 0; usage_value_index < output_usage_values.length; ++usage_value_index) {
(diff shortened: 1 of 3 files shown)


js/msp/MSPHelper.js
```diff
@@ -2409,6 +2427,10 @@ var mspHelper = (function (gui) {
         MSP.send_message(MSPCodes.MSPV2_INAV_MISC, false, false, callback);
     };
 
+    self.loadOutputMapping = function (callback) {
+        MSP.send_message(MSPCodes.MSPV2_INAV_OUTPUT_MAPPING, false, false, callback);
```
DzikuVx: @shellixyz please add a version check so call on older firmware wont cause a problem

shellixyz: I'm not sure it is a good idea to add a check there. If we add a check and the function does nothing the data won't be loaded/refreshed in the OUTPUT_MAPPING var and the version will still need to be checked by the code using the OUTPUT_MAPPING data. For other messages the version is checked directly in the tab code itself.

Example from the receiver tab JS:
```
    if (semver.gte(CONFIG.flightControllerVersion, '1.8.1')) {
        loadChain.push(mspHelper.loadRateProfileData);
    } else {
        loadChain.push(mspHelper.loadRcTuningData);
    }
```

DzikuVx: but then, if 2 tabs will be asking for that frame, then 2 frames would have to include the logic to add only when FC offers it. I'm not sure what is happening on receiver tab is good too. adding a check on msphelper level makes implementation simpler in my opinon

shellixyz: Right. I added a check `loadOutputMapping()` now returns false and clears `OUTPUT_MAPPING` if the firmware version is not compatible.

shellixyz:
Updated

PR #408, opened 2018-04-19
Added indicator for overlapping osd elements

warhog:
Added an indicator for overlapping osd elements. If you drag osd elements in a position where they overlap another element the osd gets screwed up (depending on element) by the FC later.
The changes add a feature to tint the background red for overlapping elements.
The implementation takes care of overlapping crosshair/horizon.
I've uploaded an image that shows overlapping craftname/horizon/sidebars and rssi/throttle.
[Example](https://ibb.co/k0poRn)

### DIFF
--- a/tabs/osd.js
+++ b/tabs/osd.js
@@ -928,6 +928,16 @@ OSD.GUI.preview = {
     }
 };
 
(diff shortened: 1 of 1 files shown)


Lord-Quake:
I just want to say that this is a very helpful feature!

PR #414, opened 2018-04-26
Upgrade nw.js version used for development

fiam:
Since builds are shipping with 0.29 right now, developing with
0.25 is a pain because it can't use the stored settings from
0.29+. This is not a problem for user builds, since nw.js automatically
upgrades the configuration to the newer format.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "inav-configurator",
-  "version": "1.9.3",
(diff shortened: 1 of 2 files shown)


fiam:
Don't merge this one yet, it ends up installing the non SDK version with npm install.

fiam:
Superseded by #527, closing

PR #434, opened 2018-05-07
Old versions cleanup

DzikuVx:
Always assume we are running at least INAV 1.6

### DIFF
--- a/js/fc.js
+++ b/js/fc.js
@@ -519,14 +519,10 @@ var FC = {
             );
         }
 
-        if (semver.gte(CONFIG.flightControllerVersion, "1.4.0")) {
-            features.push(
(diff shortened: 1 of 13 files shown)


js/fc.js
```diff
@@ -58,7 +58,7 @@ var FC = {
     MAX_SERVO_RATE: 125,
     MIN_SERVO_RATE: -125,
     isRatesInDps: function () {
```
digitalentity: Why do we still have this function if it's always return true?

DzikuVx: Funny fact is that this fuction is already not neede at all :)

DzikuVx:
OK, that should do it

PR #437, opened 2018-05-08
Use a switch to reverse servos rather than negative rates

fiam:
Just an UI change, same functionality.


fiam:
Merging this one, since it's now working both on 2.0 and 1.9.1.

PR #444, opened 2018-05-11
Fix OSD AHI preview

shellixyz:
Looks like #411 have been accidentally overwritten.


fiam:
Ouch, my bad when rebasing. Sorry!

PR #466, opened 2018-06-04
Fix bug in MSP_ACTIVEBOXES decoding

shellixyz:
In JS `(1<<32) == 1` so it was causing an issue when the FC was advertising more than 32 mode boxes.


DzikuVx:
@shellixyz shouyld we merge?

shellixyz:
Yes it is ready please merge it.

PR #473, opened 2018-06-13
Drop receiver tab channels plot

### LINKED ISSUE #463: Receiver Tab: Bar and Graph values differ
There are bars for each stick on the top and graphs on the bottom of the tab. The values of the bars are different from the values on the graphs

Please, see attached:
issue

For example, bar value CH7 is 2000 but the graph shows approximately 2080.
Yaw bar shows 1500 but the graph shows 1600
etc.

shellixyz:
As discussed with @DzikuVx. Closes #463.

### DIFF
--- a/tabs/receiver.html
+++ b/tabs/receiver.html
@@ -110,18 +110,6 @@
                 </div>
             </div>
         </div>
-        <div class="clear-both"></div>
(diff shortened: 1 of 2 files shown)


DzikuVx:
Thanks man!

PR #549, opened 2018-07-20
Updated 7inch preset

DzikuVx:


fiam:
I think we should merge this one for 2.0, the existing 7" profile is not very good.

PR #561, opened 2018-08-09
Add support for per-uart options for inversion and full/half duplex

fiam:
- Removed note about not disabling MSP from the top port, just not
allow the user to do it.
- Add port options as a button in a new column
- Use a modal overlay to select the port options

Version checks test for > 2.0.0, since this should be merged only
for 2.1

Configurator PR at https://github.com/iNavFlight/inav/pull/3726

screen shot 2018-08-09 at 21 57 28
screen shot 2018-08-09 at 21 57 36
screen shot 2018-08-09 at 22 00 45
screen shot 2018-08-09 at 22 27 49
screen shot 2018-08-09 at 22 27 56

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -880,10 +880,7 @@
         "message": "Until first arm"
     },
     "portsHelp": {
-        "message": "<strong>Note:</strong> not all combinations are valid.  When the flight controller firmware detects this the serial port configuration will be reset."
-    },
-    "portsMSPHelp": {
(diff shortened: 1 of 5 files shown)


danarrib:
This is super nice. Is there any reason it wasn't merged yet? Can I help?

fiam:
@danarrib I found some flaws in the UI after designing this. We do support multiple functions per port (e.g. telemetry while armed and MSP while disarmed), so right now this would need to be changed to set the inversion/"duplexity"/etc... per function and port instead of just per port. Alternatively, we might drop for multiple functions on the same uart, but I'm not sure how many people use that.

PR #595, opened 2018-11-09
Map provider user configurable

DzikuVx:
Changes:

* map provider can be either OpenStreetMap or Bing
* user has to provide own Bing Maps API key
* GPS tab is no longer compatible with packed chrome app. NW should be used


mwsadc:
How about adding google maps also please?

PR #612, opened 2018-12-17
Resize font manager window to fit the new fonts

shellixyz:


kardon18:
Hi Shelli,
I downloaded 2.1.0 configurator today. 
And wanted to bring this to your attention, before an issue is brought up. 

The configurator stalls while in the OSD tab with Windows 7 32bit, after the selected font is loaded and the hardware reboots.
The picture below shows were the OSD setup images stops at.
I have only tried it with a F411-wing and a Omnibus F4 pro board. And it acts the same with both.

If the FC is powered down after loading the font. And the Configurator is closed and reopened, it doesn't lock up then.
I haven't tried it with any other operation system.

capture

shellixyz:
Strange, I can't reproduce the issues (Linux x86_64 with F411-Wing and OMNIBUS)

shellixyz:
I don't think this has anything to do with the resizing of the font manager window. You should open a new issue.

kardon18:
@fiam 
I take this Windows 7 machine with me flying.

It does have Firefox as it default browsers. And the later versions of JavaScript play up a bit with Firefox.
So I made Chrome the default and install the latest Java.   
But.....it made no difference.   
It certainly didn't have any trouble with this Windows 7 machine, using the 2.0.0 configurator.

@shellixyz  Should I still open a new report? Or just leave this here for now. And wait and see if anyone else reports this?

kardon18:
@fiam
I take this Windows 7 machine with me flying.

It does have Firefox as its default browsers. And the later versions of JavaScript do not support Firefox.
So I made Chrome the default and install the latest Java.
But.....it made no difference.
It certainly didn't have any trouble with this Windows 7 machine, using the 2.0.0 configurator.

@shellixyz Should I still open a new report? Or just leave this here for now. And wait and see if anyone else reports this?

kardon18:
I would like to help... But I can't redirect the command location in Node.js, for the Inav Master.
Sorry for being slow! I won't take up any more of your time. 
At least your guys are aware of the issue.

PR #628, opened 2019-01-03
New mixer inputs for stabilized axises

DzikuVx:
This side of iNavFlight/inav#4148

### DIFF
--- a/js/fc.js
+++ b/js/fc.js
@@ -1129,10 +1129,10 @@ var FC = {
     },
     getServoMixInputNames: function () {
         return [
-            'Stabilised Roll',      // 0
-            'Stabilised Pitch',     // 1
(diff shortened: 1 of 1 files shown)


js/fc.js
```diff
@@ -1152,6 +1152,12 @@ var FC = {
             'RC Channel 14',        // 20
             'RC Channel 15',        // 21
             'RC Channel 16',        // 22
+            'Stabilized Roll+',     // 23
```
fiam: Maybe we should change the lines with `Stabilised...` (note the `s` instead of `z` - UK spelling) a few lines up to `Stabilized` for consistency? I think most of the configurator code favors US spelling.

DzikuVx:
I suppose it's ready to merge after 2.1 is released

PR #651, opened 2019-01-14
Make changes to code for the following reasons to Ubuntu 16.04 on a n…

powercycle:
…otebook:

 - Change to ol instead of openlayers (per suggestion from NPM)
 - Change main screen size to fit laptops with some room to spare (1024 x 700)

Note: had to add string-width and nw for this to work on NPM version 3.5.2 and nodejs version 4.2.6


sensei-hacker:
I don't know this didn't have any activity for six years, but currently on "ol": "^10.6.1" :)

Issue #648, opened 2019-01-11
2.1 Failure to restore a full backup


OlivierC-FR:
- Flash a fresh iNav 2.1.RC1
- Configure it all
- CLI dump, save to a file
- Reflash the board again with same RC1
- Paste the full dump from the TXT file in the CLI, save
- Restart it all, dump the CLI again, paste into a new file
- Compare both files : Only about the first 70-80% of the settings are saved.

Consistent over 2 different PCs (Window 10 Laptop, Windows 10 Desktop) and 4 flight controllers : Matek F405-CTR x 2, F405-wing, F405-CTD

These screencaps for one particular PC/aircraft combo, but it's doing the same with rest of the gear.
I don't recall if it was cutting at the exact same place, but was very similar, about 20-30% of the last commands are not sent when the dump is pasted.

For this sample it skips 140 lines, starting at line 818.
The 3 empty lines on the paste side are added by me for clarification.

![dif2](https://user-images.githubusercontent.com/25566932/51055876-81bd4300-15e1-11e9-9433-b6dddd074d06.jpg)

![dif4](https://user-images.githubusercontent.com/25566932/51055890-8b46ab00-15e1-11e9-850e-6066ab9662f8.jpg)

kardon18:
Did you try to paste `DIFF ALL` files, instead of `DUMP`?

OlivierC-FR:
No, but that's not really the point, the point is to make a full backup :)

kardon18:
@OlivierC-FR 
I understand. 
That's what I was getting at. Whether the error would still occur, if the backup was pasted in small blocks.

I have noticed that 2.1.0 seems to write the backup files much faster that previous versions did. By just watching how fast the written settings go down the CLI.  Have you noticed that?

I remember @stronnag  mentioning the possibility of a compiler error occurring during file dumps; on RCG once. 
That's why I only save DIFF ALL back ups.

OlivierC-FR:
Yes its a lot faster to paste now, maybe too fast for its own good !

I made another flash this morning, different FC (Matek F450-WING), different PC (my Win10 laptop instead of min Win10 desktop).
Also fails to paste the whole dump, just this time it's only the very end.

![101](https://user-images.githubusercontent.com/25566932/51083394-cb10ae00-1719-11e9-99be-ebe23b0b7d16.jpg)
[...]

PR #661, opened 2019-01-23
slowdown CLI update rate to prevent missed lines

DzikuVx:


DzikuVx:
Possible fix to #648

DzikuVx:
Seems to be solving the issue, Dump, full chip erase, restore, dump and compare shows both files are identical

PR #666, opened 2019-01-27
OSD, previews for Icon/value/unit re-ordered 

OlivierC-FR:
... so it matches this iNav PR : https://github.com/iNavFlight/inav/pull/4652


shellixyz:
:smile: didn't remember/see that you did that I redid everything :man_facepalming:

PR #670, opened 2019-02-02
Temperature sensors support

shellixyz:
Configurator support for https://github.com/iNavFlight/inav/pull/4282

- Add temperatures to sensor tab
- Add OSD elements

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -2383,6 +2383,9 @@
     "osdGroupTimers": {
         "message": "Timers"
     },
+    "osdGroupTemperature": {
+        "message": "Temperature"
+    },
     "osdGroupAttitude": {
         "message": "Attitude"
     },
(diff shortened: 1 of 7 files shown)


kardon18:
Can I make a suggestion. After viewing the labels beside their temperature readouts.
As can be seen from the picture. There is a rather large gap between the `label` and the actual temperature value. Almost enough for the glyph's above it to fall through ;-)

I do understand that this gives a space between the Label and the temp readout value. Although, it ends up being a gap of, up to 5 spaces. If only a one character label is used.

Would it be possible to make the labels, first character positioning, with one space between it and the temperature value. And every other character(up to 4) would appear to the left of that first character? So that there is only ever a one character spacing between the label and temp value. 

Don't take this as criticism...I do understand your dilemma; in keeping all the characters lined up on both the label and the °C or °F end.
I just thought it might be more viewable with them closer together. Especially if just one temp value is located in the midst of other characters on a cluttered OSD page.


capture

shellixyz:
Indeed there can be a lot of space between the label and temperature if the label is short. I thought keeping the label alignment was the best thing to do. We can maybe add an alignment setting to chose between left or right alignment for the label but I think this issue will be solved when the possibility to use a symbol instead of the label will be added, so that less space is used. What do you think ?

kardon18:
I certainly understand why you went for letter characters over symbols. Because it gives more expansion of what can be added to the label's explanation.
I'm happy either way Michel :-)
But if you do go with the symbols. And would like them changed, because they don't quite suit your liking. Let me know. I'm more than happy to change them, to suit.

Here is what the symbols look like stacked; to give you an idea of whether you like them when they are bunched together.
capture

I also did this type of symbol. But I thought it was less viewable. And didn't look as nice with other label formats.

capture

There is one other option as well.
I did another complete font a while back.
It is a mini version of Impact, and is in the same style as Clarity_med and Vision as well..  But wouldn't suit the rest of the fonts.
An idea. Is to add the 26 letter and 10 numbers from it. To all the fonts.
That way you guys have complete (small scale) alphabet of characters for use in future developments, like this one.
The current version of Impact. Uses these letters on the graphical compass.  
nt osd 1

kardon18:
Did you give it anymore thought. With respect to an alignment setting or changing to the symbols?
I'm happy with either. 
But after using it in its present form. The gap between the label and temp value is very excessive.

shellixyz:
I haven't had time to take care of it. I had some more important things to do for the release (bugfix and other stuff) and the final release was planned for today so I thought it wouldn't be merged anyway. I'll take a look tomorrow if I have time, maybe it can be merged its not critical stuff.

PR #696, opened 2019-02-18
Add MAP_SCALE and MAP_REFERENCE elements

fiam:
Allow moving the map scale and reference around instead of them
being fixed on the screen.


fiam:
INAV PR at https://github.com/iNavFlight/inav/pull/4386

fiam:
Version checks need to be updated, since they're checking for GTE 2.1.0

PR #707, opened 2019-02-26
Further old versions cleanup

DzikuVx:


DzikuVx:
I was able to do most configuration of FC without errors. Works

PR #752, opened 2019-04-26
Make firmware flasher more robust

fiam:
Check for DFU/serial port regularly during 10 seconds instead of
waiting 1s and trying once. Also, make sure that we don't mistake
DFU devices with serial ones, since the serial port might appear
briefly while the FC is rebooting.


Docteh:
is this why sometimes a configurator wont notice when a board is put into DFU mode?

PR #766, opened 2019-05-11
Tradicional helicopter mixers

vltsvn:


tabs/mixer.js
```diff
@@ -333,22 +333,51 @@ TABS.mixer.initialize = function (callback, scrollPosition) {
...
-            modal.close();
+        }
+
+        function mixerApplyAndReboot() {
+            mixerApply();
             saveAndReboot();
+        }
+
+        function mixerApplyWithModal(apply) {
+            $('#pitch-roll-weight').val(50);
```
DzikuVx: Why this is set to 50 over here? Should it not be reseted to current value?

js/model.js
```diff
@@ -622,30 +656,37 @@ helper.mixer = (function (mixerList) {
...
     }
 
-    publicScope.loadMotorRules = function (mixer) {
+    publicScope.loadMotorRules = function (mixer, heliSettings) {
         MOTOR_RULES.flush();
 
         for (const i in mixer.motorMixer) {
             if (mixer.motorMixer.hasOwnProperty(i)) {
                 const r = mixer.motorMixer[i];
+                if ((mixer.platform==PLATFORM_HELICOPTER)&&(r.getYaw()!=0)&&!heliSettings.useTailMotor)
```
DzikuVx: Please use brackets even for single line conditions, it greatly increases readability

tabs/mixer.js
```diff
@@ -333,22 +333,51 @@ TABS.mixer.initialize = function (callback, scrollPosition) {
...
+            width: 480,
+            height: 240,
+            closeButton: 'title',
+            animation: false,
+            title: chrome.i18n.getMessage("heliModalTitle"),
+            content: $('#heliModalContent')
+        });
+
+        function mixerApply() {
+            heliModal.close();
```
DzikuVx: this is not enough. jBox we have over here has a bug that causes events to be rebinded every time modal is opened. We had some troubles with it before previous release. If user opens modal, closes, and then opens it again, there will be 2 events binded to modal elements in total.
The workaround is to remove modal wrapper and delete the modal instance as well. See ae6c37118d1da1ce1ea360cd67c4b078f6ec62f4#diff-7db851820cf7a896309fa2ecbee7e185

```
delete OSD.GUI.jbox;
$('.jBox-wrapper').remove()
```

js/model.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+const enableHelicopter = false; //This is for totally hide helicopter settings from user until all heli features will be ready
```
DzikuVx: So if I understand correctly, this is a dead code for now?
I think I saw a FEATURE for helis added. Maybe after all instead of false we can have it set to true if feature is set? 
After all, it's simpler to disable a feature than disable the feature and some variables in different tabs

tabs/mixer.js
digitalentity: @DzikuVx the reason for this is that this setting only exists as a "virtual" setting to calculate mixer weight. Will be changed if this kind of mixing is handled on FC side https://github.com/iNavFlight/inav/pull/4714)

js/model.js
digitalentity: Nope, we don't have a feature for HELIs. I think we should expose board capabilities in some MSP messages so Configurator can hide features that board can't support.

tabs/mixer.js
DzikuVx: That's what I remembered from FC pull requests. And that's why I'm asking. I'd prefer not to merge to configurator something that is not working on FC code. And also prefer to have configurator pull requests as complete as they can be. The way I see it, right now it's dead code that does not work

js/model.js
DzikuVx: I think all FCs will work to some extend with this. If FC code will be merged of course. 
I suggest to assume this to true, drop the constant and only merge when FC will be able to handle those new settings

tabs/mixer.js
digitalentity: So, TL;DR is that FC Heli support should be complete and ready. Makes sense - at the moment FC pull request does not allow heli mixer to be flyable AFAU.

vltsvn: @digitalentity  is right. It's not flyable for now, but it is disabled for now, so noone can try to fly it.
From your discussion I understand that I should complete all I need for helicopter to be flyable, then make some test flights, debug that as much as I can, commit code to these two branches (inav and inav-configurator). Am I right?

js/model.js
vltsvn: OK. I'll fix it

tabs/mixer.js
vltsvn: Understood

js/model.js
vltsvn: I don't know how fast code changes in this repo. So I desided to split my work on at least two parts for code in my branch doesn't become too old and make conflicts. I need a couple of weeks more to complete this to flyable state. This constant just hides what I created, because if this code will be merged the unready helicopter features must not be available for users.

vltsvn: 50 is the default value for textfield wasn't empty when opened

tabs/mixer.js
vltsvn: Fixed

js/model.js
vltsvn: Fixed

Issue #782, opened 2019-05-28
Load/Save mission not working [2.2.0 RC1]


mirko-it:
packages build problems, wip

mirko-it:
#783

mirko-it:
solved, browserified xml2js library in a new file (under js/libraries) and gulp build

noagendafpv:
I can confirm that the Save and then Load are not working on Win64 Config 2.2RC1

PR #783, opened 2019-05-28
Fix for Load/Save mission not working

mirko-it:
partial fix of #782 
still wip


tabs/mission_control.js
```diff
@@ -739,8 +737,8 @@ TABS.mission_control.initialize = function (callback) {
     }
 
     function saveMissionFile(filename) {
-        const fs = require('fs-extra');
-        const xml2js = require('xml2js');
+        const fs = require('fs');
```
DzikuVx: @mirko-it no problems with this package

mirko-it: should be something like native, it is working even in the executable, while "fs-extra" was not.
added an error log for both packages if not found

PR #786, opened 2019-05-31
 Gforce icons for Vision + preview

OlivierC-FR:


Jetrell:
Nice work Olivier. It will be very useful to have the identification for each axis.

OlivierC-FR:
yaw/roll/pitch are rotations, Gforce is from the accelemerometers, not the gyros. It's linear movements forward/backward, left/right and up/down

PR #822, opened 2019-08-11
Real time monitoring for Mission Control

CSMachineCo:
Added real-time position monitoring on the map with bread crumb trail.  Also added real time altitude, heading, speed, and distance display.  Tested on Arm based Debian (Pi), haven't tested on any other platforms.
Screenshot 2019-08-09 at 5 19 41 PM


DzikuVx:
amazing! Thank you very much, I will take a look at it when I will find a minute or two

CSMachineCo:
Is there a way to pull the RTH position from the FC via MSP?I thought that would be something useful to display, but I couldn't find it in the MSPhelper.js.

DzikuVx:
No, there is not, but I think we can add a MSP frame for that

DzikuVx:
@CSMachineCo I've made a firmware change that exposes home position with information if home position was aquired in https://github.com/iNavFlight/inav/pull/5006
It's not tested and you would have to write all the configurator changes by yourself too, but you might give it a try

DzikuVx:
@CSMachineCo apparently HomePosition is in waypoint 0 of MSP_WP. You could check that

CSMachineCo:
Ah... It looked like WP 0 was treated special. Makes sense now. I'll check
it out.

tabs/mission_control.js
```diff
@@ -189,6 +381,7 @@ TABS.mission_control.initialize = function (callback) {
         });
 
         iconFeature.setStyle(getPointIcon());
+        //iconFeature.setGeometryName('Waypoint');
```
DzikuVx: @CSMachineCo please remove if not used

CSMachineCo: Ok.  Removed that and added marker for RTH position.

DzikuVx:
and merged, thanks a lot

haslinghuis:
Any hints about how to use this? This will allow to transform configurator to a ground control station!

1. Indirect using Crossfire TX module with BT -> Send MAVlink to RPI over Bluetooth ?
2. Direct using a second crossfire nano connected to RPI.

PR #835, opened 2019-09-26
Add support for selecting FrSky OSD in the ports tab

fiam:
INAV side at https://github.com/iNavFlight/inav/pull/5090


Jetrell:
Nice stuff Alberto.
Can you give us a quick run down of how it will work with current fonts etc?

Issue #140, opened 2017-02-05
Limit FF max value


DzikuVx:
Limit max FF value.

" compute MAX value for FF gain and inform the user that this is only MAX value and real value HAS to be smaller that that"

Max FF should give max servo deflection (500) at max stick deflection (500)

digitalentity:
>Max FF should give max servo deflection (500) at max stick deflection (500)

Yes, but only in rate mode and this is not the final tuning value. Real tuning value will probably be much smaller than this.

DzikuVx:
Yes, this is only the upper value user will be able to enter. It will probably we always higher than defaults for firmware or presets

PR #900, opened 2020-01-22
New waypoint icon in all fonts

OlivierC-FR:
... for HUD display
#140


DzikuVx:
@OlivierC-FR could you resolve conflicts with clarity? Apparently there were some changes there too

PR #934, opened 2020-02-27
Dinamically show/hide info boxes in outputs tab

nmaggioni:


nmaggioni:
@DzikuVx That's true, but this logic grants that on a clean flash the users will view the notice at least once; in case they are carrying over non-default settings they will know why they changed them. Worst case scenario is that the notice will still be visible if your aircraft needs a higher idle, isn't it?

DzikuVx:
OK, let's merge it

PR #937, opened 2020-02-28
Cleanup Failsafe page

DzikuVx:


DzikuVx:
Adnotacja 2020-02-28 150024

Jetrell:
@DzikuVx While your doing the cleanup. I was thinking it may be less confusing to move `failsafe_off_delay` into the LAND procedure. At present it appears to be part of the RTH procedure, which its not.
i.e. move the line below `Delay for turning off the motors during failsafe`, not above it.

I don't know how many people use LAND procedure. But I often thought it would be nice to have `failsafe_off_delay` / ` Delay for turning off the motors during failsafe `tied into the barometer, if the FC had one.
That way it could shut the motors down according to altitude, and not just a time guesstimate.

DzikuVx:
`Valid Pulse Range limits` stays as it is in CLI, only GUI for it was removed.

Zataras:
> @DzikuVx While your doing the cleanup. I was thinking it may be less confusing to move `failsafe_off_delay` into the LAND procedure. At present it appears to be part of the RTH procedure, which its not.
> i.e. move the dividing line below `Delay for turning off the motors during failsafe`, instead of above it.
> 
> I don't know how many people use LAND procedure. But I often thought it would be nice to have `failsafe_off_delay` / `Delay for turning off the motors during failsafe`tied into the barometer, if the FC had one.
> That way it could shut the motors down according to altitude, and not just a time guesstimate.
> 
> Just out of interest. Whats to become of the `Valid Pulse Range limits`? If total removal is the idea. Some receives like Spektrum work better with this method, than so called NO Pulses RX output.

Hi @Jetrell ,
Now I see you were having same idea about using baro during landing as I've posted recently at inav https://github.com/iNavFlight/inav/issues/7340. Should I move this feature request to inav-configurator?

PR #1016, opened 2020-06-05
Added label that converts cm into m and shows the value

miralem:
Sometimes its confusing to see the cm values for Altitude in Mission Control Tab, that's why I added a label that converts cm to m and shows while the user is typing

meters


DzikuVx:
@miralem could you please resolve the conflicts to merge this PR?

PR #1017, opened 2020-06-08
Launch settings in Advanced tunning tab

pablitous:
It requieres an INAV update also. i will write link in the comments


pablitous:
https://github.com/iNavFlight/inav/pull/5829

fiam:
@pablitous We don't add new MSP message handlers for retrieving settings, the generic settings interface should be used instead. Check e.g. the OSD tab for an example https://github.com/iNavFlight/inav-configurator/blob/a178f0781fa5ea21eef37409d2b033891a6a6d6a/tabs/osd.html#L60 (the `<input>` elements with a `data-setting` attribute. Feel free to ask in the Telegram group if you need further guidance.

pablitous:
thank you guys for the help, i already make it work without new msp messages. I did it just like osd tab

pablitous:
https://github.com/iNavFlight/inav-configurator/pull/1020

Issue #1028, opened 2020-06-24
Fixed Wing Defaults


moggiex:
Howdy,

When you first flash a board & connect, the user is presented with an option to select which type of aircraft they are using, air plane, multi rotor and some other options ( believe this is `platform_type = AIRPLANE` & `applied_defaults = 3` in iNav).

If the option airplane is selected, is it possible to set the following options by default:

### Features

Disable airmode so you see what the actual flight mode is, enable motors, pwm output and turn on auto launch by default.

I appreciate `MOTOR_STOP` enabled by default might cause a slight concern, how many comments have you seen of motors not working? 

```
feature -AIRMODE
feature MOTOR_STOP
feature PWM_OUTPUT_ENABLE
```

Slightly controversi al as a new setting in 2.5, seems to work well even with a light model
`set nav_fw_control_smoothness = 2`

### RTH defaults. 

Climb first is OFF as unlike a quad that would rise straight up vertically, a fixed wing model needs to climb steadily. 

If a RTH condition has been triggered, it is a fair assumption the pilot wants the model to come home, thus turn and return. Climbing off into the distance to increase distance between the pilot and model even further is probably not the most suitable outcome.

`nav_rth_altitude` is set to 10 metres by default. Fine for a quad (probably), fixed wing, 50 metres is a more sensible value

`nav_rth_allow_landing` again, I could appreciate with a quad why this might work, however having a fixed wing model try to land at you by default when the RTH and then loiter is the standard is a bit nuts (ref Kopilot and Vector both return, then loiter).

```
set failsafe_procedure = RTH
set nav_rth_climb_first = OFF
set nav_rth_allow_landing = NEVER
set nav_rth_altitude = 5000
set failsafe_mission = OFF
```

This defaults to 100cm, which is nuts for a fixed wing model considering the typical wing span is 100cm or more! So 30 metres would be more sensible.

`set nav_wp_radius = 3000`

That's all the important ones I've come across, thoughts?

Matt

moggiex:
Update:

**Dominic Douglas** `set failsafe_mission = OFF`
This is to continue a mission/waypoint if RC link is lost. Can't see the logic of not having this off. If someone has their failsafe set to DROP you could start a mission and have the plane drop from the sky once RC link is lost. Most people would assume RC link isn't important for a waypoint.

[...]

PR #1090, opened 2020-10-25
Update defaults_dialog.js

MrD-RC:
Added some of the defaults suggested in issue https://github.com/iNavFlight/inav-configurator/issues/1028

I omitted disabling permanent airmode as the reason for disabling it has been fixed in 2.6 https://github.com/iNavFlight/inav/pull/5711
I omitted enabling motors as servos by default for safety
I omitted setting RTH to turn first, as most prefer climb first, as shown in a poll on INAV Fixed Wing Group

### DIFF
--- a/js/defaults_dialog.js
+++ b/js/defaults_dialog.js
@@ -230,6 +230,30 @@ helper.defaultsDialog = (function() {
                     value: 180
                 },
                 {
+                    key: "nav_fw_control_smoothness",
+                    value: 2
+                },
+                {
+                    key: "failsafe_procedure",
+                    value: "RTH"
+                },
+                {
+                    key: "nav_rth_allow_landing",
+                    value: "NEVER"
+                },
+                {
+                    key: "nav_rth_altitude",
+                    value: 5000
+                },
+                {
+                    key: "failsafe_mission",
+                    value: "OFF"
+                },
+                {
+                    key: "nav_wp_radius",
+                    value: 3000
+                },
+                {
                     key: "platform_type",
                     value: "AIRPLANE"
                 },
--- a/js/defaults_dialog.js
+++ b/js/defaults_dialog.js
@@ -234,12 +234,8 @@ helper.defaultsDialog = (function() {
                     value: 2
                 },
                 {
-                    key: "failsafe_procedure",
-                    value: "RTH"
-                },
-                {
                     key: "nav_rth_allow_landing",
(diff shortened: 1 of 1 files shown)


js/defaults_dialog.js
```diff
@@ -229,6 +229,30 @@ helper.defaultsDialog = (function() {
                     key: "small_angle",
                     value: 180
                 },
+                {
+                    key: "nav_fw_control_smoothness",
+                    value: 2
+                },
+                {
+                    key: "failsafe_procedure",
+                    value: "RTH"
```
DzikuVx, line 238: We should not do it as it requires optional sensors: GPS and Baro

js/defaults_dialog.js
```diff
@@ -229,6 +229,30 @@ helper.defaultsDialog = (function() {
...
+                },
+                {
+                    key: "nav_rth_allow_landing",
+                    value: "NEVER"
```
DzikuVx, line 242: I'd say it's slightly problematic since in case of not-recoverable failsafe model would circle  above takeoff until battery dies and model uncontrollably descend in random direction with random speed

DzikuVx (changes requested):

js/defaults_dialog.js
MrD-RC: That is a fair point.

js/defaults_dialog.js
MrD-RC: This is not really what I have observed. Whenever I hear of this, the model's motor stops, and the plane glides down in the same loiter pattern that it was flying in. Though I know this is not intentional, it does seem to "land" as well as the auto land feature. It's just unpowered.

Another thing to consider is that not all flying sites are suitable for auto land, which requires a large, open space with minimal obstacles. Which is why implementing https://github.com/iNavFlight/inav/issues/6173 would be great.

```suggestion
                    value: "FS_ONLY"
```

DzikuVx: FS_ONLY is a better option indeed

DzikuVx: During brown-out everything is possible. You have no guarantee what dies first and in what state servos will be as well

MrD-RC: Agree, especially with 2S setups. With 3S and above there is usually enough energy to still power the fc and servos. Not always the case though.

DzikuVx (approved):

Issue #963, opened 2020-04-23
Allow re-ordering of modes page


MrD-RC:
Currently the modes page is a bit of a jumble, making it harder to see the set modes up; but even more so help other people with issues.

If the modes could be re-ordered by the user, that would be great. However, if not perhaps group the modes in a more user friendly way. For example:

Arm
Manual
Horizon
Angle
NAV RTH
NAV POSHOLD
NAV CRUISE
NAV ALTHOLD
NAV WP
etc.... 

The most used things closer to the top. Then grouped together appropriately; like flight modes, navigation, tuning, OSD, accessories (for example cam control, LEDs and beeper).

robertha64:
Good idea, even the re-arangement will be very good.
+1

wx4cb:
i mentioned a similar thing a couple years ago that was effectively ignored :D

https://github.com/iNavFlight/inav-configurator/issues/555

rsilk1949:
I would prefer the sort order to be by Channel #.
When I am checking for example a 3 way switch, it is often necessary to scroll the mode page up & down to check the PWM values are set as expected.

MrD-RC:
But surely all the modes on that switch will be the same type of mode?  So this would save scrolling

![image](https://user-images.githubusercontent.com/17590174/99187364-24026080-274e-11eb-85e2-63865c6cfcc2.png)

PR #1100, opened 2020-11-02
Sort modes in to better order 963

MrD-RC:
Sorts the mode order to be more intuitive in Configurator. It groups modes with similiar modes, to make setting and checking the modes easier in Configurator.

The modes are ordered by:
Arm
Flight modes
Nav modes
Flight mode extenders
Navigation
Fixed Wing specific
Multirotor specific
Feedback (buzzer/LEDs)
OSD
Camera
Misc

This is a part fix for issue #963 

Modes, before and after


Jetrell:
@MrD-RC It would be quicker to find what your after, than before. 
The only one that looks out of place is `Failsafe`. I would expect it to fall under the `Nav mode` category, because it essentially works the same as `Nav RTH`. Unless it would break the existing API's?

MrD-RC:
Totally agree @Jetrell. It will be so much quicker and easier to setup and way easier to visualise your modes setup.

Personally, I see `Failsafe` as a mode purely to use to test the failsafe state, without having to power off your transmitter. I'd always recommend that `NAV RTH` is used instead of `Failsafe` as an actual flight mode. I know they can operate differently, for example RTH land only with failsafe. But for my understanding, that was not the intention of the switch. https://github.com/iNavFlight/inav/wiki/Modes#failsafe But I'm open to suggestions :)

The ordering here only effects the modes page. So anything can really be put anywhere.

Jetrell:
> 
> 
> Totally agree @Jetrell. It will be so much quicker and easier to setup and way easier to visualise your modes setup.
> 
> Personally, I see `Failsafe` as a mode purely to use to test the failsafe state, without having to power off your transmitter. I'd always recommend that `NAV RTH` is used instead of `Failsafe` as an actual flight mode. I know they can operate differently, for example RTH land only with failsafe. 

Not knocking you view on the matter of RTH.
I once used `Nav RTH mode` for return to home. But after testing over the years. I found that using `Failsafe mode` for return to home, was touching all bases with no side effects.

 * It tests if signal loss Failsafe is working. 
 * it can trigger landing mode with multi-rotors. 
 * It can be triggered by a switch to initiate RTH.

avsaase:
@DzikuVx WDYT?

DzikuVx:
Why not. Only one question: what happens if a mode is not configured in the Configurator? Like: FC adds a new mode, where would it display? @MrD-RC

PR #1116, opened 2020-11-19
Draggable mixer wizard

yngndrw:
A picture is a thousand words:
image

Still limited to the "Quad X" preset as the existing mixer wizard is, but the implementation is now slightly closer to supporting other presets.

[JsRender](https://www.jsviews.com/#jsrapi) has been added as a templating engine.


yngndrw:
Tagging @DzikuVx as he's the original developer of the mixer wizard

sensei-hacker:
Good idea, superseded by the newer work.

PR #1117, opened 2020-11-21
OSD Element: VTX Power in mW.

Scavanger:
Corresponding PR to https://github.com/iNavFlight/inav/pull/6326


Jetrell:
@Scavanger and chance of adding the `mW` glyph -

Capture
to the right side of the power level.

PR #1123, opened 2020-11-28
Added microsecond value to motor outputs

### LINKED ISSUE #1122: Add microseconds reading back to the Outputs page
## Problem
At some point (2.5 I believe), the uS values for the motor sliders were changed for throttle percentage. I guess this is because the motor idle stuff changed, and now uses percent. This is fine for quads, but it removes a useful feature for fixed wing pilots wanting to set the launch throttle, which is in uS, not %.
## Solution
Display both the throttle percentage and the command in uS under the throttle sliders.

MrD-RC:
I added the microsecond value back on to the motor outputs, where currently a percentage is shown. This is to help with configuring auto launch on fixed wings. 3D motors only shows microseconds, as in 2.5.

Fixes #1122 

image

### DIFF
--- a/tabs/outputs.js
+++ b/tabs/outputs.js
@@ -80,16 +80,22 @@ TABS.outputs.initialize = function (callback) {
         finalize();
     }
 
-    function getMotorOutputValue(value) {
+    function getMotorOutputValue(value, isSlider) {
+        var outputValue = "";
 
         if (!self.feature3DEnabled) {
             let valueNormalized = value - MISC.mincommand;
             let maxThrottleNormalized = MISC.maxthrottle - MISC.mincommand;
 
-            return Math.round(100 * valueNormalized / maxThrottleNormalized) + "%";
+            outputValue = Math.round(100 * valueNormalized / maxThrottleNormalized) + "%";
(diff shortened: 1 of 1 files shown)


DzikuVx:
@MrD-RC the problem is that us is not really what is outputted to the motors...  It is for standard, but only for that. And it's especially misleading for DSHOT
The correct approach would be to convert throttle settings in NAV and Launch to %, not the other way around.
We got rid of so many problems and questions after we removed us from Outputs tab that I really do not wanto to go back

MrD-RC:
Fair enough @DzikuVx that makes sense. It would be good to get sorted though, as I am getting questions about where to get the uS setting to set the launch throttle.

MrD-RC:
**House keeping**
Closed as it won’t be getting merged.

avsaase:
Hoe about we change the fixed wing throttle settings to percentages instead, like @DzikuVx suggested?

DzikuVx:
That would be cool but rather not for 3.0

niedz., 18 kwi 2021, 10:12 użytkownik Alexander van Saase <
***@***.***> napisał:

PR #1142, opened 2020-12-23
Update FW presets

DzikuVx:
- lower gyro LPF
- enable Matrix filter as it's reported to improve performance
- enable `imu_acc_ignore_rate` by default


Jetrell:
@DzikuVx what do you think about adding - `mc_airmode_type = STICK_CENTER_ONCE` as well? I found it does help over come the I-term windup issue on launch.

DzikuVx:
@Jetrell I'd like to  test it before that. Still had no chance to do it

PR #1152, opened 2021-01-03
Reorganize voltage and current sensors and battery settings

### LINKED ISSUE #1144: Reorganize battery settings on Configuration tab
I think the organization of the battery settings in the Configuration tab is a bit confusing. In the screenshot below, the items indicated in red represent settings that pertain how INAV measures battery voltage and current, which stays the same regardless of the battery profile. The blue items are about the specific battery used and change with battery profile.

2020-12-26 14_42_39-INAV Configurator

It would be more intuitive if the blue and red items were grouped together. Something like this: 
Battery settings mockup

WDYT?

avsaase:
Closes #1144, which turned out to be much easier than I thought it would be. 

There is one minor alignment issue that I cannot figure out. I used the exact same code as the board alignment box which also has a help tooltip, but for some reason here it causes problems. Any ideas? 

2021-01-03 14_06_49-INAV Configurator

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -698,8 +698,8 @@
     "configurationThrottleMinimumCommand": {
         "message": "Minimum Command"
     },
-    "configurationBatteryVoltage": {
-        "message": "Battery Voltage"
+    "configurationVoltageCurrentSensor": {
+        "message": "Voltage and Current Sensors"
     },
     "configurationBatteryCurrent": {
         "message": "Battery Current"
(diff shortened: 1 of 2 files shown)


avsaase:
There is still the alignment issue in the screenshot.

DzikuVx:
We can fix it separately. No big deal

czw., 7 sty 2021, 16:57 użytkownik Alexander van Saase <
notifications@github.com> napisał:

PR #1170, opened 2021-02-06
Set default Airmode type to STICK_CENTER_ONCE for fixed wing

MrD-RC:
Added default airmode type to be STICK_CENTER_ONCE, which makes more sense for fixed wing. This will need updating if a PR I'm currently working on for the INAV firmware goes through. Works with the current setup.


MrD-RC:
Not required, as included with #1171

PR #1171, opened 2021-02-06
Removed mc_ from airmode settings

MrD-RC:
This change is to partner PR https://github.com/iNavFlight/inav/pull/6578 in the INAV firmware. It removes the mc_ prefix from the airmode CLI variables, as they are confusing, as there are now options that can also apply to fixed wing.

This change covers the default dialog, presets, and PID/Mechanics pages. I could not find any other instances of mc_airmode_x in the configurator.


DzikuVx:
Thanks a lot!

PR #1172, opened 2021-02-07
Increased FW loiter radius

MrD-RC:
We are seeing a lot of issues with crashes due to partially too tight a loiter radius. I have increased this value to have a safer default. It should of course still be tuned to the aircraft, but this will hopefully cause less issues to those who forget.


MrD-RC:
Sorry, wrong value and place.

Issue #1184, opened 2021-03-01
Remove Plane Presets and new Fixed Wing defaults based on Plane type


b14ckyy:
After some thinking and a lot of tuning help with the same problems over and over again, I had the idea of completely revised INAV defaults for fixed wing. 

So my Idea: Completely remove the Flying Wing Presets from configurator. Community engagement on this was close to zero over the years anyway and it caused more trouble than it helped. Just keep quad presets. 

During first initialization of the FC, when user is asked for craft type, we should have 2 options: Airplane (with Tail) + Flying Wing (No Tail). The current defaults are way too agressive for many planes out there, especially on wingspans above 1.5m. It is much more appropriate to have very gentle and low values, than too agressive ones for a maiden. 

Then the Initial Defaults should be set as following: 

### For both: 

platform_type = AIRPLANE
gyro_hardware_lpf = 256HZ
set gyro_lpf_hz =  25
gyro_lpf_type = BIQUAD
dynamic_gyro_notch_enabled = ON
dynamic_gyro_notch_q = 250
dynamic_gyro_notch_min_hz = 30
motor_pwm_protocol = STANDARD
rc_yaw_expo = 30
rc_expo = 30
roll_rate = 18
pitch_rate = 9
yaw_rate = 3
nav_fw_pos_z_p = 20
nav_fw_pos_z_d = 5
nav_fw_pos_xy_p = 50
small_angle=180
nav_rth_allow_landing = FS_ONLY
nav_rth_altitude = 5000
nav_wp_radius = 1500
throttle_idle =  5.000
applied_defaults = 3
imu_acc_ignore_rate = 10
airmode_type =  STICK_CENTER_ONCE
nav_rth_climb_first = OFF

### For Airplane with Tail Specific:

fw_turn_assist_pitch_gain =  0.5 //Default 1.0 is too high for most planes I have seen so far and attitude controller fights with it in navigation turns
max_angle_inclination_rll = 35 //this is the default navigation inclination for roll too
nav_fw_bank_angle = 35

fw_p_pitch = 15
fw_i_pitch = 10
fw_ff_pitch = 60
fw_p_roll = 10
fw_i_roll = 8
fw_ff_roll = 40
roll_rate = 18
pitch_rate = 9
fw_p_yaw = 20
fw_i_yaw = 0
fw_ff_yaw = 100

### For Airplanes with No Tail (Wings, Drak style, etc):

[...]

PR #1188, opened 2021-03-04
New fixed Wing generic presets

b14ckyy:
Removed the outdated Fixed wing model specific presets and created generic presets for Tailed and Tail-less Planes. 
These are tested on different planes already from 680mm size up to 1.6m and bigger planes tested soon. Stable flight and RTH behavior as well as position hold for basic flight and safe maiden.

### DIFF
--- a/js/defaults_dialog.js
+++ b/js/defaults_dialog.js
@@ -190,122 +190,165 @@ helper.defaultsDialog = (function() {
             "reboot": true,
             "settings": [
                 { 
-					key: "platform_type"
-					value: "AIRPLANE"
-				},
-				{
-                    key: "gyro_hardware_lpf",
-                    value: "256HZ"
+			key: "platform_type"
+			value: "AIRPLANE"
+		},
+		{
+                	key: "gyro_hardware_lpf",
+                    	value: "256HZ"
                 },
                 { 
-					key: "gyro_lpf_hz"
-					value: "25"
-				},
-				{ 
-					key: "gyro_lpf_type"
-					value: "BIQUAD"
-				},
-				{ 
-					key: "dynamic_gyro_notch_enabled"
-					value: "ON"
-				},
-				{ 
-					key: "dynamic_gyro_notch_q"
-					value: "250"
-				},
-				{ 
-					key: "dynamic_gyro_notch_min_hz"
-					value: "30"
-				},
-				{ 
-					key: "motor_pwm_protocol"
-					value: "STANDARD"
-				},
-				{ 
-					key: "rc_yaw_expo"
-					value: "30"
-				},
-				{ 
-					key: "rc_expo"
-					value: "30"
-				},
-				{ 
-					key: "roll_rate"
-					value: "18"
-				},
-				{ 
-					key: "pitch_rate"
-					value: "9"
-				},
-				{ 
-					key: "yaw_rate"
-					value: "3"
-				},
-				{ 
-					key: "nav_fw_pos_z_p"
-					value: "20"
-				},
-				{ 
-					key: "nav_fw_pos_z_d"
-					value: "5"
-				},
-				{ 
-					key: "nav_fw_pos_xy_p"
-					value: "50"
-				},
-				{ 
-					key: "small_angle"
-					value: "180"
-				},
-				{ 
-					key: "nav_rth_allow_landing"
-					value: "FS_ONLY"
-				},
-				{ 
-					key: "nav_rth_altitude"
-					value: "5000"
-				},
-				{ 
-					key: "nav_wp_radius"
-					value: "1500"
-				},
-				{ 
-					key: "throttle_idle"
-					value: "5.000"
-				},
-				{ 
-					key: "applied_defaults"
-					value: "3"
-				},
-				{ 
-					key: "imu_acc_ignore_rate"
-					value: "10"
-				},
-				{ 
-					key: "airmode_type"
-					value: "STICK_CENTER_ONCE"
-				},
-				{ 
-					key: "airmode_type"
-					value: "STICK_CENTER_ONCE"
-				},
-				{ 
-					key: "nav_rth_climb_first"
-					value: "OFF"
-				},
-				{ 
-					key: "fw_turn_assist_pitch_gain"
-					value: "0.5"
-				},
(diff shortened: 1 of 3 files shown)


b14ckyy:
fixes #1184

avsaase:
Many of these defaults are also hard coded into the firmware and listed in the cli settings docs. I think they should be changed there as well. At least if it concerns fixed wing specific settings that apply to both presets.

js/defaults_dialog.js
```diff
@@ -255,36 +335,186 @@ helper.defaultsDialog = (function() {
                 },
                 {
                     key: "failsafe_mission",
-                    value: "OFF"
+                    value: "ON"
                 },
                 {
                     key: "nav_wp_radius",
-                    value: 3000
+                    value: 1500
```
avsaase, line 342: This is quite a small radius for a fixed wing. If there are strong winds then the WP can easily be missed leading to a sort of loiter around the WP.

b14ckyy:
@avsaase I agree on FW specific values that are common in both presets but we should keep them as selectable too. 
I added the settings to the defaults dialog and also to the presets page so people can upgrade to 2.7 with a diff and then apply the defaults just for theior plane type and tune from scratch. 

Otherwise people won't get the new defaults anyway if these values where changed before.

js/defaults_dialog.js
b14ckyy: default is 1m in the past. Pawel raised it to 30m but IMHO that was too much. I usually set my planes all to 10m. 15m should be a good compromise. Kepp in mind that we don't use compass on planes and inav compensates or wind. A radius of 15m is still a 30m wide target to hit the waypoint. Never had any failed waypoint mission even with 10m before.

b14ckyy: oh and 30m can be too big for aerial mapping missions when you want to fully cover ground. it makes a big difference if the pathes can be offset by 60m.

avsaase: My bad, I misinterpreted this setting as 1.5 meter. I think 10-15 meters would be fine.

js/preset_definitions.js
```diff
@@ -1157,489 +1157,328 @@ presets.presets = [
...
-            presets.elementHelper("RC_tuning", "roll_rate", 360),
-            presets.elementHelper("RC_tuning", "pitch_rate", 130),
-            presets.elementHelper("RC_tuning", "dynamic_THR_PID", 30),
-            presets.elementHelper("RC_tuning", "dynamic_THR_breakpoint", 1500)
-        ],
+        settingsMSP: [],
+		type: 'flyingwing',
         settings: [
-            {
-                key: "gyro_hardware_lpf",
```
DzikuVx, line 1492: @b14ckyy actually I would leave this kind of settings in place, not remove them

DzikuVx (changes requested):

js/defaults_dialog.js
```diff
@@ -227,15 +239,83 @@ helper.defaultsDialog = (function() {
...
+                    key: "fw_ff_roll",
+                    value: 40
+		},
+		{ 
+                    key: "fw_p_yaw",
+                    value: 20
+		},
+		{ 
+                    key: "fw_i_yaw",
+                    value: 0
```
avsaase, line 306: I don't agree with disabling I-term stabilization on yaw as a solution to the rudder counteracting turns. https://github.com/iNavFlight/inav/pull/6387 is a better solution to that so I propose adding `fw_yaw_iterm_freeze_bank_angle = 10` as a default instead.

b14ckyy: @avsaase this is only for the initial preset. It will be tuned with autotune later. and it is hard to determine a good I term as the rudder size and authority can massively differ from plane to plane. I wanna avoid a yaw spin or stall due to excessive rudder deflection after a bank turn on planes like a tundra or Extra.

js/preset_definitions.js
b14ckyy: @DzikuVx it is still there in all presets. It just moved position as i have rewritten the whole list. that's why it shows as removed.

js/defaults_dialog.js
avsaase: But Yaw I helps with trimming the plane, which you typically do before autotune.

b14ckyy: @avsaase hmm actually a good point. maybe a value of 5 is okay as a starting point as yaw trim is usually only minimal as ti can be visually aligned well.

js/defaults_dialog.js
```diff
@@ -255,36 +335,186 @@ helper.defaultsDialog = (function() {
...
+                    key: "nav_fw_pos_z_d",
+                    value: 5
+		},
+		{ 
+                    key: "nav_fw_pos_xy_p",
+                    value: 50
+		},
+		{ 
+                    key: "fw_turn_assist_pitch_gain",
+                    value: 0.2
```
avsaase, line 432: I only have one plane without a tail, but it needs 0.1 to get smooth turns. In general, I think it is best to err on the low side and lose some altitude that the altitude PIDs will have to correct, then to get this very bumpy pitch behavior during turns when this setting is set too high.

b14ckyy: @avsaase I tested 0.2 on the Raphax and it was slightly low (dipping nose in turns), it's perfect on the AR Pro and a tad too high on the 250G. I think 0.2 should be a good starting value in general. It also depends on the max navigation bank angle how much gain is needed. so it will be hard to find a perfect value. maybe this should go into the advanced tuning tab in configurator?

js/preset_definitions.js
b14ckyy: as discussed in Discord. The Preset should allow people to apply the default settings to planes after applying their DIFF. And then tune from there with the new autotune feature. This should go into the Update steps of the release notes.

avsaase:
@b14ckyy can you add `dterm_lpf_hz = 10` or `15` for both fixed wing presets. I think for now D-gain default should stay zero but if people want to experiment with positive D-gain then having a bit more filtering on the D-term should reduce the wear and tear of the servos.

b14ckyy:
@avsaase check and done.

PR #1190, opened 2021-03-05
Remove expo from Rates & Expo tab

MrD-RC:
Expo has been duplicated in multiple locations. This adds confusion and brings up questions as to where it should be changed. To simplify things, I recommend removing expo from the _PID Tuning_ page. It makes more sense to have it on the _Receiver_ page as:

- There are more expo settings already on the _Receiver_ page (**RC Expo**, **RC Expo Yaw**, **Manual RC Expo**, **Manual RC Expo Yaw**, and **Throttle Expo**), compared to only RC Expo (**Roll & Pitch Expo**) and RC Expo Yaw (**Yaw Expo**) on the _Rates & Expo_ page.
- Expo is something traditionally related to a feature on the transmitter, so it makes more sense to have it with the receiver settings.

To lessen confusion, the tab name has also been renamed from **Rates & Expo** to simply **Rates**.

image
The new **Rates** page

image
The **Receiver** page for reference

### DIFF
--- a/tabs/pid_tuning.html
+++ b/tabs/pid_tuning.html
@@ -4,7 +4,7 @@
     <div class="content_wrapper">
         <div class="tab_title subtab__header">
             <span class="subtab__header_label subtab__header_label--current" for="subtab-pid">PID gains</span>
-            <span class="subtab__header_label" for="subtab-rates">Rates & Expo</span>
+            <span class="subtab__header_label" for="subtab-rates">Rates</span>
             <span class="subtab__header_label" for="subtab-filters">Filters</span>
             <span class="subtab__header_label" for="subtab-mechanics">Mechanics</span>
         </div>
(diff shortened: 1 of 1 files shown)


DzikuVx:
@MrD-RC it would make more sense to remove EXPO and Rates from Receiver and move it to PID Tuning.
To think about it, Expo and rates are not really about the receiver but about stabilization

MrD-RC:
I disagree. I don’t associate expo with stabilsation at all. I associate it with relative surface outputs to control stick inputs. The places that make most sense for me for expo is the receiver page. Traditionally, expo would be set on inputs or in the mixer. In iNav, I see the receiver page as inputs. The mixer page is already pretty full. 

PID Tuning seems like the wrong place to me. Especially when there are manual expo settings. They have nothing to do with stabilisation at all.

Mateyhv:
I do agree with Pawel, expo is not related to receiver settings at all. From logic I would search for expo next to Rates when I am setting up a model. But I do agree that the rates tab need to be reorganized to make it more clear way where everything related to roll is grouped in one place, the same for rates, etc.

PR #1196, opened 2021-03-11
Add modulus operator support

MingchenZhang:
Add configurator side of https://github.com/iNavFlight/inav/pull/6699

### DIFF
--- a/js/fc.js
+++ b/js/fc.js
@@ -1113,106 +1113,111 @@ var FC = {
                 output: "raw"
             },
             18: {
+                name: "MOD",
+                hasOperand: [true, true],
+                output: "raw"
+            },
+            19: {
                 name: "GVAR SET",
                 hasOperand: [true, true],
                 output: "none"
             },
-            19: {
+            20: {
                 name: "GVAR INC",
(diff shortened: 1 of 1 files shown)


js/fc.js
```diff
@@ -1113,106 +1113,111 @@ var FC = {
                 output: "raw"
             },
             18: {
+                name: "MOD",
```
DzikuVx, line 1116: Please add it a LAST entry not to break compatibility with existing solutions

DzikuVx (changes requested):

js/fc.js
MingchenZhang: I just moved the mod to the last entry.

DzikuVx (approved):

avsaase:
Is it not possible to keep the ordering as originally intended by @MingchenZhang? It is kind of annoying if new operators can only be added at the end of the dropdown menu to avoid breaking compatibility with old diffs.

DzikuVx:
@avsaase one point in a future we should do updated UI for the operands with grouping and ordering. But not now

PR #1212, opened 2021-03-29
Re-arrange advanced page

### LINKED ISSUE #1207: Setting minimum satellites in the UI
Hey guys, just wondering if this is something we should look at removing?

image

It's something that should rarely be used. And people in the situation where they want to, can still set it in the CLI. I just see it as a temptation that could cause people who don't know what they're doing problems. Then I guess the same could be said for the whole position estimator block. However, the number of satellites for a lock, is something that is easier for people to understand; so some are tempted to play. They don't realise that below 5 3D locks the data starts getting bad.

Perhaps we could also move the entire position estimator block to the bottom of the page, just to keep it out of immediate sight.

I'm happy to do the work. I'm just looking for thoughts from you guys. @DzikuVx @digitalentity @avsaase Anyone else?

MrD-RC:
Re-arranged the advanced tuning page. This was mainly to move the position estimator to the bottom of the page. Out of sight, out of mind for those who don't need to use it. I also took the opportunity to put the fixed wing and multirotor areas together.

image
image
image

Fixes #1207

### DIFF
--- a/tabs/advanced_tuning.html
+++ b/tabs/advanced_tuning.html
@@ -5,315 +5,6 @@
         <div class="leftWrapper">
             <div class="config-section gui_box grey">
                 <div class="gui_box_titlebar">
-                    <div class="spacer_box_title" data-i18n="multiRotorNavigationConfiguration"></div>
-                </div>
-                <div class="spacer_box">
-                    <div class="select">
-                        <select id="user-control-mode"></select>
-                        <label for="user-control-mode"> <span data-i18n="userControlMode"></span></label>
-                    </div>
-                    <div class="number">
-                        <input id="max-speed" type="number" data-simple-bind="NAV_POSHOLD.maxSpeed" step="1" min="10" max="2000">
-                        <label for="max-speed">
-                            <span data-i18n="posholdMaxSpeed"></span>
-                        </label>
-                    </div>
-                    <div class="number">
-                        <input id="max-manual-speed" type="number" data-simple-bind="NAV_POSHOLD.maxManualSpeed" step="1" min="10" max="2000">
-                        <label for="max-manual-speed">
-                            <span data-i18n="posholdMaxManualSpeed"></span>
-                        </label>
-                    </div>
-                    <div class="number">
-                        <input id="max-climb-rate" type="number" data-simple-bind="NAV_POSHOLD.maxClimbRate" step="1" min="10" max="2000">
-                        <label for="max-climb-rate">
-                            <span data-i18n="posholdMaxClimbRate"></span>
-                        </label>
-                    </div>
-                    <div class="number">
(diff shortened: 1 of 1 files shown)


rts18:
@MrD-RC Good a idea.
Can I suggest that you add `RTH and landing settings `to the top of the right column?  

I recently tested multi-rotor braking mode and wrote a write-up on RCG iNav blog. In one of dozens of tests I performed, it did not brake; instead it accelerated away until I disarmed it.
This was a known bug from its alpha testing days, that may not get fixed anytime soon.  

It may be better to have that feature at the bottom of the right column next to position estimator.

rts18:
@MrD-RC Good idea.
Can I suggest that you add RTH and landing settings to the top of the right column?

I recently tested multi-rotor braking mode and wrote a write-up on RCG iNav blog. In one of dozens of tests I performed, it did not brake; instead it accelerated away until I disarmed it.
This was a known bug from its alpha testing days, that may not get fixed anytime soon.

It may be better to have that feature at the bottom of the right column next to position estimator.

flyinginger:
Would be great to add opflow and lidar concerned parameters
Cheers!

MrD-RC:
@rts18 That sounds like a very good point. Maybe @DzikuVx could shed some light on that bug. It could be worth adding an **experimental** warning to the braking section if that is the case. 

I’m not keen on moving RTH above multirotor, as I like how the page is balanced at the moment. With equal status for both fixed wings and multirotors. But I can look at moving the multirotor braking section down, level with the position estimator. Lets see what Paweł says with regards to the bug.

MrD-RC:
@flyinginger maybe raise this as an issue with details of which parts you would like added and who would benefit.

MrD-RC:
@rts18 I figure this would be a good layout if the multirotor braking is to move to the bottom. I've not uploaded this yet, as I'd still like to hear from @DzikuVx with regards to the braking bug, the status of it, and whether it may be a good idea to include a warning. Again, sorry for the multiple screen shots.

image
image
image

MrD-RC:
> @MrD-RC Good idea.
> Can I suggest that you add RTH and landing settings to the top of the right column?
> 
> I recently tested multi-rotor braking mode and wrote a write-up on RCG iNav blog. In one of dozens of tests I performed, it did not brake; instead it accelerated away until I disarmed it.
> This was a known bug from its alpha testing days, that may not get fixed anytime soon.
> 
> It may be better to have that feature at the bottom of the right column next to position estimator.

Have you raised an issue about this in the inav firmware github? I’ve chatted with Paweł and he knows nothing about this bug. He says it shouldn’t be possible as the braking has a 2 second timeout.

rts18:
@MrD-RC I didn't open a Github issue. Because I only have DVR footage and no logs of the incident. I'm not sure if that would be conclusive enough.

PR #1217, opened 2021-04-01
Allow editing of derivative on airplane

MrD-RC:
Fix for the original change. Now correctly shows FeedForward and not Control Derivative.


avsaase:
Tested and works correctly now so I'll merge this. Thanks.

PR #1227, opened 2021-04-11
Added the new Prearm to the mode ordering

MrD-RC:
Added **PREARM** to the mode ordering, so that it grouped with **ARM**.


avsaase:
NAV CRUISE has been renamed to NAV COURSE HOLD (https://github.com/iNavFlight/inav/pull/6297) so it's now shown at the bottom of the modes list. I was about to make a PR for this but maybe you can include it in this one.

MrD-RC:
I knew there was something I forgot!

Is cruise still there for what was previously 3D cruise? It's just that adding alt hold to course hold to make cruise sounds a bit disjointed.

MrD-RC:
I've just had a look at the PR and that doesn't look to be the case. In which case, I think the current change of cruise will be confusing. It would be better to have 2 box modes, one for NAV CRUISE, and another for NAv COURSE HOLD. Otherwise, I think it worked better the old way. Much less confusing.

avsaase:
Well I've always found it a bit confusing to have to enable two modes I don't want to use to get one that I do. But once you accept that I think the new naming scheme makes sense. Course hold holds the course, and alt hold holds the altitude. Put them together and you can sit back and cruise. Having said that, we could add a separate cruise box mode that under the cover enables course hold and alt hold.

MrD-RC:
I think that makes more sense, to have a separate box mode. Cruise to 3D Cruise made sense. Course Hold to Cruise really has no flow. For someone new to iNav, it would be confusing.

MrD-RC:
I've added NAV COURSE HOLD. But I've kept NAV CRUISE in the hope that it is added. I don't think how it is at the moment with COURSE HOLD + ALT HOLD = CRUISE is good at all to be honest. Very confusing for new pilots.

PR #1236, opened 2021-04-20
Add kW symbol to fonts

avsaase:
Font support for https://github.com/iNavFlight/inav/pull/6796. Symbols are based on the kilometer and watt symbols for font consistency. I'm not the best at artistic stuff so I'm happy to take suggestions to make them look better.

bold: 
clarity: 
clarity_medium: 
default: 
impact:


avsaase:
impact_mini: 
large: 
vision:

avsaase:
Forgot to add that I get an invalid font warning with these fonts but they otherwise work fine. Do I need to add some metadata to the font file somehow?

avsaase:
@shellixyz a i understand it, @Jetrell will send you updated fonts that also includes new kW symbols (as well as Ah/km and Ah/mi symbols). This PR is then no longer needed.

PR #1243, opened 2021-04-24
Improve autolanding settings layout and documentation

shellixyz:


avsaase:
I have to say I think the bold text doesn't look very nice and also inconsistent with the rest of the configurator text formatting. I think the explanation under the tooltip is sufficient to remove the second sentence from the setting name and change the first to normal text.

PR #1258, opened 2021-05-06
Disable fixedwing dboost

avsaase:


DzikuVx:
@avsaase conflicts please ;)

avsaase:
@DzikuVx will fix them tonight

rts18:
@avsaase @DzikuVx I agree with the removal of multirotor D-term boost for a fixedwing. Because its stick position based
Most people will run lower PID gains on a fixedwing, to prevent oscillations over a broad speed range. Rather than tune them higher and only cruise around.

I can't help wonder if a fixedwing version of D-term boost would work, by boosting D-term at lower throttle positions, when higher than average accelerometer rotation rates are detected.  What's your view on this?

avsaase:
For now we want to keep the D-term handling on fixed wing as simple as possible, at least untill we have a better understanding of how it affects flight performance. That said, TPA also scales the D-gain according to throttle level.

rts18:
> 
> For now we want to keep the D-term handling on fixed wing as simple as possible, at least untill we have a better understanding of how it affects flight performance. 

I understand you point.



>
>That said, TPA also scales the D-gain according to throttle level.

I find TPA rather limited for fixed wings. I believe a lot of this is due to the different scaling of the P and I, compared to the D-term.
I've tried using higher gain settings, with a TPA off 95%. And its usefulness is average at best.

rts18:
> 
> For now we want to keep the D-term handling on fixed wing as simple as possible, at least untill we have a better understanding of how it affects flight performance. 

I understand your point.



>
>That said, TPA also scales the D-gain according to throttle level.

I find TPA rather limited for fixed wings. I believe a lot of this is due to the different scaling of the P and I, compared to the D-term.
I've tried using higher gain settings, with a TPA of 95%. And its usefulness is average at best.

PR #1264, opened 2021-05-12
Add new cruise mode to modes ordering

MrD-RC:
I thought I left Cruise in there. It wasn't, so it's back now we have a Cruise mode again.


Jetrell:
@MrD-RC Does this PR also prevent NavCruise and Failsafe modes from highlighting blue when not active?

modes

MrD-RC:
@Jetrell at first glance, that doesn't appear to be caused by the ordering. I removed the ordering code, and both remain highlighted.
Modes

I added the IDs as debug, and the ID matches the original ID, so no ordering has taken place.
modes 2
...and the ordered version for comparison
modes 3
The highlighting is based on the original ID.

I also removed every range except for those 2, and they stay highlighted, no matter what channel I select, or where I move the ranges. I think this is something in the firmware causing it.

Jetrell:
@MrD-RC That is odd. Thank you for looking at it. Maybe it can be fixed at some later time.  
For now its good to have the two Cruise modes together. :+1:

PR #1265, opened 2021-05-14
CRSF RSSI alarm

avsaase:
Configurator support for https://github.com/iNavFlight/inav/pull/6960.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -2804,7 +2804,7 @@
         "message": "CRSF RSSI dBm Alarm"
     },
     "osd_rssi_dbm_alarm_HELP": {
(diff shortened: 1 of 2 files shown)


OptimusTi:
Is there going to be updated fonts for the new changes?

avsaase:
No, I decided to keep the symbols as they are. I kept going back and forth between having the RSSI and dBm symbol separate and combined, so I thought let's not change it any more for now.

OptimusTi:
Understood. Will spacing remain the same as 2.6 so elements align correctly?

avsaase:
For 3.0 it looks like in https://github.com/iNavFlight/inav/pull/6950#issuecomment-840098629.

PR #1267, opened 2021-05-16
Mission distance + few MP stuffs correction

ArnoTlse:
Update of Mission planner tab with :

- correction on Mission distance computation to take into account Jump loop
- Add a field to display mission filename loaded/saved
- Add option to switch from TAKE-OFF altitude reference to WGS84 ellispoid reference (see [INAV #6662](https://github.com/iNavFlight/inav/pull/6662)
- ALT & SPEED values for a Waypoint are saved directly when changed to be consistent with all the other fields.

### DIFF
--- a/js/waypointCollection.js
+++ b/js/waypointCollection.js
@@ -341,7 +341,7 @@ let WaypointCollection = function () {
         let n = 0 ;
         let startCount = true;
         while (startCount && (nLoop!=-1)) {
-            if (nStart > data[data.length -1].getNumber()) {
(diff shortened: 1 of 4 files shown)


ArnoTlse:
image

rts18:
@ArnoTlse Is `wait time(s)` added to `PH_TIME` in this update? Beside it only appearing if the mission is enter via the CLI.

PH_time

ArnoTlse:
Hi, 

yes, `wait time(s)` is appearing when you select PH_TIME in this update.

image


regards

Arno

PR #1271, opened 2021-05-21
preset reorder and default optimization

b14ckyy:
Changed some of the Plane preset values after more testing to be more reliable
also restructured preset code with help of @gailingmic to be easier to edit in the future


b14ckyy:
don't merge yet. Seems there is a bug as the GUI is not loading completely.

b14ckyy:
build error is fixed. Just a last bug that prevents Preset page from loading. Working on that.

b14ckyy:
All fixed and Presets working now with new order. Ready to merge.

js/default_presets/airplane_tail.js
```diff
@@ -0,0 +1,178 @@
+window.default_preset_airplane_tail = {
```
DzikuVx, line 1: Idea is great, but let's do it in a much better way!
1. No more assigning anything to `window` object, this is so 90s
2. Let's store definitions in an array and push new elements as we go

so it would look like

```
var presetDefinitions = presetDefinitions || [];

presetDefinitions.push({
//here comes the preset object
})

```

DzikuVx (changes requested):

js/default_presets/airplane_tail.js
b14ckyy: I think @michellgai tried this first when we did this but it did not work out. I don't know the exact reason. But I'll ask him, maybe I'm wrong here.

DzikuVx:
@b14ckyy time to resolve conflicts and merge I suppose :)

DzikuVx:
blocked as #1327

PR #1272, opened 2021-05-22
Updated NW.js and changed to SDK Version

b14ckyy:
current NW.js caused the non-SDK version of NW to be loaded so no debugging was possible. Michell changed this to a strict NW version to prohibit his.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "jquery-ui-npm": "1.12.0",
     "marked": "^0.3.17",
     "minimist": "^1.2.0",
-    "nw": "^0.50.3-sdk",
+    "nw": "0.53.1-sdk",
     "nw-dialog": "^1.0.7",
     "openlayers": "^4.6.5",
     "temp": "^0.8.3",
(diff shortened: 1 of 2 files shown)


DzikuVx:
To debug you should install NW in SDK in global scope or override locally after `npm install`

` npm install nw@0.50.3-sdk --nwjs_build_type=sdk`

And I do not want to update NW.js between RC and final. It always breaks stuff

package.json
```diff
@@ -35,7 +35,7 @@
     "jquery-ui-npm": "1.12.0",
     "marked": "^0.3.17",
     "minimist": "^1.2.0",
-    "nw": "^0.50.3-sdk",
+    "nw": "0.53.1-sdk",
```
DzikuVx, line 38: TBH we do not need sdk flavor of nw and this might go only for 3.1
No major component upgrade between RC

DzikuVx: And see my comment under the PR

DzikuVx (changes requested):

package.json
b14ckyy: Got it. In this case I will close the PR

b14ckyy:
closing as mentioned above

PR #1298, opened 2021-06-13
Update STM32DFU procedure for H7 support

DzikuVx:
Port of betaflight/betaflight-configurator

@bkleiner this seems to fix H7 flashing problems. Care to check?

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -160,6 +160,88 @@
         "message": "Are <strong>udev rules</strong> installed correctly? See docs for instructions"
     },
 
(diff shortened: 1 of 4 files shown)


bkleiner:
The flash/erase issue seem indeed fixed, however i get an error if i try to flash a board not already in DFU mode.

```
script.js:23816 Uncaught TypeError: Error in invocation of serial.send(integer connectionId, binary data, function callback): No matching signature.
    at send (script.js:23816)
    at Object.send (script.js:23849)
    at STM32_protocol.send (script.js:27807)
    at script.js:27936
    at script.js:19058
```

PR #1299, opened 2021-06-15
[Mission Planner] Elevation Profile & Home

ArnoTlse:
PR to add :

- Elevation profile plot : display Earth elevation under mission + Mission elevation with regards to reference defined by user (relative to Take-off home or to MSL)
- Home (Take-off location defined by user) icon + box to define an Home location
- ability to switch Elevation reference from Take-Off to MSL + display into WP under edition the value of the Earth elevation beneath to help settings the altitude in MSL
- add arrow to indicate sens for Jumps
- Add number of WP in the WP edition Box

image

image

image

image

image

image


breadoven:
@ArnoTlse. Awesome bit of work Arno, not sure why it hasn't been merged yet.

One thing that did occur to me though with the Sea Level Ref option as it is now is that it could be a bit dangerous to use if you didn't entirely understand what it does. This is because when SL Ref is selected ON it doesn't change the previously set WP altitude to account for the ground elevation which means the set altitude could be lower than ground elevation ... the craft will fly into the ground, e.g. if Alt is set to 50m and SL Ref is selected ON at a location with a ground elevation of 200m the WP will save with P3 set to 1 and an Alt setting of 50m ... which means the craft will try and fly to an altitude 150m below ground level.

I was thinking it would be better to set the altitude to the ground elevation + the previously set WP altitude when SL Ref is selected ON which ensures you will be clear of the ground, i.e. in the above example the Alt will change from 50m to 250m keeping you 50m above ground elevation. If selected OFF the ground elevation is subtracted from the altitude. I managed to make this change myself and it seems to work fine. What do you think ?

PR #1307, opened 2021-06-23
Preview for efficiency in imperial units

avsaase:
This adds a preview for the efficiency elements when using imperial units. There is a small problem however:
image

It seems the left half of the mAh/mi symbol in the preview is a line break instead of the actual symbol, which causes the right half to appear on the next line. I don't see why this is. On the actual OSD the symbol is drawn correctly.


MrD-RC:
I can check this with the new fonts. I had a similar with LON because the symbol ID translated to new line. Worked fine in the OSD, but not configurator. I just swapped it with a character that wasn’t used in the preview.

MrD-RC:
Just checked, and all is fine with the new fonts. Sorry it took so long.

image
image

PR #1309, opened 2021-06-26
Change to facilitate the new Metric + MPH units set (#7195)

MrD-RC:
Ref https://github.com/iNavFlight/inav/pull/7195
Fixes https://github.com/iNavFlight/inav/issues/7191

The units used as "UK" were incorrect. We use imperial for distance and speed. To correct this, I have added a new units set called Metric + MPH. Check the firmware PR (above) for full details.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -2770,11 +2770,17 @@
     "osdUnitMetric": {
         "message": "Metric"
     },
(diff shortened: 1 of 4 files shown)


MrD-RC:
Merged with https://github.com/iNavFlight/inav-configurator/pull/1322

PR #1314, opened 2021-07-11
Search box for filtering OSD elements

nmaggioni:
Demo


DzikuVx:
awesome!

OptimusTi:
This is great!  There's a couple of small issues though:
1. Can only drag/change position of filtered items.
2. Artificial horizon and sidebars disappeared when moving filtering/moving GPS elements.

PR #1315, opened 2021-07-13
Added search box for targets

MrD-RC:
Added a search field for targets on the firmware flasher page. 

I was inspired by @nmaggioni's search box on the OSD page. Now we can search targets, which should make them easier to find.

https://youtu.be/F-aPR9DaTcw

image

### DIFF
--- a/src/css/tabs/firmware_flasher.css
+++ b/src/css/tabs/firmware_flasher.css
@@ -262,3 +262,16 @@
     text-shadow: none;
     opacity: 0.5;
 }
+
+.tab-firmware_flasher .target_search {
+    padding-left: 3px;
+    height: 20px;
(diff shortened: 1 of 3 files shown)


MrD-RC:
I will make the search box as wide as the drop down lists. It’s messing with my head having different widths 🤣

avsaase:
Is it possible to make the search box part of the drop-down list itself? Like this:
drop-down-list-with-search-jquery

DzikuVx:
@MrD-RC please mark PR when ready for merging

MrD-RC:
@avsaase I had a look at implementing this. But I don't think the datalist is supported. Everything else used external libraries

@DzikuVx ready now.

PR #1320, opened 2021-07-23
Cleanup ESC protocols

DzikuVx:


DzikuVx:
This side of https://github.com/iNavFlight/inav/pull/7295

Issue #1194, opened 2021-03-10
Modes tab improvements for beginners


avsaase:
Two questions I see frequently from fixed wing pilots new to INAV are: "Can I do ... with Airmode?" and "How do I enable ACRO mode? I don't see it on the modes tab". I think these misunderstandings could easily be solved by making two changes:
1. ~~Remove airmode on the modes tab when airmode is permanently enabled.~~ It already works like this. An alternative would be to remove it if the aircraft type is airplane.
2. Add a bar for ACRO mode that has no channel selection sliders, and that lights up blue when no other conflicting mode is selected. I think ACRO is compatible with ALT HOLD and maybe others, so the configurator would need to be aware of this.

MrD-RC:
I think these changes would be great. I'd also +1 on removing Airmode from airplane modes completely. Permanently enabled on or off should be fine for planes. The Acro bar would also be a handy visual aid.

I'd even add a 3. To implement the auto channel select feature from BetaFlight.

avsaase:
I'm not familiar with that feature (or betaflight really). What does it do?

MrD-RC:
The drop down select box where you choose the channel. In BetaFlight the top option is Auto. You would select Auto, then move the switch on the transmitter. It would then set the correct channel in the select box.

avsaase:
Okay yeah, that sounds pretty nice.

MrD-RC:
Also, just a note. Alt Hold is not compatible with Acro. Only Angle and Course Hold are.

MrD-RC:
I've started looking in to the ACRO box thing, but have come across a slight snag. The problem at the moment is that the checks are based on if the flight controller can enable a mode, not if the channels are in range. So I can get this working, but it is currently it would only work on active modes. So RTH could be in range, but it would still say ACRO is ok.

MrD-RC:
Acro indicator done. Plus a bonus of the in-range indicator.

PR #1342, opened 2021-09-10
Mode in range and Acro indicators

MrD-RC:
This adds functionality that shows when a mode is in range, not just currently active on the flight controller.

It also has an indicator to show when ACRO mode would be enabled.

image
In this image you can see the 4 indicators on the mode blocks:
- light grey: no ranges
- dark grey: mode range not active
- dark blue: mode range active
- light blue: mode active on flight controller (only changes on save, as with old configurators)

Also in the lower left corner is the ACRO indicator. When acro is enabled it is light blue. Else, it is dark grey. This works based on both active on flight controller and active ranges.

Modes that stop Acro from activating: Angle, Horizon, Manual, RTH, Position Hold, Cruise, Course Hold, Waypoint Mission, and GCS NAV.

Demo video https://youtu.be/tVq00H0g2Ps

Part fix for #1194


DzikuVx (changes requested):
Thais is super cool! But you know what @MrD-RC , please make the Acro indicator not looking like a button. People will mistake it for a button and will want to click

MrD-RC:
Updated the acro indicator to no longer look like a button. It now mimics the ranges.

image
image

I did have it so that just the text changed. But I think it looks better matching the ranges. If this is still too button like, I can change it so that just the text that changes colour.

DzikuVx (approved):

DzikuVx (approved):

PR #1347, opened 2021-09-17
Added value confirmation

MrD-RC:
This change adds a confirmation of entered values in the pilot's preferred units. The units are determined by the pilot's choice on the OSD page.

Example with Metric + MPH
image

Example with GA
image

## Adding more confirmation elements
To add this to other fields, just add the `<div class="displayPilotUnit _cm" id="loiterRadius_PU"></div>` div as the last html entry in the block for that field. The first class should always be _displayPilotUnit_. The second class changes depending on the units coming from iNav. The options so far are:
- __cm_ centimetres. Converts to metres for **metric** or feet for **imperial** and **GA**
- __cms_ centimetres/seconds. Converts to MPH for **imperial** and **metric + MPH**, Km/h for **metric**, and Knots for **GA**
-  __ms_ milliseconds. Converts to seconds
-  __us_ microseconds. Converts to percent, based on a 1000-2000 range
All have 2 decimal places, except for uS which has 1.

The ID of the div should match the ID of the input, followed by __PU_.

### Adding on more pages
Currently this is set up on the Advanced Tuning page and the Configuration page. So to add more to these pages, you just need to change the html.

To add this to other pages, you need to add the line `initialisePilotUnits();` to the JS file. It needs to be positioned after the inputs have been added.


avsaase:
This will save a lot of confusion but I'm not sure I like the visual appearance of this. Eventually we need to apply the transformation to the input box itself so the user just types the RTH altitude in meters and the CLI setting is set in cm in the background. 

As for the non-metric units, that should be a drop-down menu in the top bar of the configurator. This could then change the transformation applied to the input boxes. one problem with this is what you already see here, you don't end up with nicely rounded values. I'd say the unusual units things is more pressing than the non-metric units.

Don't get me wrong, I agree with the criticism that our cm values are confusing, and I appreciate you are trying to address that. I just think we need a more permanent solution.

MrD-RC:
The problem is I think that to have the input boxes change, would need a big re-write of the configurator. I honestly can’t see anyone doing that any time soon. Also, this would add inconsistency between the configurator UI and the CLI, which will always be in cm, cm/s, etc. I agree, a more permanent solution would be great. But I honestly can’t see it happening. Hence this stopgap, which can give pilots confirmation now. 

I agree that the confirmation values are not the prettiest of solutions. But it does add a function solution. Perhaps adjusting the colour of the text will make it less intrusive?

I can tidy up the rounding so that decimals are only shown if there are any. But we’re going to end up with decimals more often than not. After all, one of the reasons for using cm and cm/s was to avoid floating point. 

I’m also unsure of the unit selection needing its own box. Surely pilots are selecting the units they like to work with the most for the OSD. So why should configurator need to use anything different? Sure, it’s something that could be added. But is that system better, or does it just add another place where inconsistencies are added?

MrD-RC:
Made things a little neater
image

HerpDerpula:
(Forgive me if this in the wrong place) 
Continuing from my previous public posts in the FB group, I will pledge $500 to the developers who integrate human factors improvements like this. I'm one person, and the devs steer this program, but I am dead serious about supporting these improvements and want the community to back it. 

Thank you for taking the time to look into this. It's important that we retain an intuitive interface for iNav pilots.

elgansayer:
Pretty good start. 
I have already done some input work. It's really quite straight forward. I'll get a PR going soon.
I think a combination of what ive done and what you have done with the uit display would be a good move forward. I have a few more ideas.Give me a few days to get time here and there and i'll post a photo.

The conversion display. What is it like as tiny writing under the inputs like on my inav configurator screenshot on facebook?

avsaase:
There's a `data-setting-multiplier` setting for the number input fields. If set to 100 the cm values are shown in meters. I think that's all that's needed to make the units more user-friendly.

elgansayer:
Yep. I saw the multiplier. Inav basically has this built in. I did wonder if the multiplier is actually used anywhere. I can't find anywhere that isn't set as 1.

Still it's a bad way of handling it. Allowing all the data attributes to handle the multiplier value. It's a bit dangerous and repetitive.

I've basically copied it but made a single point of failure or as some people say. A single point of truth. This also allows handling the unit display for transition period. It also allows calculation of back conversation.

I do wonder if the conversions add too much info on the screen. 

When I'm home I can get some screenshots of my progress with inputs and add the conversations like this or but obviously the opposite values. The inav ones and the inputs show the desired 

Basically nearly there. I might end up upgrading some tooling too in another Pr. Also a few optimisation opportunities I've spotted. I think we could do so so much more to this app

MrD-RC:
> There's a `data-setting-multiplier` setting for the number input fields. If set to 100 the cm values are shown in meters. I think that's all that's needed to make the units more user-friendly.

Why should it be metres though? Surely if this is being done properly, it should accept units that the individual pilot are most happy using? What about the other fields like speeds, which are in cm/s?

elgansayer:
monkeys

Found a bit more time tonight after watching the matrix trilogy! anyway
Not quite there. The internal inputs were no big deal. The options were no big deal. Some better styling needed on options.

Only a few advanced options done. Super easy though. Just add `data-units="cm"` .

`<input type="number" id="launchMaxAltitude" data-unit="cm" data-setting="nav_fw_launch_max_altitude" data-setting-multiplier="1" step="1" min="0" max="60000" />`

As an example. The code handles the conversions and sets up the multipliers. The inav code already handles the rest.

Next? I will look at this PR code and add the right hand side inav unit displays. Kinda the opposite of this PR but in the same place and style? My CSS sucks but my JS is hot haha

Nandox7:
Just my two cents as I was also looking into this.

I agree with the option to allow the user to select whatever system they understand (Metric, Imperial,...)  while the FW stores all in the same unit, be it  anything.

The only issue (minor) I see with this approach is that changing between UIs that use different unit systems will show odd values BUT as long as the current unit system is clearly stated in the UI that is not a problem.
I only understand Metric if I see it stated as Imperial I know I need to go and change the UI to metric. As easy as that.

PS: Happy to help with making the change

elgansayer:
Yeah hence the option. Just need to make it refresh. An easy way is to
reconnect.

The code is all there working. Just needs nicer styling and maybe
conversion displayed  which I didn't get around to yet.

HerpDerpula:
I'd like to verify something here, is there any feasible way a pilot could enter a value in feet, for example, and the configurator fails to convert that to the correct number before it's applied? Do we have sanity verifications in place?

elgansayer:
There is no way to tell. The same way there is no way to tell if they enter
50cm instead of 5000cm

elgansayer:
Not possible

Nandox7:
> You're right, but I think I failed to frame my question properly. Is it possible that the configurator is set to feet, the user enters their value in feet, but the configurator fails to update the value from feet to its converted "final" value in cm, and it applies the original number instead of the corrected number. I.e. I enter 100 feet for the RTH altitude, with the option of "feet" selected, the configurator instead of converting the number, it enters 100 in as the final value, in cm.

Certain params can have min and/or max values.
Picking that example, why would someone  have RTH set as 100cm? Makes no sense, so min value check would be triggered and it would return in error or default to a predefined value warning the user.
Ok, but the user really wants 100RTH (it is a flying lanwmower), he goes to the CLI and set that up.

elgansayer:
That's a good idea. Easy to add too.

I will get a PR together soon. Just super busy atm. I think people will like the code. Just need to do the styling soon

elgansayer:
Still coming!! Sorry. been moving house. Not had a chance to fight CSS yet

DzikuVx:
@MrD-RC is this in a mergable state?

DzikuVx:
@MrD-RC please resolve conflicts

MrD-RC:
I've resolved the conflicts. But this change will be redundant if @elgansayer [PR](https://github.com/iNavFlight/inav-configurator/pull/1377) is merged. His would be better.

DzikuVx:
@MrD-RC the thing is that PR from @elgansayer has conflicts that have to be resolved before. And I want to cut 4.0 probably tomorrow

MrD-RC:
No problem mate. It's your call 😄

MrD-RC:
I've been chatting and working with @elgansayer and think it would be worth waiting for his PR to be ready. Mine is just a band aid for the issue, whereas his solves it. There's not much to do, but may have to wait for 4.1.

PR #1358, opened 2021-09-30
tiny typographical error messages.json

ryet9:
herrrp & derrrp, happy now mooiweertje? I know u don't like me making changes to my own local builds or as you put it "stealing code" so for this most valueable contribution your welcome!

save-->safe

To everyone else: sorry to waste ur time with a pull request for this minor non-functional change, but I was amused.

<         "message": "I2C speed should be kept at the highest level that allows for all connected devices to work. Default 400kHz is a save value and it is suggested to switch to 800kHz in case of Multirotors."
---
>         "message": "I2C speed should be kept at the highest level that allows for all connected devices to work. Default 400kHz is a safe value and it is suggested to switch to 800kHz in case of Multirotors."

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -3966,7 +3966,7 @@
         "message": "I2C Speed"
     },
     "configurationI2cSpeedHelp": {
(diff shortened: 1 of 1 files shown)


DzikuVx:
https://github.com/iNavFlight/inav-configurator/pull/1358

PR #1376, opened 2021-10-16
Mission control bug fix

breadoven:
Fixes an unseen error (only apparent if you look at the console).

`return data == []` appears to always returns false even when the array length is 0.

`repaintLine4Waypoints(mission);` shouldn't be called when there's < 2 WPs in the mission.


breadoven:
I'll merge this @DzikuVx if you don't object ?

PR #1381, opened 2021-10-19
Mission control Multi Missions

breadoven:
Provides Configurator Mission Control support for https://github.com/iNavFlight/inav/pull/6765.

Allows multiple missions to be edited and saved to the FC or to file.

Missions can be added to a multi mission either by plotting on the map or loading missions from file, with a max limit of 9 missions. Any number of single mission files can loaded. However, files containing multi missions will overwrite existing missions so need to be loaded first before editing or adding more missions. A multi mission can also be loaded from the FC which will also overwrite existing missions.

All missions can be displayed on the map together or missions selected individually for display. Missions can only be edited when displayed individually. When all missions are displayed editing is disabled. To edit a single mission it must first be selected either from a mission list or by selecting a WP in the mission and clicking a Mission Edit button on the WP edit box. After editing a mission it can be saved by updating the multi mission which will then display all missions on the map again.

Single missions can be saved/deleted when displayed individually on the map or the full multi mission saved/deleted when all missions are displayed on the map. Saving to the FC works the same way, can be individual missions or the full multi mission depending on what is displayed on the map when saved.

Seems to work OK with older INAV FC versions but since they're limited to a single mission version checking has been added to limit Multi mission use to version 4.0.0 anyway.


stronnag:
Kudos on solving the mission planner aspect.
btw the Mission File XSD is now part of [inav dev documentation set](https://github.com/iNavFlight/inav/tree/master/docs/development/wp_mission_schema). Note that its README mentions allowing the mission file to have either sequential or 'per segment' WP nos (and provides examples of both). You might consider supporting both in the configurator.

breadoven:
> Kudos on solving the mission planner aspect. btw the Mission File XSD is now part of [inav dev documentation set](https://github.com/iNavFlight/inav/tree/master/docs/development/wp_mission_schema). Note that its README mentions allowing the mission file to have either sequential or 'per segment' WP nos (and provides examples of both). You might consider supporting both in the configurator.<

Thanks for reviewing. Makes sense to add the per segment format to Configurator, simple to understand the Jump references. I did think it would be useful to add a mission name rather than relying on an index number, could be included in the header for each mission. The mission name would be displayed in the OSD when selecting missions. Would need to be saved to NVM though possibly using a special WP entry with the name encoded into the Lat Lon fields (8 characters). This would waste a WP but this isn't such an issue now with the 120 WP limit. Or maybe it's easier to just note down the mission indexes on a Post-It.

breadoven:
@DzikuVx @stronnag

Any issues merging this ?

Issue #1270, opened 2021-05-21
Configurator 3.0.0 - Mission Control - Load Eeprom mission button


ayoaina:
In Configurator 3.0.0 on Windows, when you click on the Load Eeprom mission button, nothing happens. You have to click on it a second time before it loads the mission.

ayoaina:
Update on this issue:
It appears that it works OK the first time you load eeprom mission after opening the Mission Control tab.
You get the issue if you delete the points and then try loading again.

Jetrell:
@ArnoTlse Is this edge case fixed in #1267 ?

ArnoTlse:
Hi,

no, not in this one, Eeprom loading was corrected in a former PR. 
I will try to reproduce this sequence with my current dev version.

Arno

ArnoTlse:
Hi @ayoaina , hi @Jetrell ,

On the last [version 3.0.0](https://github.com/iNavFlight/inav-configurator/releases/tag/3.0.0), It seems to work fine on my side.
The PR which corrects this seems to have been included.

Are you still facing issue with the last 3.0.0 version ?

If not, can we close this Issue ?

Regards

Arno

Jetrell:
@ArnoTlse With 3.0 Configurator and FW, using Windows, the issue still remains for myself.
It's not a deal breaker. It just means that I have to click on the `load eeprom mission` button twice, about 1 second apart.  Then it loads every time.

Jetrell:
@ArnoTlse With 3.0 Configurator and FW, using Windows, the issue still remains for myself.
It's not a deal breaker. It just means that I have to double click the `load eeprom mission` button.  

I just tested it on Window 7. 
If the `load eeprom mission` button is only pressed once.  About 5 seconds later the bing maps image disappears. And will not re-appear until the mission file is loaded again. Or if I press the `load mission from FC` button AFTER the FC has been power cycled, the mission will then load in MP bing maps. 
How does that work? I always assumed the the mission is stored in eeprom until its transferred to the volatile memory after boot-up. 

![Capture](https://user-images.githubusercontent.com/47995726/122657506-68a01f80-d1a7-11eb-9896-0bd94bcdfd47.JPG)

Jetrell:
@ArnoTlse I found the is issue. It occurs when `nav_wp_load_on_boot = ON` is used.

_Steps to reproduce. With a mission already saved to the eeprom and  `nav_wp_load_on_boot = ON` enabled._

[...]

PR #1388, opened 2021-10-31
Mission control WP send/receive fix

breadoven:
Allows single WP missions to be sent and received from Mission Control. This currently isn't possible without false RTH's appearing when loading and locked send buttons when sending.

Also adds send/receive times to log.

### DIFF
--- a/js/msp/MSPHelper.js
+++ b/js/msp/MSPHelper.js
@@ -519,7 +519,7 @@ var mspHelper = (function (gui) {
                         ));
                     }
                 }
-                
+
                 break;
 
             case MSPCodes.MSP2_INAV_LOGIC_CONDITIONS_STATUS:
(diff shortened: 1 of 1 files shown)


breadoven:
@DzikuVx @stronnag 
I'll merge this unless some reason is given as to why the current method only works with a minimum of 2 WPs ?

stronnag:
Sure. Don't often use configurator for missions; IMO it should allow the upload of a mission with a single RTH WP as a means of "clear the current in memory mission".

breadoven:
> Sure. Don't often use configurator for missions; IMO it should allow the upload of a mission with a single RTH WP as a means of "clear the current in memory mission".

I tried loading a single WP RTH mission. It doesn't clear missions entirely just shows the single WP RTH mission ... as you might expect. Not sure if this is what @ArnoTlse was after in https://github.com/iNavFlight/inav-configurator/issues/1270. I think he was hoping to clear it completely but the only way to do that is to zero the array so it's back to the post flashing state. I guess a single WP RTH mission just does the same as selecting WP mode with an invalid or empty/unloaded mission so equates to a cleared mission in practical terms.

stronnag:
If you use `mwp` or `impload` it works as expected (and as designed, some 6 years). e.g. this valid mission:
```
<?xml version="1.0" encoding="utf-8"?>
<mission>
  <!--mw planner 0.01-->
  <version value="2.3-pre8"></version>
  <mwp save-date="2021-11-01T15:53:15+0000" zoom="14" cx="-4.7021103" cy="54.1384679" generator="mwp (mwptools)"></mwp>
  <missionitem no="1" action="RTH" lat="0.0000000" lon="0.0000000" alt="0" parameter1="0" parameter2="0" parameter3="0"></missionitem>
</mission>

```

stronnag:
I can load the above file into the configurator, I can save it from the configurator to a file (with the same 1 RTH WP). I just can't upload it the FC via the configurator.

stronnag:
Indeed, this PR should be merged.

PR #1389, opened 2021-11-02
Changed GPS Speed to Ground Speed

MrD-RC:
This change is to make the term for velocity along the ground consistent with aeronautical terms. It also makes it consistent with the DJI speed selector drop down on the right side of the page.

I have also slightly modified the DJI labels. Using italics to highlight the sensor name in the DJI OSD setup.

image

image


MrD-RC:
I've noticed that in programming it's also ground speed. It seems like the OSD is the only place that isn't.

PR #1410, opened 2021-12-02
Block configurator until defaults have been applied

MrD-RC:
Currently, after clicking the chosen default on the defaults dialog, the box disappears and people can click away and change settings. This is obviously bad. This change fixes that. Configurator is locked until the reboot occurs.

[Demo of the change working](https://youtu.be/LyNAuGmqyxE)

I have set 5.0 as the milestone. But, it may be worth fast tracking this in to 4.0. What do you think @DzikuVx


JulioCesarMatias:
I'm glad someone decided to fix this, this is something that bothers me too

PR #1413, opened 2021-12-18
Fix defaults dialog bug when clicking Keep current settings

MrD-RC:
Fixes the bug where clicking "Keep current settings" will stop configurators working until reboot.


MrD-RC:
This would be a good fix to include if there is a 4.0.1 or 4.1 version of Configurator.

Issue #1319, opened 2021-07-21
3.0.x : Comport didn't open, depends on OS language version.


vcselvcsel:
Hello.

I've been flown with inav 1.9.x~2.6.1,

But 3.0.x configurator didn't reconize comport automatically and can't open comport.

I live in south korea, so I use korean language version windows10(x64).

One of my friend run 3.0.x configurator on english language version windows10(x64), it works.

Please fix the configurator.

Thanks.

* test condition : 
PC OS : Windows10 x64
target : matek F411WSE, matek F722miniSE.
configurator : win64 3.0.0, downloaded from github directly(not a chrome app).
firmware : 3.0.0 using configurator online download & downloaded from matek hompage(3.0.0)

jsungho:
I also use Korean Windows 64 version.
I have the same problem.

hogenejang:
same here. Can not open comport with this the lastest version. (win10 64bit (korean))
older versions are functional.

stronnag:
OS specific (i.e. Windows) issue, works fine in Linux with `LANG=ko_KR.UTF-8`.  Probably not fixable by the configurator maintainers. 

![Screenshot-20210804153746-1364x735](https://user-images.githubusercontent.com/158229/128201937-da1d280a-cb39-440a-8302-02443e69ce66.png)

SunjunKim:
I resolved the issue by downgrading nw.js version from the current `0.50.3` to the version `0.42.2` used in 2.x.

I tested various nw.js versions, as follow:

nw.js ver / Chrome vers / Working
0.49.2 / 86.0.4240.111 / FALSE
0.47.3 / 84.0.4147.125 / FALSE
0.46.4 / 83.0.4103.116 / FALSE
0.45.6 / 81.0.4044.138 / TRUE

The latest nw.js version worked is 0.45.6, on Windows 10 Pro (20H2), language set to Korean.
This trick worked for all INAV 3.x and 4.0 versions.

If INAV configurator doesn't have an explicit reason for bumping up the nw.js version, please let them stay under 0.45.6, which will prevent the further occasion of this issue.

bike797:
Betaflight configurator have this kind of issue too.
 nwbuilder version.
 - 44.6 serial port connection ok
 - 47.0 serial port connection fail
 - https://github.com/betaflight/betaflight-configurator/issues/2691

MorrisR2:
There were 65 changes between those NW versions 45.6 and 46.4.
[...]

PR #1415, opened 2021-12-23
Resolving issue #1319 - Serial port is not visible on some Windows versions

SunjunKim:
Issue #1319 reported serial port is not visible on Windows with some languages (notably Korean).

I found the issue is resolved by downgrading the nw.js version from the current 0.50.3 to the version 0.42.2 used in 2.x.

I further tested various nw.js versions, as follow:

nw.js ver / Chrome vers / Working
0.49.2 / 86.0.4240.111 / FALSE
0.47.3 / 84.0.4147.125 / FALSE
0.46.4 / 83.0.4103.116 / FALSE
0.45.6 / 81.0.4044.138 / TRUE

I suspect that Chrome messed up serial.getDevice() for some versions after 81.

The latest nw.js version worked is 0.45.6, on Windows 10 Pro (20H2), language set to Korean.
This trick worked for all INAV 3.x and 4.0 versions.

Unless INAV configurator has an explicit reason for bumping up the nw.js version, I'd like to recommend using v0.45.6, which will prevent the further occasion of this issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "jquery-ui-npm": "1.12.0",
     "marked": "^0.3.17",
     "minimist": "^1.2.0",
-    "nw": "^0.45.6-sdk",
+    "nw": "^0.61.0-sdk",
     "nw-dialog": "^1.0.7",
     "openlayers": "^4.6.5",
     "plotly": "^1.0.6",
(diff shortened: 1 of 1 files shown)


DzikuVx:
Please check if this problem occurs in nw `0.6.1`.

SunjunKim:
> Please check if this problem occurs in nw `0.6.1`.

I confirm that nw.js version 0.61.0 resolves the issue.

PR #1422, opened 2021-12-28
Fixed PID Tuning refresh. Extra table no longer is created

### LINKED ISSUE #1359: [BUG] In PID Tuning Tab :  "Refresh" button causes second form-table
**Issue:**
When you click refresh under the "PID tuning tab" there is another form table generated below the original table.
The First table is just like the previous table and below that secondary table is generated.

**How to recreate this bug?**
- Connect to flight controller [In my case I used Matek F405 Wing]
- Go to "PID tuning" Tab.
- Click "refresh" button directly.
- you will see another table just below the main table.

image

MrD-RC:
Fixes #1359

### DIFF
--- a/tabs/pid_tuning.js
+++ b/tabs/pid_tuning.js
@@ -190,6 +190,9 @@ TABS.pid_tuning.initialize = function (callback) {
         // UI Hooks
 
         $('a.refresh').click(function () {
+            $("#content-watermark").remove();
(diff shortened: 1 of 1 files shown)


MrD-RC:
Another bug fix that would be good if a 4.0.1 or 4.1.0 version of Configurator is released.

Issue #1001, opened 2020-05-22
Missing feature: BLE support


marianomd:
BLE doesn't have RFCOMM support, so it would be very useful to configure iNav through HC-08 or HM-10 modules or directly on FCs like SpeedyBee F4/F7, Racerstar MELO F4, Diatone Mamba F722S, and other FCs that have CC2541 chip.

Chrome has support for BLE: https://googlechrome.github.io/samples/web-bluetooth/
I managed to discover the device with one example from that page.

I guess we could take the implementation details from a project like this: https://github.com/monteslu/ble-serial

It seems it should be straightforward. I will investigate further.

tonyyng:
I would find this very useful.  I found it mentioned in this [issue](https://github.com/iNavFlight/inav-configurator/issues/2), but it doesn't appear it went anywhere at that point.

elepete:
Another thought on BLE (HM-10 / HM-11 like modules). As rfcomm is not supported there is a way to get a "pts" device linked to serial communcation of the BLE module. Please see this link:
https://github.com/Jakeler/ble-serial
This is an easy way to get a "serial" for BLE in Linux (e.g. as /dev/pts/2). I have tested that under Debian Buster, works fine and stable. Only issue is that you cannot open any device that is not dialout or a full tty. Terminal programs perform just fine on a 'pts' and communicate well via BLE to the MSP serial port on the FC.
Issue is that INAV_Configurator is not accepting or really creating an error opening such 'pts' device instead of a 'tty'. If the configurator could be changed that would be another way of having BLE serial support.

If all that works I would add a BLE to my copter, and have a logic switch to enable it in disarmed state only. In that way the USB cable is only needed in super rare cases + phone is always a good option to do changes if needed.

BLE support in Configurator would indeed be a real plus.

elepete:
Under Linux it works with the ble-serial.
A thanks to Jakeler who created the ble-serial and looked behind things:
https://github.com/Jakeler/ble-serial/issues/6#issuecomment-653365093

michaelvanderhoek:
ANy updates on this topic?

MrD-RC:
Fixed with https://github.com/iNavFlight/inav-configurator/issues/1436

PR #1441, opened 2022-01-29
Adding BLE Support (SpeedyBee, HM-1X)

Scavanger:
Support for Bluetooth low Energy devices:

Supported are:
BLE to UART adapters: 
- HM-1X (HM-10, HM-11, -..)
- HC-08
- Nordic Semiconductor BLE Shield

SpeedyBee adapters and FCs with integrated Bluetooth.

Unfortunately, the Chrome BLE API is somewhat limited compared to Android. This results in the effect with the UART adapters, due to the buffers, that either the order of the bytes is messed up or even bytes are missing. Therefore, the FC must split large frames and send them one after the other with a certain delay.
There is a new MSP command for this: MSP2_INAV_SET_MSP_OPTIONS, with which the configurator can request transmission in chunks and a delay between the cunks, see here: 

The Speedybee adapters are not affected by this, they seem to have a large enough buffer.

#1001

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -87,6 +87,8 @@ sources.js = [
     './js/msp/MSPchainer.js',
     './js/port_handler.js',
     './js/serial.js',
+    './js/serial_com.js',
+    './js/serial_ble.js',
     './js/servoMixRule.js',
     './js/motorMixRule.js',
     './js/logicCondition.js',
--- a/js/msp/MSPCodes.js
+++ b/js/msp/MSPCodes.js
@@ -227,5 +227,6 @@ var MSPCodes = {
     MSP2_INAV_FWUPDT_ROLLBACK_EXEC:     0x2037,
     
     MSP2_INAV_SAFEHOME:                 0x2038,
-    MSP2_INAV_SET_SAFEHOME:             0x2039
+    MSP2_INAV_SET_SAFEHOME:             0x2039,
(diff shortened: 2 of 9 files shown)


Scavanger:
Will be redone with UDP and TCP support.

jbienz:
@Scavanger, are you still planning BLE support, or have you changed your mind and only planning WiFI / UDP / TCP support? I don't mind the WiFi approach, but it does either mean that you're having your PC join an AP created by the device or the device has to be on your network. It seems that neither of those are ideal when out at the field. One of the nice things about BLE is it "just works" and doesn't need any network configuration.

Scavanger:
@jbienz 
Yes, BLE support will come.
I have merged this with TCP and UDP so that they share a common code base.I think I can finish the tests soon and make a new PR.

jbienz:
Amazing @Scavanger, this would be so helpful. I started looking into this library as a workaround:

https://github.com/Jakeler/ble-serial

It's supposed to allow you to map BLE devices to COM ports with some hackery of finding the right BLE characteristic to talk to. But unfortunately, if you read all the way down to the end of the page he mentions that Chromium / Electron-based apps won't be able to open the port. So your BLE support directly in iNav is the next best hope!

Please let me know if you'd like any additional help testing. I have two different HC-08 modules on the way (one with pincode and one without) as well as a Flywoo BT-Nano.

PR #1447, opened 2022-02-08
Add groups to programming for ease of reading

MrD-RC:
This PR allows people to group sections of the Programming. This is to help make the code easier to read, without having to have empty lines of code as separators.

It requires firmware update https://github.com/iNavFlight/inav/pull/7812

[A video of this in action.](https://youtu.be/2mgB9D4kaTc) Since recording the video, I have made some subtle cosmetic tweaks. Mostly how the divider lines appear.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -3790,6 +3790,9 @@
     "logicEnabled": {
         "message": "Enabled"
     },
+    "logicCodeGroup": {
+        "message": "Code Group"
+    },
     "logicOperation": {
         "message": "Operation"
     },
--- a/js/logicCondition.js
+++ b/js/logicCondition.js
@@ -1,7 +1,7 @@
 /*global $,FC*/
 'use strict';
 
-let LogicCondition = function (enabled, activatorId, operation, operandAType, operandAValue, operandBType, operandBValue, flags) {
+let LogicCondition = function (enabled, activatorId, operation, operandAType, operandAValue, operandBType, operandBValue, flags, codeGroup) {
     let self = {};
     let $row;
 
@@ -69,12 +69,26 @@ let LogicCondition = function (enabled, activatorId, operation, operandAType, op
         flags = data;
     };
 
+    self.getCodeGroup = function () {
+        return codeGroup;
+    };
+
+    self.setCodeGroup = function (data) {
+        codeGroup = data;
+    };
+
     self.onEnabledChange = function (event) {
         let $cT = $(event.currentTarget);
         self.setEnabled(!!$cT.prop('checked'));
         self.renderStatus();
         self.renderActivator();
     };
+
+    self.onCodeGroupChange = function (event) {
+        let $cT =$(event.currentTarget);
+        self.setCodeGroup($cT.val());
+        renderCodeGroups();
+    }
 
     self.getOperatorMetadata = function () {
         return FC.getLogicOperators()[self.getOperation()];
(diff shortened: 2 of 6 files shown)


error414:
https://github.com/error414/inav/tree/MrD-Add-groups-to-programming-for-ease-of-reading

I tested it and it works nicely.

MrD-RC:
We decided not to use the PR as the GUI display options shouldn't be stored on the FC. It should be stored on the PC/Configurator. Ideally in the cloud somewhere, so that the groups will work on and computer for that flight controller.

error414:
ahh, it's a pity :(

sensei-hacker:
> We decided not to use the PR as the GUI display options shouldn't be stored on the FC. It should be stored on the PC/Configurator.

As I recall, the concern was that each line gets a new byte, the section number, correct?

It occurs to me a slight change to the implementation would make it there would be NO storage cost at all, when the feature isn't actively being used. That could be done in the same spirit as the html BR or HR tags or the rtf "page break" and "paragraph break" tags. We could add a PLC command for "section break".


In that way, any existing PLC would be stored exactly the same way it has always been stored. Only if the user choose to mark a section break 

IMHO if you have a chunk of code for setting the VTX power and a separate chunk of code for adjusting the throttle, those are separate LOGICALLY. Demarcating separate units is a semantic thing, IMHO (and in the view of those who created the HTML standard). The html standard considers DIV or HEADER to be a *semantic* fact because the different sections are logically separate, while "blue" is a display issue controlled by CSS.

It seems not a decision on how to *display* different sections. The display decision is whether to display a horizontal bar between units, highlight them in different colors, draw a box around each unit  ...

MrD-RC:
No, it's a decision based on not having code or parameters in the firmware which solely deal with UI. That would be the only reason they exist. So the firmware is not the right place to keep them.

error414:
As I see it, saving groups to PC or to cloud has as many disadvantages that may be would better not to do implement groups. Local storage is good as cache,  not for data which are important for users. 

I would bet that all users who have more than one script in programming framework use logic condition line as separator beetween scrips. If you use coloured groups you don't have to use logic condition line as separator, but if you lost somehow saved groups info, then it will be mess in programming framework. 

I understand why you don't want to have GUI params saved in FC, but it's only single solution how to do it user friendly.

MrD-RC:
I can see your point. Maybe an implementation like @sensei-hacker suggested with a separate logic_display parameter would be better. Then it's only used if logic is used. Though after 32 LCs being used, it would take up more memory.

But yes, I use an LC as a space between sections of logic.

sensei-hacker:
>  But yes, I use an LC as a space between sections of logic.

So do I. You and I write the examples, and post things to guide others on YouTube, Facebook, Discord, etc. So it's likely others picked that up as well.

Therefore, without changing anything at all in firmware, we could choose in Configurator to render an empty LC in-between non-empty ones similar to how most browers render a break element (HR), or render it as a color shift, or maybe just use a slightly different background color for the empty ones.

Aesthetics isn't my thing, but that last option is sounding the best to me - render empty conditions with a slightly darker background. That would make them visually delimit the sections nicely, perhaps.

error414:
I like it, it's a simple solution, 

_"render empty conditions with a slightly darker background. That would make them visually delimit the sections nicely, perhaps."_

MrD-RC:
But that would still need storing somewhere. Or do you just mean the empty LC is displayed differently? In which case, we still lose an LC wherever we want a space.

sensei-hacker:
> But yes, I use an LC as a space between sections of logic.

I do too. Since you and I create a lot of the examples others learn from, it's a good bet that some others do too.
This image has no changes at all to the firmware, just CSS:

image


> Or do you just mean the empty LC is displayed differently? In which case, we still lose an LC wherever we want a space.

Yes, there's a cost - always. But we only pay that cost when we use a break, so people there is no cost to anyone who doesn't use the feature.

> Though after 32 LCs being used, it would take up more memory.

We have 64 now. :)

PR #1449, opened 2022-02-09
Magnetometer Tab without iFrame [do not merge yet]

ThisIsJustARandomGuy:
This PR expands on the amazing work by @kernel-machine in https://github.com/iNavFlight/inav-configurator/pull/1439 and adds a magnetometer alignment tab to the configurator. It is just a WIP for now, for the following reasons. The code still needs some cleaning up, but most importantly: I do not use INAV or GPS modules so I have some trouble understanding how the alignment angles are actually supposed to work. Before this gets merged, someone probably needs to take a look at

1) The angle calculations (I'm pretty sure they're wrong ATM). I do not understand what is supposed to happen at all - i.e. what is the orientation of the module when CW 180° is set and so on. If I understand the comments in @kernel-machine's pull request PR#1439 correctly, it appears to depend on FC orientation as well? Anyways, these should be pretty easy to adapt depending on what magic is actually going on as soon as I get it (or someone provides me with the correct mapping function(s)).

2) The "model" I've used for the GPS module is a red box. Though, I imagine it might be quite useful for the user to be able to see the actual orientation of the module. I didn't want to learn blender for this PR (I tried!) so it is what it is, for now. Maybe someone can chime in with a better gps.json model?

3) Currently, the positioning of the gps model is static. Since I have chosen to use the correct model for the vehicle, depending on the selected mixer, the module is currently rendered inside of the model for some configurations. For example, since the boat model is just a large box, the GPS is hidden inside. Easy fix.

4) The CSS is basically a messed up copy of the GPS tab. I'm really sorry but I have a natural allergy to anything CSS related. I'd be so happy if someone could clean it up somewhat. It works on any screen size but it's definitely not a beauty.

5) Maybe add some sliders in addition to the input fields so that large changes are easier to make?

image


rts18:
> 
>it appears to depend on FC orientation as well? 

@ThisIsJustARandomGuy  Yes, The FC orientation is just as important as the magnetometers to get the correct setup.
Its good that you have asked this question. Because it proves this is little known. And by having the Magnetometer alignment tool include a FC board alignment tool as well, would make it complete. @kernel-machine Any plans?
Take note of the arrow on both the FC and the GPS/magnetometer.  When both arrows are facing UP and towards the FRONT of the multicopter. That is iNav's default.  

compass_mag_orient
Note the magnetometer chip circled on the bottom of the GPS unit. It is upside down and rotated 270° clockwise.  This is standard on most combine GPS/magnetometers. That is why CW 270° FLIP is the iNav default. In other words. Just take note of the arrow on the board and not the manual if that's what they recommend.

Like this for example, from the Matek M8Q+5883 manual : 
* _Compass Alignment(Arrow forward and flat mounting): Tilting the magnetometer is strongly discouraged !_
* _INav/BetaFLight: CW 270° Flip when flight controller arrow is facing forward also._



FC board direction arrow

But if any other orientation of GPS/Mag and FC is used, its can quickly get very confusing.

ThisIsJustARandomGuy:
Sorry fo the late reply! Ill merge as soon as possible

DzikuVx:
@kernel-machine @ThisIsJustARandomGuy should we merge then?

DzikuVx:
Replaced by #1499 thank you all for your work! Highly appreciated

PR #1451, opened 2022-02-12
Enhance and extend units conversion

### LINKED ISSUE #1428: Change displayed units in metric rendering option
## Current Behavior
I truly welcome the new rendering option in the configurator. I came across a few "looked over" that i suggest will be adjusted for the next release:

## Steps to Reproduce
1. set "metric" in the rendering option
2. select "Advanced tuning"
3. check "General Navigation Settings"
4. select "Mission Control"
5. create Waypoint
6. activate Waypoint and check "speed"

## Expected behavior
Consistency and realism in the unit conversion

## Suggested solution(s)
Change the following units:
- Advanced Tuning -> General Navigation Settings -> Max. Alt-hold climb rate from **km/h to m/min**
- Advanced Tuning -> General Navigation Settings -> Max. navigation climb rate from **km/h to m/min** 
- Mission Control -> Waypoint configuration -> Speed from **cm/s to km/h**

## Additional context
Climb rates are usually defined in ft/min so i suggest the metric equivalent would be m/min. This is more consistent to the rest and realistic when planning than km/h.
I have not tested the other rendering options. Could be worth a look to see if the same values were overlooked as well.
Overall a great new feature!

---

# Configurator 4.0.0
# INAV/MATEKF765 4.0.0 Dec 15 2021 / 15:15:30 (c20a9852)
# GCC-10.2.1 20201103 (release)

MrD-RC:
1. Fixed rounding error
2. Fixed error that displayed centi° for nav_fw_pitch2thr_threshold. It should have been deci°. The calculation was fine (10 x multiplier). Only the label was wrong.
3. Added conversions for:
a) vertical speed
b) metres
c) temperature
4. Added support on Failsafe and OSD pages
5. Sorted out alignment issues, due to different implementations on different pages
6. Hovering over a field shows the CLI command

This should now take care of all places that need unit conversion, other than **Mission Control**.

### Advanced Tuning
image
Alignment all looks fine. Note the speeds in the **General Navigation Settings** and **Automatic landing settings** are now in metres per second for vertical speeds. This matches the units used in the OSD for the digital vario. Imperial is feet per second and GA is 100 feet per second.

### Failsafe
image
Looks the same, but quite a lot has changed. All data except for the procedure checkboxes no uses the _data-setting_ method. This will be able to be fully converted one PR [#7734 ](https://github.com/iNavFlight/inav/pull/7734) is merged. As that has the changes required to handle string over the _data-setting_ method.

### OSD
image
Just the alarms here. I chose to keep **Fly Time** as minutes. But I could change it to be hours with non-default units if people wish. You may also notice that the imperial temperatures are in Fahrenheit.

As always, open to comments or suggestions.

Half-Fixes #1428

### DIFF
--- a/js/settings.js
+++ b/js/settings.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const GulpClient = require("gulp");
+
 var Settings = (function () {
     let self = {};
 
@@ -27,6 +29,8 @@ var Settings = (function () {
                     return;
                 }
                 parent.show();
+
+                input.prop('title', 'CLI: ' + input.data('setting'));
 
                 if (input.prop('tagName') == 'SELECT' || s.setting.table) {
                     if (input.attr('type') == 'checkbox') {
@@ -69,9 +73,9 @@ var Settings = (function () {
                     input.val(s.value.toFixed(2));
                 } else {
                     var multiplier = parseFloat(input.data('setting-multiplier') || 1);
-                    input.attr('type', 'number');
-                    input.val((s.value / multiplier).toFixed(Math.log10(multiplier)));
 
+                    input.val((s.value / multiplier).toFixed(Math.log10(multiplier)));
+                    input.attr('type', 'number');
                     if (typeof s.setting.min !== 'undefined' && s.setting.min !== null) {
                         input.attr('min', (s.setting.min / multiplier).toFixed(Math.log10(multiplier)));
                     }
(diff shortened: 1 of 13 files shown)


MrD-RC:
@DzikuVx review requested as it would be handy to get this merged before I continue on another PR.

DzikuVx:
@MrD-RC go with the merge

Jetrell:
@MrD-RC I noticed in the 5.0 build, the `Automatic Landing Settings `have turned out a bit jumbled.

landing 5 0

Due to this requiring a fix... I thought I would suggest an idea. 
Recently on Discord, I explained how these settings work. Because the user gets confused by the label's, rather than what the setting actually does.

And due to the current label's in the Configurator having nothing to do with the CLI commands for these set points,
`nav_land_maxalt_vspd`
 `nav_land_minalt_vspd`
 `nav_land _slowdown_maxalt`
 `nav_land_slowdown_minalt`

I thought it could be more directly worded to something like this. Which explains exactly what each of these settings do.
unknown

**It does require a change in the code, to swap setting 3 and 4. Which makes it more systematic, according to the landing sequence than at present.**

What do you think?

MrD-RC:
@Jetrell I agree. Those labels have never made any sense to me. They’ve been like that since the landing stuff was changed; think that was 3.0. Can’t remember when I saw that PR? What you’ve suggested would make a lot more sense. Though there’s no need for the max values, these are set automatically in the fields and change with the unit types now. 

BTW, to find out the CLI commands, just hover over the input box.

MrD-RC:
@Jetrell https://github.com/iNavFlight/inav-configurator/pull/1477

Jetrell:
Nice work! Thanks Darren :+1:

js/settings.js
```diff
@@ -239,18 +306,31 @@ var Settings = (function () {
...
+                element.attr('min', (element.attr('min') / multiplier).toFixed(2));
+                element.attr('max', (element.attr('max') / multiplier).toFixed(2));
+            }
         }
 
         // Update the input with a new formatted unit
-        const convertedValue = Number((oldValue / multiplier).toFixed(2));
-        const newValue = Number.isInteger(convertedValue) ? Math.round(convertedValue) : convertedValue;
+        let newValue = "";
+        if (multiplier == 'FAHREN') {
```
elgansayer, line 317: ===

elgansayer:
Nice to see. Great work!
There a few minor JS coding improvements i could suggest. Like literally nothing impacting but i'd like to see the JS codebase slowly modernise and use more robust JS.

I wish i had time finishing my own Inav configurator i started in dart. The best i could do is try and keep an eye on the PR's and suggest some edits if you wish?

Infact in the future we could try and replace some of the ever so outdated page systems with something more modern, robust and slowly strangle it out. oo That would be fun. At least bump some deps and add in some unit testing on UI. Oh yes yes now im drooling.

PR #1452, opened 2022-02-14
Updated mixer

MrD-RC:
- Tidied up the code a little and grouped mixer types in model.js.
- Removed second motor from all fixed wing mixers that aren't differential thrust. A bug was discovered that if an FC only has a single motor output, and 2 motors are set in the mixer, the motors won't work.
- Added mixer for V-Tail with differential thrust
- Enhanced the new servo button. Clicking it now selects the next available servo number, rather than just 0.

### DIFF
--- a/js/model.js
+++ b/js/model.js
@@ -44,6 +44,7 @@ const
 
 // generate mixer
 const mixerList = [
+    // ** Multirotor
     {
         id: 1,
         name: 'Tricopter',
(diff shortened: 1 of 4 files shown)


wx4cb:
is there a way that there can be a "renumber servo" button? i've found that at the start the numbers are completely messed up with you load a mixer and the first thing i do is renumber the servos to be consistant

MrD-RC:
Huh? With 4.1 the servos all start at 1.

wx4cb:
well... as you just said, that was fixed in 4.1 which i havent checked out yet so just ignore my drunk ass :D

MrD-RC:
image
Mixer no long shows more than 3 decimal places. Correct values only are shown.

PR #1478, opened 2022-03-09
CLI reboot on disconnect fix

breadoven:
Fixes issue where the FC fails to reliably reboot when Configurator is disconnected with the CLI active. 

Related to https://github.com/iNavFlight/inav/issues/7281.


breadoven:
Well this seems to work OK. Not had a failed CLI reboot and the other tabs work as before so should be good to merge.

PR #1486, opened 2022-03-13
Dynamic airplane mixer images

MrD-RC:
A video says a million words, so [click this](https://youtu.be/Bzgj4aXjiII). Note the coloured bars on the servo rows and outputs table are just video edits. The mixer colours have not changed. On the last change in the video I forgot to swap the blue and green colours in the servo mixer table, when the stabilised yaw was set to servo 12. But, you can see the numbers all match up.

The video has everything, except for the notice that the mixer is not loaded. So here's a screen shot of that.
image


wx4cb:
you could do the whole row like in the video, but i think that just the "servo number" on the bottom part of the mixer would be enough

MrD-RC:
image

DzikuVx:
Seems like most problem are caused when on initial state there is no mixer defined at all.

For example

image


This happened when `Refresh mixer` was clicked without any mixer rules

MrD-RC:
Cool, I'll get on that. Seems like the check for the Refresh mixer needs to be more stringent.

MrD-RC:
[Bug fix resolution](https://youtu.be/9uqnM833ybk)

PR #1489, opened 2022-03-14
Add props in configuration to mixer page

### LINKED ISSUE #793: motor direction switch in GUI 2.1.4 missing !!!
as title says motor direction switch in GUI is missing 
the interface is supposed to make life easyer right. 

same thing i have seen missing in latest BF GUI interface app aswell. 
 its annoying because when i fly low all the grass get's thrown as a salad in the camera lens so i prefer reversing motor rotation to avoid it.
BFmotor_direction

meaning i prefer having a graphical representation of how motor direction is set in the FC. 
if this has been submitted just close the issue.
thanks.

DzikuVx:
closes #793


MrD-RC:
https://github.com/iNavFlight/inav-configurator/issues/1495

PR #1493, opened 2022-03-21
This is a Pull request that contains only the INAV configurator Chine…

YI-BOYANG:
…se language table, which contains the Chinese translation for most of the content. However, the INAV configurator seems to run with the last-ordered language list as the display language, regardless of the default language for Chrom and OS. So it may be up to developers to solve the default language problem.


YI-BOYANG:
微信截图_20220321192625

Issue #1436, opened 2022-01-17
Support MSP over UDP


digitalentity:
There are multiple solutions to provide connectivity to the FC via WiFi, i.e. https://ardupilot.org/copter/docs/common-esp8266-telemetry.html

Typically they encapsulate telemetry packets (MAVLink) into UDP or TCP and FC doesn't even know that it's talking over the wireless connection.

However, on the Configurator side support for MSP-over-UDP is necessary (as the bare minimum). Corresponding ESP8266 or ESP32 firmware can provide the access-point functionality and UDP endpoint to wrap/unwrap the MSP packets and talk to the FC over the conventional UART.

digitalentity:
@MATEKSYS

wx4cb:
to be honest, TCP would be the better way of doing this... UDP as a fire n forget protocol could cause more hassles than it's worth.

the data rate isnt that high to be honest and easily handled by a TCP connection

digitalentity:
TCP or UDP doesn't matter much IMO. WiFi is acknowledging medium by itself (it's ACKing every frame, except broadcast, on the medium control level), so even in case of UDP that would be pretty reliable. Also, the existing MSP stack in the configurator is retriable and stateless, so this shouldn't be a problem even if some of the packets are lost. But supporting both TCP and UDP would indeed be beneficial and IMO not so different from the implementation perspective.

MATEKSYS:
via UDP with ESP8266-wifi-temeletry, very smooth.
https://user-images.githubusercontent.com/23047106/149765209-e2528e17-81ef-4e1d-ba22-db00934ced7e.mp4

wx4cb:
@MATEKSYS nice. 

@digitalentity i've used an ESP in there hooked up to a uart in wifi to serial mode for the (similar as to how we hook one up to crossfire wifi for the mini crossbow AAT), and i've found TCP mode to be more reliable even at 5 feet away from the airframe, but obviously supporting both is trivial.

digitalentity:
Either way - this is about supporting MSP-over-network in Configurator, which I believe will mostly change the UI and the serial connection layer. We might need custom MSP-aware firmware for the ESP8266 though.

FPVZaphod:
If UDP or TCP becomes a bit more common for us iNav folks, an enlightened person might invest time to create an iOS version? Then we don't depend on not existing serial support of iOS anymore. Native configurator on iPad  or iPhone would be amazing!

[...]

PR #1498, opened 2022-03-31
BLE, TCP and UDP Support

Scavanger:
Adds support for wireless connections.

Bluetooth Low Energy:
Hardware:
- CC2541 based modules like HM1X/HC08/09
- Nordic Semiconductor NRF (Adafruit BLE Shield)
- SpeedyBee adapters

For CC2541 based modules a small hack in iNav is necessary: https://github.com/iNavFlight/inav/pull/7931
See also: https://github.com/iNavFlight/inav/pull/7783

TCP and UDP
See discussion here: #1436

I have written a firmware for ESP8266 modules that allows easy configuration via a web interface.
It checks MSP frames for errors and wraps an MSP frame in one UDP/TCP packet. 
UDP runs very smoothly, TCP is a bit laggy.
https://github.com/Scavanger/MSPWifiBridge
But any device that forwards the data of a UART one-to-one to TCP/UDP should work the same way. 

if #1447/https://github.com/iNavFlight/inav/pull/7812 is merged, this PR/https://github.com/iNavFlight/inav/pull/7931 may needs a rebase.


js/msp/MSPHelper.js
```diff
@@ -463,22 +463,33 @@ var mspHelper = (function (gui) {
                 console.log("Servo mix saved");
                 break;
             case MSPCodes.MSP2_INAV_LOGIC_CONDITIONS:
-                LOGIC_CONDITIONS.flush();
```
DzikuVx, line 466: @Scavanger this is a great PR. But we have to change how it's done. I've left a review in the INAV repo but it ultimately goes down to:

1. Leave `MSP2_INAV_LOGIC_CONDITIONS` unchanged
1. Add `MSP2_INAV_LOGIC_CONDITIONS_SINGLE` frame
1. Switch configurator to use `MSP2_INAV_LOGIC_CONDITIONS_SINGLE`

And then, in one of future releases we can remove support for `MSP2_INAV_LOGIC_CONDITIONS` completely.

What do you think?

Scavanger:
@DzikuVx 

Yeah, this is the cleaner way.
Should the version check for firmwares < 5.0 remain in and then be switched back to MSP2_INAV_LOGIC_CONDITIONS if necessary?

DzikuVx:
@Scavanger yes, we'd like for Configurator 5 to support INAV 4. So it should indeed check for supported message.
If this will be to time consuming or just too complex we can always say: sorry, we do not support older version. Your call

DzikuVx (approved):

DzikuVx:
@Scavanger could you please add a small doc here or in the firmware repo on how to use it? It would be awesome to have at least a short explanation of what's going on

Scavanger:
@DzikuVx 
https://github.com/iNavFlight/inav/pull/7931

EduardTheThird:
Tried this feature using this board, https://www.adafruit.com/product/2821,
running this code:
https://github.com/EduardTheThird/inav-configurator-ESP8266-WiFi-UART-transparent-bridge/blob/master/WiFiUartTransparentBridge/WiFiUartTransparentBridge.ino

Worked, thanks! I only had to change my baud rate to 57600.

PR #1507, opened 2022-04-21
Moved font select/upload box

MrD-RC:
Moved the font select/upload box down. This allows the OSD preview window to still be visible in most cases.

image


Jetrell:
Nice.. That's always been in the road when previewing a font before uploading.

PR #1516, opened 2022-05-10
Universal Sliders for Settings framework

DzikuVx:
image


OptimusTi:
Niiice. Something to show defaults or a reset to defaults would be cool.

MrD-RC:
With reset to defaults, it should reset to the defaults selected in the defaults dialog box. Not the FC defaults. That was the problem with the old one.

PR #1519, opened 2022-05-13
Move the 3D config to Outputs tab

DzikuVx:


MrD-RC:
Cool. I think that when functionality is moved between tabs. It should be mentioned in the release notes. I’ve certainly had “where the f**k has that gone?” moments in the past. it would be nice to avoid those for folks who take the time to read the notes.

DzikuVx:
@MrD-RC nobody reads release notes :)

MrD-RC:
Then they need a slap 🤣

PR #1529, opened 2022-05-29
add new requirement to make chrome_crashpad_handler executable

stronnag:


stronnag:
5.0.0+ needs this when generated on Windows, new framework requirement.

stronnag:
this should be added to the RN as well.

PR #1539, opened 2022-06-03
Add glide time, glide distance, and climb efficiency OSD elements

MrD-RC:
Firmware part https://github.com/iNavFlight/inav/pull/8081

Details and demo on the firmware PR.

Don't merge, I have a question for the group.

Currently the icons for climb efficiency show  and  for Ah per vertical metre or vertical foot respectively. I'm questioning if the v/m and v/ft parts are correct. Or if they could be something better and easier to understand. Some thoughts are

- Dash  
- No space   This option would use 2 less characters, but I'm not sure vm and vft would work
- Arrow  

I think I'm leaning towards the arrow option. I think it looks clearer as you have a direction and the units in a format everyone is used to. But I thought I'd ask everyone and get a consensus.


molokaiboy:
Arrow

MrD-RC:
From comments here and on Discord, arrows look like the preferred choice. It was unanimous, so I'll get those font changes made and this will be ready to merge.

molokaiboy:
Another awesome feature!!

Mahalo,
Les

PR #1547, opened 2022-06-07
Drop the possibility to adjust ESC rate from GUI

### LINKED ISSUE #1543: Changing Idle throttle resets ESC Refresh rate to 50Hz
Changing the idle throttle value, so the green info box disappears, changes the ESC refresh rate value to 50Hz. If user saves, the 50Hz refresh value is also saved. See screen gab.

This is Configurator 5.0 RC2 release.

https://user-images.githubusercontent.com/33039058/172062765-db72a00e-dc2f-4110-b1b6-80a20e401809.mp4

DzikuVx:
fixes #1543

### DIFF
--- a/tabs/outputs.html
+++ b/tabs/outputs.html
@@ -16,20 +16,12 @@
                 </div>
                 <!--list of generated features goes here-->
                 <div id="esc-protocols">
-                    <div id="esc-protocol-warning" class="warning-box"></div>
                     <div class="select">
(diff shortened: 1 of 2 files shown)


seagateuk:
Why did you do this?  There are many ESCs that won't arm with 40OHz and need to start with 50Hz signal, particularly EFlight ESCs.  Can the rate still be set in the CLI?

PR #1551, opened 2022-06-10
Add expanded labels for units

MrD-RC:
This PR adds expanded hover labels for units.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,9 @@ Depending on target operating system, _INAV Configurator_ is distributed as _sta
 1. Visit [release page](https://github.com/iNavFlight/inav-configurator/releases)
 1. Download Configurator for Linux platform (linux32 and linux64 are present)
 1. Extract tar.gz archive
-1. Make the inav-configurator file executable (chmod +x inav-configurator)
+1. Make the following files executable:
+   * inav-configurator `chmod +x inav-configurator`
+   * (5.0.0+) chrome_crashpad_handler `chmod +x chrome_crashpad_handler`
 1. Run INAV Configurator app from unpacked folder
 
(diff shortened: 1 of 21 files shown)


Jetrell:
`Pitch to Throttle ratio` is explained as a ratio, therefor it has no unit...  But it appears to be calculable in `uS`?

MrD-RC:
@Jetrell Yes, you're right. It should be in uS. But you've now got me thinking is _Ratio_ the correct term? It multiplies the `pitch2**thr**` value (which is throttle in uS) by the pitch angle. It's a multiplier rather than a ratio surely?

Jetrell:
I think it depends on how the equation is viewed.
In its simplest form. Its how many `uS` of throttle are applied per degree of Pitch offset from a level attitude. 
But due to this value also being constrained by `nav_fw_min_thr` and `nav_fw_max_thr` ... I think a Ratio would encompass all of that better.
Does that seem right?

But, is it P2T ratio or T2P ratio? 

Something I haven't tested. Is the P2T value constrained?
If someone was to entered a value like 50 (5% ). That would give a 50% throttle increase for 10 degrees of pitch... rather extreme.

Jetrell:
I think it depends if its easier for a person to view it as an equation, or an interaction between variables.
In its simplest form. Its how many `uS` of throttle are changed per degree of Pitch offset from a level attitude.
But due to this value also being constrained by `nav_fw_min_thr` and `nav_fw_max_thr` ... I think a Ratio may encompass all of that better.
Does that seem right?

But, is it a Pitch to Throttle ratio (P2T) or a Throttle value changed, as a result of Pitch (T2P)?

Something I haven't tested. Is the P2T value constrained?
If someone was to entered a value like 50 (5% ). That would give a 50% throttle increase for 10 degrees of pitch... rather extreme.

We understand why its in `uS`.. for resolution. But I wonder if it would be easier for the average person to understand.. e.g. a value ratio of 5, being written as 0.5% throttle change to 1 degree of pitch change?

PR #1560, opened 2022-06-22
reworked rc channel remapping

Woga65:
Allow remapping of rc channels 1-8 for example TAER1234, AERT2134 or 3AER12T4.
[https://github.com/iNavFlight/inav/pull/8150](url)


MrD-RC:
This was put forward before. But it is not necessary, and would confuse things. This is only really relevant for flybar helis, which are not supported by INAV.

PR #1568, opened 2022-07-01
Reset PIDs button to use applied defaults

MrD-RC:
This changes the functionality of the **PID Tuning** > **Reset PIDs** button. Before it would reset the PIDs and other settings to the firmware defaults. This change will reset to the defaults of the selected platform. For example, if you selected **Airplane with a tail**, the defaults from that platform will be used.

[[DEMO]](https://youtu.be/clOmn2CntVg)

Firmware PR https://github.com/iNavFlight/inav/pull/8185 is required for this change.

## Developer notes
All the settings are held within the `defaults_dialog.js` file. A new parameter, `allow_reset`, has been added to the platform settings. If this is set to true, it will be reset when the **Reset PIDs** button is clicked. It may be worth quad guys making sure that the correct settings are not reset.

### Can this be better?
[defaults_dialog.js: 930] One thing that I'm not happy with is to get this working, I've had to use a timeout. Though this does not effect how it works. I feel that is shouldn't need it. The problem is the MSPHelper commands are not complete before the `finalise` function is called. From what I can see, these are inside a Promise.mapSeries block, which from what I've read should be asynchronous. So the MSPHelper calls should all complete before the `.then` block. Unfortunately, this is not the case; and `finalise` is called before all the MSP commands have been sent. Hence the timeout. 4 seconds was the magic number with an F405 over USB. I added 2 seconds to allow for slower FCs or communications. Ideally, the timeout should not be used, and `finalise` called once everything is written. Any JavaScript gurus have any ideas? @elgansayer ?


js/defaults_dialog.js
```diff
@@ -20,185 +20,226 @@ helper.defaultsDialog = (function () {
         "settings": [
             {
                 key: "model_preview_type",
-                value: 3
+                value: 3, 
+                allow_reset: true
```
DzikuVx, line 24: We should not add this here. Rather create a separate list of settings that should not be changed during reset

js/defaults_dialog.js
```diff
@@ -718,7 +867,12 @@ helper.defaultsDialog = (function () {
                 GUI.tab_switch_cleanup(function () {
                     MSP.send_message(MSPCodes.MSP_SET_REBOOT, false, false, function () {
                         //noinspection JSUnresolvedVariable
-                        savingDefaultsModal.close();
+                        if (savingDefaultsModal) {
+                            savingDefaultsModal.close();
+                        }
+                        if (resettingPIDsModal) {
```
DzikuVx, line 873: This shooul;d be checked if exists with `typeof resettingPIDsModal !== 'undefined'`

js/defaults_dialog.js
```diff
@@ -751,17 +905,60 @@ helper.defaultsDialog = (function () {
...
+        }); 
+    };
+
+    privateScope.setSetting = function (key, value) {
+        MSP.send_message(MSPCodes.MSP_SELECT_SETTING, [0], false, function () {
+            mspHelper.setSetting(key, value, function() {
+                MSP.send_message(MSPCodes.MSP_SELECT_SETTING, [1], false, function () {
+                    mspHelper.setSetting(key, value, function() {
+                        MSP.send_message(MSPCodes.MSP_SELECT_SETTING, [2], false, function () {
+                            mspHelper.setSetting(key, value);
```
DzikuVx, line 944: But why 3 times? This causes everything to load 3 times and finally does not work. I don't get it.

js/defaults_dialog.js
```diff
@@ -729,14 +883,14 @@ helper.defaultsDialog = (function () {
...
     privateScope.setSettings = function (selectedDefaultPreset) {
         //Save analytics
+        var currentProfile = parseInt($("#profilechange").val());
         googleAnalytics.sendEvent('Setting', 'Defaults', selectedDefaultPreset.title);
         Promise.mapSeries(selectedDefaultPreset.settings, function (input, ii) {
             return mspHelper.getSetting(input.key);
         }).then(function () {
             Promise.mapSeries(selectedDefaultPreset.settings, function (input, ii) {
-                return mspHelper.setSetting(input.key, input.value);
+                return privateScope.setSetting(input.key, input.value);
```
DzikuVx, line 892: It shopuld use `return mspHelper.setSetting(input.key, input.value);` not this internally invented method

PR #1571, opened 2022-07-03
Fixes for modes page

MrD-RC:
- Moved to turn assist in to **flight mode modifiers** section. As it applies to fixed wing and multirotors.
- Re-established name changes for `CAMERA CONTROL` 1 to 3 when a RunCam Device is used.


MrD-RC:
No idea why those conflicts exist. This was based on release_5.1.0 branch, not master.

PR #1590, opened 2022-07-29
Add version number to CLI save filename

MrD-RC:
This PR adds the version number to the CLI filename.

The version number is very handy for keeping track of the diff and dump saves. To be honest, I got bored of adding it all the time manually 😉

### DIFF
--- a/js/helpers.js
+++ b/js/helpers.js
@@ -42,12 +42,27 @@ function generateFilename(prefix, suffix) {
 
     if (CONFIG) {
         if (CONFIG.flightControllerIdentifier) {
-            filename = CONFIG.flightControllerIdentifier + '_' + filename;
-        }
+            filename = CONFIG.flightControllerIdentifier + '_' + CONFIG.flightControllerVersion + "_" + filename;
+        } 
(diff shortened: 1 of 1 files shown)


js/helpers.js
```diff
@@ -42,13 +42,28 @@ function generateFilename(prefix, suffix) {
...
-            filename = CONFIG.flightControllerIdentifier + '_' + filename;
-        }
+            filename = CONFIG.flightControllerIdentifier + '_' + CONFIG.flightControllerVersion + "_" + filename;
+        } 
         if (CONFIG.name && CONFIG.name.trim() !== '') {
             filename = filename + '_' + CONFIG.name.trim().replace(' ', '_');
         }
+    } else {
+        // Attempt to get the version number and 
+        MSPHelper.getSetting("version").then(function (data) {
```
DzikuVx, line 52: @MrD-RC are you sure this even works? When CLI is on, MSP on that port no longer works

MrD-RC: Test 1: Same version
image

Test 2: 4.1.0
image

Test 3:
image

You're right. It's using CONFIG each time. I'll update this as I've seen another improvement.

PR #1593, opened 2022-07-29
Cleanup Autolaunch settings

b14ckyy:
Removed Settings (should not be changed by user, but kept in CLI): 
Launch Velocity
Launch Accel
Launch Detect Time

Moved Launch and Idle throttle together

### DIFF
--- a/js/defaults_dialog.js
+++ b/js/defaults_dialog.js
@@ -395,7 +395,7 @@ helper.defaultsDialog = (function () {
             },
             {
                 key: "nav_fw_launch_max_angle",
-                value: 90
+                value: 45
             },
             {
                 key: "nav_fw_launch_motor_delay",
@@ -610,7 +610,7 @@ helper.defaultsDialog = (function () {
             },
             {
                 key: "nav_fw_launch_max_angle",
-                value: 90
+                value: 75
             },
             {
                 key: "nav_fw_launch_motor_delay",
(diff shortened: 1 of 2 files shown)


js/defaults_dialog.js
```diff
@@ -393,6 +393,22 @@ helper.defaultsDialog = (function () {
                 key: "nav_wp_radius",
                 value: 5000
             },
+            {
+                key: "nav_fw_launch_max_angle",
+                value: 90
```
MrD-RC, line 398: I do think 90 degrees is a bit too steep. Most planes aren't overhead launched, so I think a 45-60 degree angle would be better for safety. Same applies to the other fixed wing defaults. In fact, you would never launch a plane with a tail overhead. So 45 degrees is plenty for them.

js/defaults_dialog.js
```diff
@@ -592,6 +608,22 @@ helper.defaultsDialog = (function () {
...
```
MrD-RC, line 613: I feel this would be better at 45-60 degrees.

MrD-RC (changes requested):

js/defaults_dialog.js
b14ckyy: You are right with the Plane with tail, that makes no sense. 
As a compromise for wings: Would you agree on a 75° limit for wings and 45° limit for planes with tail? 

for wings it makes also sense to keep it lower to avoid a power up if the plane is thrown in a vertical position where a gimbal lock can occur and you never know what the resulting direciton will be. 75° should be plenty for over head and also reduce the chance of a gimbal flip.

MrD-RC: Sounds good.

PR #1603, opened 2022-08-13
Fixed and enhanced LC display

### LINKED ISSUE #1602: Logic ID operand limit
### Discussed in https://github.com/iNavFlight/inav/discussions/8315

<div type='discussions-op-text'>

<sup>Originally posted by **rts18** August 13, 2022</sup>
My inquiry is related to the number of logic ID's that can be accessed under operand A or B.  The limit is 31. While the activator can handle up to 63.
Is this by design to save flash? Or possibly something that has been over looked?</div>

MrD-RC:
Fixed an issue where only the first 32 logic conditions were showing as options for operands. This is now dynamic, based on the maximum number of logic conditions.

Fixes #1602 (fc.js:1252)

I also enhanced the display of the **Active** column of both **Programming** and **Mixer**. The select boxes will now only show Logic Conditions that are enabled.

image

image

Logic Conditions that were selected and then later disabled are shown in grey
image

### DIFF
--- a/js/fc.js
+++ b/js/fc.js
@@ -1249,7 +1249,7 @@ var FC = {
             4: {
                 name: "Logic Condition",
                 type: "range",
-                range: [0, 31],
+                range: [0, (LOGIC_CONDITIONS.getMaxLogicConditionCount()-1)],
                 default: 0
             },
             5: {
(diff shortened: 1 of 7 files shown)


rts18:
@MrD-RC Thank you sir for the reply and the incredibly quick fix.

PR #1627, opened 2022-10-01
Added support for DJI wtfos MSP-OSD full screen 60x22 OSD

jeffhendrix:
Changes to support new feature in iNav.
Feature will allow DJI goggles that have the wtfos installed to use the full screen OSD 60x22
New peripheral was added to the Ports tab "DJI WTF" to enable new mode.
The MSP_OPTIONS subcommand will then send a 2 to the goggles to enable the full screen mode.
OSD tab added the video format option DJIWTF and the preview is enlarged to allow placing elements in the full 60x22 grid.

### DIFF
--- a/tabs/osd.html
+++ b/tabs/osd.html
@@ -31,10 +31,14 @@
                 </div>
                 <div class="display-layout">
                     <div class="col right">
-	                    <div class="left_43_margin"></div>
-    	                <div class="right_43_margin"></div>
-                        <div class="preview">
-                        </div>
+                        <div class="hd_43_margin_left"></div>
+                        <div class="hd_43_margin_right"></div>
+                        <div class="hd_3016_box_top"></div>
+                        <div class="hd_3016_box_bottom"></div>
+                        <div class="hd_3016_box_left"></div>
+                        <div class="hd_3016_box_right"></div>
+                        <div class="pal_ntsc_box_bottom"></div>
+                        <div class="preview"></div>
                     </div>
                 </div>
             </div>
@@ -45,6 +49,12 @@
                     </div>
                     <div class="spacer_box">
                         <div class="video-types"></div>
+                    </div>
+                    <div class="spacer_box settings">
+                        <div for="osd_video_show_guides" class="helpicon cf_tip" data-i18n_title="osd_video_HELP"></div>
+                        <label id="videoGuides">
+                           <span data-i18n="osd_video_show_guides"></span>
+                        </label>
                     </div>
                 </div>
                 <div class="gui_box grey settings-container">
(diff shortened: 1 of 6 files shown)


DzikuVx (changes requested):
@jeffhendrix I left my comment in https://github.com/iNavFlight/inav/pull/8434#pullrequestreview-1127497368

But the same applies

DzikuVx:
I rechecked the code and `displayport_hdzero_osd.c` and `displayport_djiwtf_osd.c` are basically the same. They differ mostly with different canvas size

tabs/osd.html
```diff
@@ -46,6 +50,12 @@ <h1 class="tab_title" data-i18n="tabOSD"></h1>
                     <div class="spacer_box">
                         <div class="video-types"></div>
                     </div>
+                    <div class="spacer_box settings">
```
MrD-RC, line 53: This is a nice touch. Though it would be much better is the state was stored in Configurator. You can use the chrome storage for this.

tabs/osd.js
```diff
@@ -2354,15 +2365,36 @@ OSD.GUI.checkAndProcessSymbolPosition = function(pos, charCode) {
...
+    } else {
+        if (OSD.constants.VIDEO_TYPES[OSD.data.preferences.video_system] == 'HDZERO' || OSD.constants.VIDEO_TYPES[OSD.data.preferences.video_system] == 'DJIWTF') {
+            OSD.data.preferences.video_system = OSD.constants.VIDEO_TYPES.indexOf('AUTO')
+            OSD.updateDisplaySize();
+            OSD.GUI.saveConfig();
+        }
+    }
+
+    for (var i = 0; i < OSD.constants.VIDEO_TYPES.length; i++) {
+        if ((OSD.constants.VIDEO_TYPES[i] != 'HDZERO' && OSD.constants.VIDEO_TYPES[i] != 'DJIWTF') || OSD.data.isMspDisplay)
```
MrD-RC, line 2388: There is a bug that if you have DJIWTF selected and saved. If you leave the OSD page and return, HDZERO is selected instead. Is should show that you are using DJIWTF. I think the bug is in the above IF statement. I think it should be something like:

```
if ( OSD.data.isMspDisplay || 
     (!OSD.data.isMspDisplay && (OSD.constants.VIDEO_TYPES[i] === 'HDZERO' || OSD.constants.VIDEO_TYPES[i] === 'DJIWTF'))
   )
```

MrD-RC (changes requested):

tabs/osd.js
jeffhendrix: I was working, then I kept adding changes. Must have broke it in the process. Will fix.

tabs/osd.html
jeffhendrix: Tell me more about this and how to make it work (html, css, js, etc. are not my 1st languages)

jeffhendrix:
I have a question (looking for opinions).
When MSPPort is selected, should I remove the AUTO/PAL/NTSC options?
I don't know if HDZERO supports those layouts or not. 
DJI (with wtf) does support the smaller layouts (it puts everything in the middle of the screen unless you have fakehd option set). With this change, don't know why anybody would select the smaller OSD layout, so debating is those should be removed when in MSP mode.

Issue #636, opened 2019-01-08
New symbols for the estimated remaining distance and time


shellixyz:
See https://github.com/iNavFlight/inav-configurator/issues/605#issuecomment-451788497

kardon18:
Other symbols that would be nice to have added. Would be -

![remaining time and distance](https://user-images.githubusercontent.com/43928973/50636753-6fbd1e00-0fab-11e9-97ed-7d288b561cf8.PNG)

At present they use the same symbols as _Flight Time_ and _Total Distance_. Which can be confusing, if you have both sets on the OSD at the same time.

Although I am aware this would require changes to the code, and a location to be assigned for these symbols in every font.

These two give the general idea -

![182](https://user-images.githubusercontent.com/43928973/50637280-7187e100-0fad-11e9-9fbf-53ffe0681bcb.png)
![156](https://user-images.githubusercontent.com/43928973/50637281-7187e100-0fad-11e9-9434-640673e6d70e.png)

The file here includes :

    1. Flight time remaining in - hours / minutes

    2. Distance remaining in - Km / Meters / Miles / Feet


There are six glyphs in total, for each font.
With each gylph having its own characteristics style for that font.


[Remaining time_Remaining distance_glyphs.zip](https://github.com/iNavFlight/inav-configurator/files/2731024/Remaining.time_Remaining.distance_glyphs.zip)

PR #1656, opened 2022-11-26
Add correct symbols for remaining time and distance

MrD-RC:
Fixes https://github.com/iNavFlight/inav-configurator/issues/636

Adds symbols for remaining flight distance and flight time OSD elements. These replace the current fly time and total distance symbols, which are not specific.

I have added temporary symbols which can be improved.

Firmware element https://github.com/iNavFlight/inav/pull/8584

image


b14ckyy:
I suggest something that represents, that these values are the time until RTH has to be engaged to make it home. As it is not the actual remaining flight time but the remaining flight time and distance before the pilot has to start the way home. (unless this was changed?)

PR #1660, opened 2022-12-04
OSD Ground Course and Cross Track Error fields

breadoven:
Support for https://github.com/iNavFlight/inav/pull/8604.


Jetrell:
@breadoven I'm not sure if you're aware. A symbol for this function is already in there? `252`

Jetrell:
I'm not sure if you're aware. A symbol for this function is already in there? `252`
@breadoven The reason its there. Is because it was added to the HD font. That was given to Caddx for the Avatar goggle. In anticipation of this feature being added one day. This goes for the 4 symbols in that location.

tabs/osd.js
```diff
@@ -111,6 +111,8 @@ SYM.GLIDE_MINS = 0xD5;
 SYM.GLIDE_RANGE = 0xD4;
 SYM.FLIGHT_MINS_REMAINING = 0xDA;
 SYM.FLIGHT_DIST_REMAINING = 0x167;
+SYM.GROUND_COURSE = 0xDC;
+SYM.TRACK_ERROR = 0xDD;
```
MrD-RC, line 115: Please use 0xFC for course track error symbol.

breadoven: OK, missed those new icons. Added a few days ago by the looks of it ? Is there any reason they aren't included in `iNav character map.md` ? I take it 251, the mountain with an arrow sticking out of it, is terrain following ?

MrD-RC: Not being in the .md file is my mistake. Yes, the mountain with and arrow is terrain following.

MrD-RC: Yes, thank you 👍🏻

MrD-RC (approved):

PR #1667, opened 2022-12-09
Fix-cp-error

b14ckyy:
fix little c&p error witht he new settings


b14ckyy:
there was more in VSCode than intended. Will redo.

PR #1669, opened 2022-12-10
Added "Disable GPS Sensor Fix" logic condition

RomanLut:
This submit accompanies: https://github.com/iNavFlight/inav/pull/8347
and adds "Disable GPS sensor fix" logic condition which can be used for testing estimated GPS fix functionality.


Jetrell:
Any chance of this being added for 7.0 ?  
I think this feature would also be useful for testing and tuning fixedwing Emergency landing.. The default settings are often never changed and they leave a bit to be desired on some airframes.
But its difficult to safely turn off the GPS and simulate an emergency landing condition in RTH.

DzikuVx:
@RomanLut if you could resolve conflight, I'd like to merge it

RomanLut:
> @RomanLut if you could resolve conflight, I'd like to merge it

@DzikuVx Resolved.

PR #1692, opened 2023-01-15
Add BFHDCOMPAT

mmosca:
Add HD support to BF compatibility mode.


mmosca:
Still need to confirm canvas size

PR #1697, opened 2023-01-17
Add user actions to Mission Control

MrD-RC:
Mission Control updates
- Added user actions
- Added hover over text for P1 and P2 in actions

[Demo](https://youtu.be/P9X_K9kgLLQ)

Examples of hover text
image

image

### DIFF
--- a/src/css/tabs/mission_planer.css
+++ b/src/css/tabs/mission_planer.css
@@ -522,3 +522,7 @@
 .tab-mission-control .valueLegend{
     float:right;
 }
+
+.tab-mission-control .userActionContainer {
+    display: inline-block;
+}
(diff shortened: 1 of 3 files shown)


stronnag:
I tested User Actions will a prior example with WP#11 and WP#14 having user actions, one with relative and one with AMSL altitudes.

Screenshot from 2023-01-17 21-44-46
and
Screenshot from 2023-01-17 21-45-34

Which in the configurator, look like:
Screenshot from 2023-01-17 21-41-57
and
Screenshot from 2023-01-17 21-41-44

Ignoring implementation details like numbering philosophy and different altitude sources (try a 120 WP abs alt mission one day ....); they are functionally identical and generate equivalent mission files.

Setting and clearing bits as expected.

stronnag (approved):
Tested and functioning as expected.

MrD-RC:
Thanks @stronnag

PR #1722, opened 2023-02-14
Mission Control: Bug fixes related to P3 field

### LINKED ISSUE #1721: WP "user actions" breaks waypoint navigation
Short video demonstrating issue:
https://user-images.githubusercontent.com/8989151/218552591-65620b35-67e2-451c-ae52-334931ab7444.mp4
Text version - mission waypoint with user action set has it's altitude set to MSL+preset altitude, while AGL setting is selected.
This was registered as issue on another branch but it looks like firmware has nothing to do with it:
https://github.com/iNavFlight/inav/issues/8798

In short - looks like INAV-configurator incorrectly parses bit0 in P3 field and and this is reason why alt is set to wrong value.

MrD-RC:
Fixes #1721 

- Added const to make bitfield stuff more readable
- changed `checkAltElevationSanity()` as it was treating P3 as a Boolean. Some times this function was called with explicitly not using absolute altitude. This is still possible. The function can accept a Boolean _or_ the P3 bitfield.
- Fixed issue where alt and user action checkboxes could not be unset.
- Added **.mission** filter to the load mission from file dialog box.

### DIFF
--- a/tabs/mission_control.js
+++ b/tabs/mission_control.js
@@ -2123,12 +2123,13 @@ TABS.mission_control.initialize = function (callback) {
                 }
 
                 P3Value = TABS.mission_control.setBit(P3Value, MWNP.P3.ALT_TYPE, $('#pointP3Alt').prop("checked"));
-                selectedMarker.setP3(P3Value);
                 (async () => {
                     const elevationAtWP = await selectedMarker.getElevation(globalSettings);
                     $('#elevationValueAtWP').text(elevationAtWP);
                     var altitude = Number($('#pointAlt').val());
                     if (P3Value != selectedMarker.getP3()) {
+                        selectedMarker.setP3(P3Value);
+                        
                         if ($('#pointP3Alt').prop("checked")) {
                             if (altitude < 0) {
                                 altitude = settings.alt;
--- a/tabs/mission_control.js
+++ b/tabs/mission_control.js
@@ -2811,13 +2811,14 @@ TABS.mission_control.initialize = function (callback) {
     }
 
     /* resetAltitude = true : For selected WPs only. Changes WP Altitude value back to previous value if setting below ground level.
-     ^ resetAltitude = false : changes WP Altitude to value required to give ground clearance = default Altitude setting */
-    function checkAltElevSanity(resetAltitude, checkAltitude, elevation, P3Datum) {
+     ^ resetAltitude = false : changes WP Altitude to value required to give ground clearance = default Altitude setting
+     ^ AbsAltCheck : check value for whether or not to use absolute altitude. This can be the P3 bitset or excplicitly set to true or false */
+    function checkAltElevSanity(resetAltitude, checkAltitude, elevation, AbsAltCheck) {
         let groundClearance = "NO HOME";
         let altitude = checkAltitude;
-        let altUsingAbsolute = (typeof P3Datum == "boolean") ? P3Datum : TABS.mission_control.isBitSet(P3Datum, MWNP.P3.ALT_TYPE);
+        AbsAltCheck = (typeof AbsAltCheck == "boolean") ? AbsAltCheck : TABS.mission_control.isBitSet(AbsAltCheck, MWNP.P3.ALT_TYPE);
 
-        if (altUsingAbsolute) {
+        if (AbsAltCheck) {
             if (checkAltitude < 100 * elevation) {
                 if (resetAltitude) {
                     alert(chrome.i18n.getMessage('MissionPlannerAltitudeChangeReset'));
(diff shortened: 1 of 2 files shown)


tabs/mission_control.js
```diff
@@ -2100,7 +2109,7 @@ TABS.mission_control.initialize = function (callback) {
             if (selectedMarker) {
                 selectedMarker.setP2(Number($('#pointP2').val()));
                 mission.updateWaypoint(selectedMarker);
-                mission.update(singleMissionActive());
+                mission.update(singleMissionActive());setAlt
```
breadoven, line 2112: setAlt ?

MrD-RC: Damn cat. I was performing a find, and she trod on the keyboard. I'll get that sorted.

tabs/mission_control.js
```diff
@@ -2110,13 +2119,10 @@ TABS.mission_control.initialize = function (callback) {
...
                 if (disableMarkerEdit) {
-                    changeSwitchery($('#pointP3Alt'), TABS.mission_control.isBitSet(P3Value, 0));
+                    changeSwitchery($('#pointP3Alt'), TABS.mission_control.isBitSet(P3Value, MWNP.P3.ALT_TYPE));
                 }
 
-                P3Value = TABS.mission_control.setBit(P3Value, 0, $('#pointP3Alt').prop("checked"));
-                selectedMarker.setP3(P3Value);
-
-                P3Value = TABS.mission_control.setBit(P3Value, 0, $('#pointP3Alt').prop("checked"));
+                P3Value = TABS.mission_control.setBit(P3Value, MWNP.P3.ALT_TYPE, $('#pointP3Alt').prop("checked"));
```
breadoven, line 2125: Line 2131: this needs to compare the BIT 0 value of P3 on entering the function (line 2119) with the BIT 0 value after updating with the state of the checkbox. So P3Value can't be redefined on line 2125, needs to be called something else.

`P3Value != selectedMarker.getP3()` will always be false as this is now although it won't work anyway because it's not checking BIT 0 specifically.

tabs/mission_control.js
```diff
@@ -2942,7 +2950,8 @@ TABS.mission_control.isBitSet = function(bits, testBit) {
 
 TABS.mission_control.setBit = function(bits, bit, value) {
     if ((initParam3 & (1 << bit)) != 0) {
```
breadoven, line 2952: What is the purpose of `initParam3 `? I tested this with just:

    `return value ? bits |= (1 << bit) : bits &= ~(1 << bit);`

and it seemed to work as expected.

tabs/mission_control.js
```diff
@@ -2809,7 +2815,9 @@ TABS.mission_control.initialize = function (callback) {
     function checkAltElevSanity(resetAltitude, checkAltitude, elevation, P3Datum) {
```
breadoven, line 2815: Better to change `P3Datum `to `P3Value `since it no longer refers to a datum any more unlike previously ?

MrD-RC: I've renamed the variable.

tabs/mission_control.js
MrD-RC: This is needed in the initial setup of the checkboxes. Without this test, when the checkboxes are set up (lines 1823 -1827). setting the correct value for the checkbox triggers the onChange event. Which then changes the value. This test stops that from happening.

tabs/mission_control.js
MrD-RC: I've moved the storage of the newly calculated P3 to after the check, if they are different.

MrD-RC:
This seems to be correct now, after moving the storing of P3 around line 2125. Changing the toggle now changes the alt in the _Alt (cm)_ box above.

image

image

[test mission.mission.txt](https://github.com/iNavFlight/inav-configurator/files/10739719/test.mission.mission.txt)

tabs/mission_control.js
breadoven: Got to admit I'm not seeing that. With `initParam3 `removed it doesn't seem to make any difference. The onChange event triggers but it just sets P3 again to the same correct setting. Or is this just to avoid unwanted trigger events which do seem to happen a lot with JS regardless.

breadoven (approved):
Other than the slightly odd `setBit `function it all seems to work as expected now so all good.

MrD-RC:
> Other than the slightly odd `setBit `function it all seems to work as expected now so all good.

I tried your function and it worked fine. So I swapped it in as it it neater too. I'm still fairly new to all the bitwise stuff. Also, I remembered I found a bug in it which I fixed. That could have also been causing the unwanted behaviour. 

Either way. This is all working nicely now, so I'll get it merged, ready for RC3.

PR #1741, opened 2023-03-21
Small update on fw_level_pritch_trim field

OptimusTi:
Changed input to allow decimal values. It was only saving 0.


MrD-RC:
I think the problem is more that in settings.yaml, it's not defined as a float. The units stuff should take care of the step correctly.

OptimusTi:
Setting fw_level_pritch_trim to float on settings.yaml didn't work. Configurator only takes integers (deg). Not sure if this is really needed, so I may close this if I can't find a solution.

PR #1756, opened 2023-04-14
Add i18n to the LED_strip Tab

YI-BOYANG:
This PR adds all the text in LED_strip.html to i18n.

**Why would I want to do this?**

Since INAV configurator only supports English language for the time being, since INAV4.0, I have provided a local built Chinese version of the configurator for Chinese players, but I found that there is still a lot of text in the code that has not been added to i18n. Every time INAV releases a new version of the configurator, I have to spend a lot of time manually changing the text to Chinese before building.

This may be a relic of history, but if INAV eventually wants to support multiple languages like Betaflight, these texts will have to be added to i18n. So why not start preparing for multilingualism now?

For my first attempt, I chose the LED_strip Tab because the Betaflight configurator code gives me a good reference, since INAV and Betaflight both originate from Cleanflight.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -161,7 +161,7 @@
     "connectionBleType": {
         "message": "BLE device type: $1"
     },
-    "connectionBleNotSupported" : {
+    "connectionBleNotSupported": {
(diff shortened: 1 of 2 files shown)


mmosca:
This looks good. I'll test it later and report back.

mmosca:
Looks good. I didn't find any regressions.

PR #1765, opened 2023-04-24
move current meter scale and offset to battery profile

RomanLut:
Accompanying submit for:
https://github.com/iNavFlight/inav/pull/9004

to display current meter scale and offset inputs with yellow background color (indication of battery profile setting).


RomanLut:
Closing as not accepted

PR #1767, opened 2023-04-25
update SITLs (and cygwin dll)

stronnag:


stronnag:
Placeholder for 6.1.0 release. If there are more firmware updates prior to release, this PR will be updated.

stronnag:
Note that the 32bit builds (Linux and Windows) are broken in that they install **64bit** executables.
This PR does not address this extant bug.

PR #1768, opened 2023-04-25
[SITL] Update binaries to 6.1

Scavanger:


Scavanger:
Ups, overlooked #1767

PR #1774, opened 2023-05-03
[SITL] update SITLs for inav/#9029

stronnag:


stronnag:
The Windows build appears somewhat unstable with inav/#9029, withdrawing for now

PR #1783, opened 2023-05-16
Add the text from the HTML and the string from the GUI console to i18n

YI-BOYANG:
In this PR, I've added the vast majority of the text from the HTML and strings from the GUI console to the i18n and everything works fine running on my computer.


shota3527:
支持一个

PR #1799, opened 2023-06-30
fixed: magnetometer align tool does not take board orientation into a…

RomanLut:
**IS:** Magnetometer alignment tool does not take FC alignment into consideration.

**SHOULD:** Magnetometer alignment is specified relative to FC. Magnetometer model and axes should be rendered properly, showing correct orientation relative to UAV frame.

Fixed

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -2038,7 +2038,7 @@
         "message": "Magnetometer Alignment"
     },
     "magnetometerHelp": {
-        "message": "Adjust the magnetometer position until you get the same on the quad.<br/><strong>Note:</strong> If you have changed your flight controller alignment, make your adjustment relative to the FC position and not to the UAV position "
+        "message": "Adjust the magnetometer position until you get the same on the quad.<br/><strong>Note:</strong> Magnetometer alignment is relative to board. Make sure to align board first."
     },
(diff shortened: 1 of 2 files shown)


Jetrell:
@RomanLut Firstly, I'd like to say thanks for your willingness to make some changes here.
I only got a chance to test it today.

I've had many questions on INAV Discord over the last year, concerning this matter. And the first thing people will do is rotate the GPS/Mag on the GUI, to the same position its mounted on the quad.. Even if the FC board is offset. 
I tried this tool on two quads using this method. Both have their FC boards either rotated or flipped. I understand your reasoning behind these changes. Your math and idea work well... And it does align the mag with respect to the FC. But it doesn't appear to be very intuitive with respect to the GUI.  i.e. having the GPS/Mag facing a different direction than its mounted on the quad.

_What would be ideal and is expected of this tool, buy new users.. Is if the GUI had a representation of a flight control board (shown below the quads virtual image) with a forward arrow on one side. So users can rotate that image to suit the way their FC is mounted in the quad._ Then the math will automatically set `board` and `mag` alignment based on both of these parameters.

Without meaning to sound critical or ungrateful.. This method using the GUI, unfortunately exchanges one means of setup confusion for another.. Sorry.

PR #1801, opened 2023-07-04
implemented fc display in mag align helper tool

RomanLut:
image


DzikuVx:
Nice!

Jetrell:
@RomanLut Thanks again for your work on this alignment tool.. It looks great.

Although I've found something else. Can I ask you to check the correlation between the `Orientation Preset` and the `align_mag_yaw` referenced from the slider.

At present, using the Slider to set `mag_align_yaw`, via the GUI mag arrow, appears also to be based on the  `Orientation Preset`.
This may seem correct, and is if the FC board is mounted default.. But when the FC board is flipped or placed in another orientation that differs from the mag. Then it appears the `Orientation Preset`, sets the magnetometer chips mounting position, for magnetic heading, based on the front end of the quad.
While `align_mag_yaw` references the compass mounting position with respect to the FC's `align_board_yaw` setting. Which in my case both are zero. Because both arrows are facing forwards on the quad, with the FC flipped.

The working quad, this image below is taken from, has these settings in its CL. With the FC board flipped on the roll axis, and the mag arrow facing forward in a default position.

`align_board_roll = 1800`
`align_board_pitch = 0`
`align_board_yaw = 0`
`align_mag = CW270` (no flip, because the FC board is flipped)
`align_mag_pitch = 0`
`align_mag_roll = 0`
`align_mag_yaw = 0
`

Notice how the Yaw slider, and the direction the compass is facing, does not align with these settings.. And if I move the Yaw slider, to change the compass arrow position to face forward, as it is on the quad. Then the IMU and Mag will clash in real time.
Its also interesting to see the Roll slider is incorrectly showing 180.


Image from my setup

Jetrell:
I suppose the biggest issue with this type of tool.. Is it has to be ALL or nothing.. Because there are multiple ways to come to the end result. Usually two ways. But there is also a third, that looks right to a new user. But in the end will lead to  slow drift problems.

* If people can rotate the FC GUI board arrow to what ever orientation it is in the quad.. And have a selection for them to enter what side of the board the IMU is located. (top side with the arrow, or bottom side). 
 
* Then also rotate the compass GUI arrow to what ever orientation its facing in the quad. And select the Orientation preset for that specific GPS/Mag unit. (CW 270 flip in 99% of cases).
_Then the rest should be best worked out in the tool.. By one means._

But as is, there is still room for guess work and interpretation.
The confusion of `CW 270` default meaning the compass unit arrow is actually facing forward. Leaving many people thinking they have to work that out with respect to FC orientation... And Mag `Flip` being just as relevant to which side the IMU is mounted on the flight controller...

rts18:
Roman, I've checked out these changes, and there appears to be conflicting calculations between `align_mag` and `align_mag_yaw.` 
* `Align_mag` or _orientate preset_ is obviously based on the hardware designers component mounting of the magnetometer chip, and referenced by the arrow on the device.  This setting should only be adjusted to suit the hardware specifications.
* While `align_mag_yaw` is a software setting used to adjust the mounting position of the device in the craft. Which uses the compass arrow as a reference marker for this purpose.

The problem I've observed when the flight controller board has a different Yaw mounting position. And the _Yaw slider_ is moved to set the compass arrows direction according to its mounting orientation in the craft. Is the _orientation preset_ become shaded and requires it to be changed to line up with `align_mag_yaw.` With it also locking up the Configurator.
This makes no sense, because the _orientation preset_ should not be changed from the manufactures specification. But if it isn't changed the Gyro and Mag will clash. 
And once _orientation preset_ is changed to suit the tools requirements, it becomes a mess when trying to workout the actual _align_mag_ position based on the arrow and magnetometer hardware..

RomanLut:
I hope now everything works correctly:
https://github.com/iNavFlight/inav-configurator/pull/1806

PR #1803, opened 2023-07-12
Add more baud rates to msp connection

mmosca:
Is it faster? Is it just a number?


stronnag:
Shouldn't this match the list in `src/main/io/serial.c` for maximum placebo effect?

mmosca:
> Shouldn't this match the list in `src/main/io/serial.c` for maximum placebo effect?

Easy change. I was picking values from asm-generic/termbits.h in linux. :)

stronnag:
And the final piece would be  the Ports tab, such that

```
# serial
serial 20 32769 921600 115200 0 115200
serial 0 17 230400 115200 0 115200
```
doesn't result in 

image

mmosca:
Should be all ready for maximum placebo effect.

image

mmosca:
- [ ]

PR #1806, opened 2023-07-21
magnetometer alignment tool => alignment tool

RomanLut:
I hope now alignment tool works correctly and is  clear for the user.

image

1) Removed board yaw and magnetometer alignment preset from Configuration tab
2) Renamed Magnetometer alignment tab to Alignment tool
3) Added board alignment roll, pitch, yaw sliders to Aligmnent tool tab
4) Fixed multiple bugs related to roll/pitch/yaw magnetometer orientation
5) Align_mag preset is relative to FC, but alignment_mag_roll/pitch/yaw are NOT!
6) Clear indication which settings are used: align_mag or align_mag_xxx
7) Align_mag_xxx are caclulated from preset + FC orientation, allowing to switch from preset to explicit angles easily

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -900,18 +900,6 @@
     "configurationSerialRXHelp": {
         "message": "<strong>Note:</strong> Remember to configure a Serial Port (via Ports tab) for the serial receiver"
     },
-    "configurationBoardAlignment": {
-        "message": "Board and Sensor Alignment"
-    },
-    "configurationBoardAlignmentRoll": {
-        "message": "Roll Degrees"
-    },
-    "configurationBoardAlignmentPitch": {
-        "message": "Pitch Degrees"
-    },
-    "configurationBoardAlignmentYaw": {
-        "message": "Yaw Degrees"
-    },
     "configurationSensorAlignmentMag": {
         "message": "MAG Alignment"
     },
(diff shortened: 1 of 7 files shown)


rts18:
Excellent work Roman!! This seems to cover all the bases, making the GUI more user friendly, with all the related settings now being found in this tab.
I'm away presently, **so I can't test it**. 

If I pick up on anything its this.
_5) Align_mag preset is relative to FC, but alignment_mag_roll/pitch/yaw are NOT!_

It may be your wording, from a calculation perspective. But I thought it was different from a hardware perspective. I could be wrong.

Something like this : `Align_mag` preset is relative to the GPS/compass manufactures hardware alone. 
With the _Align_mag_X_ sliders degrees, only being used as a GUI position reference, based on the direction the user faces the compass arrow in their aircraft. 
Then the tool will calculate and modify the actual `align_mag_roll/pitch/yaw` settings the flight software requires, relative to the FC `Align_board_roll/pitch/yaw`.

 I think orientation preset is a pain to be truthful. It was only the laziness of the first designers who made a GPS with an integrated magnetometer. Who didn't mount the magnetometer chip with CW0(flip) being the direction of their orientation arrow, with respect to the magnetometer chip; instead of default CW270(flip). 
Now all the manufactures seem to follow suit, continuing the confusion. With most using this rule. But the _Orientation Preset_ is still required for the manufactures that actually make their hardware correctly. Or even do something different again, by having their arrow facing forward at CW 90(flip) or CW180(flip).

You may already know this, and have implemented it. But I thought I'd spell it out, just in case. Before I get a chance to test it.

mmosca:
Improving the usability there was in my todo list. We get plenty of questions on this on discord :)

Another item that could help user validate the alignment is to add the heading there somewhere, users can validate that they are pointing N/E/S/W correctly.

mmosca:
Functionality wise, it is great now!

Is it possible to change the layout to have the 3d model side by side with the sliders when the screen is big enough?

The way the layout works right now I end up having to scroll up and down to see the sliders and check the values on the model, and the sliders don't really need to be that wide.

RomanLut:
> Is it possible to change the layout to have the 3d model side by side with the sliders when the screen is big enough?
Changed layout, updated screenshot in the first post

tdemiralp:
Hi. Very usefull tool. is it possible to add reset button to restrat alignment tool. Thanks.

PR #1808, opened 2023-07-24
Add HUD settings to OSD tab

mmosca:
INAV radar may soon get a poppulary boost as  https://github.com/FormationFlight/FormationFlight enables ExpressLRS receiver hardware to work as INAV radar unit, and ExpressLRS may also soon include radar support https://github.com/ExpressLRS/ExpressLRS/pull/2270. There is also work to add support for it in both Ardupilot (https://github.com/MUSTARDTIGERFPV/ardupilot) and 
Betaflight (https://github.com/MUSTARDTIGERFPV/Betaflight).

Currently, setup requires the user to both enable the Crosshair on the osd, as well as use the cli to to allow more than 0 radar elements on the OSD and tweak other parameters.

This pull request will add HUD settings section to the OSD tab with the options most likely to be used for a new setup.

image

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -3431,12 +3431,42 @@
     "osd_hud_radar_disp_help": {
         "message": "This is used for INAV Radar/FormationFlight. 0 disables this feature."
     },
+
+    "osd_hud_radar_range_min": {
+        "message": "Minimum radar range"
+    },
+    "osd_hud_radar_range_min_help": {
+        "message": "Radar aircrafts closer than this will not be displayed in the HUD."
+    },
+
+    "osd_hud_radar_range_max": {
+        "message": "Maximum radar range"
+    },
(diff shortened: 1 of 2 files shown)


CREASE-gum-EAR:
Right now only letters A,B,C,D get assigned to each radar element. FormationFlight supports aircraft name, can this get added to INAV? You seamed to be the person to do it?

MrD-RC:
@CREASE-gum-EAR this should be raised as a separate issue in the INAV Firmware repo. This is not a configurator issue. It is also dependent on if the pilot name is accessible over MSP for ESP32 radar and formation flight. This would also increase the cluttrer on the OSD.  So it’s something that should be optional.  

Another update is that ESP32 Radar, I also believe FormationFlight, uses the craft name as the pilot name. There is now an actual pilot name element, which would be better to use.

PR #1812, opened 2023-07-27
Break up HUD elements into individual OSD elements (configurator changes)

mmosca:
Break up hud elements as individual HUD elements so they can be turned on/off by selecting a different OSD profile.


mmosca:
Haven't tested yet, but should make it easier to test iNavFlight/inav#9199

mmosca:
This is what the new section will look like.
image

Potentially, the crosshair can be moved here as well.

MrD-RC:
I would probably change **INAV Radar** to **ESP32 Radar / Formation Flight**. 

It hasn’t been called INAV Radar for a while.

mmosca:
> I would probably change **INAV Radar** to **ESP32 Radar / Formation Flight**.
> 
> It hasn’t been called INAV Radar for a while.

That is an easy change. What about changing it to MSP Radar? Now it is more about the protocol, as there will soon be 3 possible sources for this data. (ESP32, Formation Flight and ExpressLRS).

mmosca:
Also, should we move the crosshairs to the HUD section so all related elements are grouped together?

PR #1814, opened 2023-07-29
Add camera uptilt to HUD settings

mmosca:
camera uptilt was missing from the previous pull request.

image

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -3453,6 +3453,12 @@
     "osd_hud_wp_disp_help": {
         "message": "Number of Wayponts to show on screen. 0 disables this feature."
     },
+    "osd_camera_uptilt": {
+        "message": "Camera uptilt"
+    },
+    "osd_camera_uptilt_help": {
+        "message": "Set the camera uptilt for the FPV camera in degres, positive is up, negative is down, relative to the horizontal. Used for correct display of HUD items and AHI (when enabled)."
+    },
     "osd_camera_fov_h": {
         "message": "Camera Horizontal FOV"
     },
--- a/tabs/osd.html
+++ b/tabs/osd.html
@@ -385,8 +385,11 @@
                                 <input type="number" data-step="1" data-setting-multiplier="1" data-setting="osd_camera_fov_v" class="toggle update_preview" data-live="true">
                                 <span data-i18n="osd_camera_fov_v"></span>
                             </label>
-
-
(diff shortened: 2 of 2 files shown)


DzikuVx:
I think we have to thing about lowering the number of element on OSD tab. There are just too many settings and page takes very long to load. Maybe if we spliy OSD into 2 separate ones? Like "OSD Layout" and "OSD Settings" ?
What do you think about that?

mmosca:
> I think we have to thing about lowering the number of element on OSD tab. There are just too many settings and page takes very long to load. Maybe if we spliy OSD into 2 separate ones? Like "OSD Layout" and "OSD Settings" ? What do you think about that?

Would the idea be to move the items on the right side to its own tab?

MrD-RC:
What about if the sections were collapsable?  That could look neater.

mmosca:
Just collapsing the sections probably wouldn't make it load faster.
 I am leaning on merging this one and addressing the split/load time on a subsequent pull request.

Something like this.

image

We can probably leave the video format and settings on the osd layout tab

And move alarms and the rest to the other tab, and possibly rename it to Advanced OSD settings, or something along those lines.

MrD-RC:
I guess so long as it doesn’t change anything on the layout. It could go in a different tab. But, that would only be about 10 things. I don’t think that would increase the load time by a great deal.

mmosca:
I will try to get it working, to at least check if it makes a difference. []s,Marcelo Bezerra ***@***.***>On 4 Aug 2023, at 23:41, Darren Lines ***@***.***> wrote:﻿
I guess so long as it doesn’t change anything on the layout. It could go in a different tab. But, that would only be about 10 things. I don’t think that would increase the load time by a great deal.

mmosca:
I will this one for now, while we investigate how to make the osd tab faster

Jetrell:
Its interesting to notice the OSD page load time, based on the FC processor clock speed I assume... The H743 or F765 generally load within 10s.. While the F722 and then F405 is a little slower. But the F411 takes 35s or so... It teaches me patience.

mmosca:
One thing I would like to do is investigate the loading process of the osd tab.

We may be able to cut the cpu usage on the fc and number of calls (and time to load) if we can:
1) Get list of visible elements
2) Only query the state of visible elements (hopefully, in bulk, async to fool the user the loading is fast)

PR #1819, opened 2023-09-01
INAV Configurator with Helicopter support

Woga65:
Intended for use with Heli-Quads, Chinook style Twin-Rotor-Helis and regular Helicopters.

Corresponding pull reguest for the firmware: [iNavFlight/inav PR 9271](https://github.com/iNavFlight/inav/pull/9271)

*_VP - target files are variable pitch while regular target files should behave like regular INAV. The *_VP - target files can also handle regular aircraft, albeit they are intended for use with Collective-Pitch aircraft. The *_VP - targets offer extended RC-Channel remapping to account for the COLLECTIVE + GYRO_GAIN RC-channels. 

**Latest changes**

- added configurable PID scaling (CLI)
- added timer usage flags for helicopter (TIM_USE_HC_MOTOR + TIM_USE_HC_SERVO)
- prevent the use of turtle mode on helicopter-like aircraft
- ensure throttle is cut on failsafe (helicopter) this might change when RTH + AUTO-LAND are sufficiently tested
- separated helicopter altitude controller from multirotor altitude controller
- improvement of RC-channel remapping from the CLI
- improved the handling of variable-pitch targets in the Configurator
- NAV-ALTHOLD now confirmed working on (at least the developer's) Collective-Pitch aircraft
- prevent I-term from winding up on the ground while spinning up the blades
- reworked on-the-ground detection to ensure stability in the air + calmness on the ground
- added separate hover settings for NORMAL + IDLE-UP 1 + IDLE-UP 2 flight modes (CLI)
- added configurable soft spool-up feature (CLI)
- show the RPM per motor in the configurator's Outputs tab
- added support for different NAV-ALTHOLD modes (CLI)

**Notes**
- to make use of this INAV Configurator, the matching firmware (*_VP targets) is needed (pull request in progress)
- mixer presets for regular helicopters in progress / coming soon
- 1500us on the collective-pitch rc-channel should result in a collective-pitch of 0 degrees
- when using BLheli_s, BLheli_32 or similar ESC, deactivate damped mode / braking and configure soft spool-up in the INAV CLI  
- no pirouette compensation at this point
- governor in progress / coming soon
- D-gains should be very low:  1 or 2 (one or two) maximum. To mitigate this, there is a group of CLI variables named `pid_scaling_factor_...` which allow for PID scaling between 0 and 1 or 2. For the D-gain it is advised to set the scaling factors (for roll/pitch/yaw) to values between 0.1 and 0.05 so you have a finer grained resolution and can use Derivative values between 1 and 15 or 1 and 30 in the PID tuning tab of the configurator 
- the P and D gains should be lowered with rising head-speed, this is what the Gyro-Gain RC-channel is for 
- the Gyro-Gain RC-channel works kind of reversed: 1499 - 0 elevate gains | 1500 gains untouched | 1501 - 2000 lower the gains (more head-speed => higher value of the Gyro-gain RC-channel)
- the Gyro-Gain and Collective-Pitch RC-channels replace AUX3 and AUX4. 
- CH7/AUX3 channel in the Modes Tab is internally forwarded to the throttle channel, which allows for switching between `nav_hc_hover_collective_normal` + `nav_hc_hover_collective_idleup_1` + `nav_hc_hover_collective_idleup_2` with the throttle (= flight-mode) switch. The values stored in the `nav_hc_hover_collective_xxx` CLI variables are used as a starting point for NAV-ALTHOLD.
- the soft spool-up time can be configured in the CLI: `hc_rotor_spoolup_time`  (10 = default, 0 = soft spool-up disabled). Active on the ground, disabled in the air to allow for bailing out of an auto rotation.
- the behaviour of the collective stick during NAV-ALTHOLD can be configured in the CLI: `nav_hc_althold_collective_type` (STICK | MID_STICK | HOVER).


tabs/firmware_flasher.html
```diff
@@ -55,7 +55,8 @@
                         <td><span class="description" i18n="firmwareFlasherManualBaudDescription"></span></td>
                     </tr>
                     <tr class="option noboarder">
-                        <td><label> <input class="show_development_releases toggle" type="checkbox" /> <span
+                        <!-- woga65: default to show development releases -->
+                        <td><label> <input class="show_development_releases toggle" type="checkbox" checked/> <span
```
MrD-RC, line 59: I'm assuming that this is for your testing only. Development releasees should **not** be shown by default.

tabs/firmware_flasher.js
```diff
@@ -170,7 +170,8 @@ TABS.firmware_flasher.initialize = function (callback) {
             TABS.firmware_flasher.releases = releases;
             return;
         };
-
+        // woga65: load firmware (pre)releases from variable pitch repository
+        //$.get('https://api.github.com/repos/Woga65/inav/releases?per_page=10', function (releasesData){
```
MrD-RC, line 174: Likewise, I assume this is for testing. The official iNavFlight repos should be used in release trim.

tabs/firmware_flasher.js
```diff
@@ -292,7 +293,8 @@ TABS.firmware_flasher.initialize = function (callback) {
 
                         $('a.flash_firmware').removeClass('disabled');
 
-                        if (summary.commit) {
+                        if (summary.commit) {   //woga65:
+                            //$.get('https://api.github.com/repos/Woga65/inav/commits/' + summary.commit, function (data) {
```
MrD-RC, line 297: Likewise, I assume this is for testing. The official iNavFlight repos should be used in release trim.

tabs/receiver.html
```diff
@@ -20,11 +20,11 @@
...
                 <div class="hybrid_element">
                     <input type="text" name="rcmap" spellcheck="false" /> 
-                    <select class="hybrid_helper"
-                        name="rcmap_helper">
-                        <option value="AETR">Default</option>
-                        <option value="AETR">Futaba / Hitec</option>
-                        <option value="TAER">JR / Spektrum / Graupner</option>
+                    <select class="hybrid_helper" name="rcmap_helper" id="rcmap_element">
+                        <!-- woga65: also map the AUX channels (will be overridden if the firmware sends only 4 channel's mapping) -->
+                        <option value="AETR1234">Default</option>
```
MrD-RC, line 25: If we are having to also have additional channels in the RC mapping. Please don't start an AUX section at 1. Just continue with the channel numbers.

BetaFlight uses the AUX1, AUX2... system and it is confusing, especially when new. Using AETR5678 makes more sense. People know what channel they have mapped their functions to on the radio. There's no need for them to have to remap their radio mixer in their head.

tabs/receiver.js
```diff
@@ -76,6 +76,28 @@ TABS.receiver.initialize = function (callback) {
...
+            RC_MAP.length === 8 ||
+            TARGET.isVariablePitch  // woga65: RC_MAP.length might be 4 if target is SITL, so make sure, heli mode is viable in simulator
+        ) {
+            rcMapElement.innerHTML = MIXER_CONFIG.platformType == PLATFORM_HELICOPTER
+                ? ` <option value="AECR12TG">Default</option>
+                    <option value="CAER12TG">Collective Pitch / Spektrum</option>
+                    <option value="AECR12TG">Collective Pitch / Futaba</option>`.trim()
+                : ` <option value="AETR1234">Default</option>
+                    <option value="TAER1234">JR / Spektrum / Graupner</option>
+                    <option value="AETR1234">Futaba / Hitec</option>`.trim();
```
MrD-RC, line 92: Please see my notes about **aux channels** from `receiver.html`.

I was going to suggest that **AECRTG78** would be better. But, I guess channel 5 has to be avoided due to _limitations_ with ELRS. So **AECR56TG** makes sense because of that.

tabs/receiver.js
```diff
@@ -120,21 +142,24 @@ TABS.receiver.initialize = function (callback) {
...
 
         // generate bars
+        // woga65: including AUX- and/or variable pitch specific channel names
         var bar_names = [
                 chrome.i18n.getMessage('controlAxisRoll'),
                 chrome.i18n.getMessage('controlAxisPitch'),
                 chrome.i18n.getMessage('controlAxisYaw'),
-                chrome.i18n.getMessage('controlAxisThrottle')
+                chrome.i18n.getMessage('controlAxisThrottle'),
+                'Aux 1 [1]',
```
MrD-RC, line 151: Please just use channel numbers, not **aux** numbers. The 4 channels after AETR should be: 
CH 5
CH 6
CH 7 or Collective [7]
CH 8 or Gyro gain [8]

It needs to be consistent with the current release.
image

MrD-RC (changes requested):

tabs/receiver.html
Woga65: Remember, "AETR5678" or "AETR1234" are channel names, not numbers. The channel number equals to the position of the letter from the left starting with one. 
So the wording for "TAER5687" would be: Throttle, Aileron, Elevator, Rudder, Channel five, Channel six, the channel that was previously channel seven, the channel that was previously channel eight.
(Just to explain the way I am thinking and why 5678 is more confusing to me. The requested change is on the way.)

tabs/firmware_flasher.html
Woga65: Yes, for testing. I will clean this up as soon as all probebly needed changes are discussed and applied.

tabs/firmware_flasher.js
Woga65: Yes, for testing. I will clean this up as soon as all probebly needed changes are discussed and applied.

tabs/firmware_flasher.js
Woga65: Yes, for testing. I will clean this up as soon as all probebly needed changes are discussed and applied.

PR #1820, opened 2023-09-02
Smarter pwm timer allocation, configurator changes

mmosca:
This is the configurator side changes for iNavFlight/inav#9268

Timers are clearly identified on the output mapping table.
Individual timers can be set to AUTO, SERVOS or MOTORS.

Output mapping may be wrong until after save/reboot.

image

TODO:
* [x] Load and save timer overrides via msp


mmosca:
Survived my basic abuse, more eyes on it wouldn't hurt.

PR #1829, opened 2023-09-25
mixer profile support

shota3527:
minimum mixer profile support:
- moved motorStopOnLow from feature to pid profile related settings
- added Mixer Transition servo mixing
- added Active MixerProfile and MixerTransition to OperandTypes
- allow speed setting for MAX smix

known issues:
- need to modify presets according to the motorStopOnLow changes
- save and reboot is broke in mixer tab, confirmed in SITL, not validated on real board

### DIFF
--- a/js/defaults_dialog.js
+++ b/js/defaults_dialog.js
@@ -822,13 +822,11 @@ helper.defaultsDialog = (function () {
                 key: "nav_fw_launch_climb_angle",
                 value: 25
             },
-        ],
-        "features": [
             {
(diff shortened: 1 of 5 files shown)


Jetrell:
@shota3527  I've noticed on a fixedwing platform, when using these changes... Causes **Stop motors on low throttle**  in the Outputs tab, to be turn OFF.  Every time _Save/Reboot_ is entered in the Mixer tab..
Is this by design ?

shota3527:
@Jetrell 
Thanks for reporting, that is not by design

PR #1833, opened 2023-09-29
Fix timer naming so they are indexed from 1 and aligned with firmware

DzikuVx:
image


mmosca:
I didn't call it TIMn on purpose, to detach it from the definition in the source code.
AT32 uses TMR instead of TIM, and the command line is 0 based.

So using a 0 based numbering and not calling it TIM, gives you a direct relationship to the command line, which the user is more like to interface with than the source.

mmosca (approved):

PR #1847, opened 2023-10-09
Handle monitor changes by not making the window larger than the screen

sensei-hacker:
@mmosca  wrote:

> I often have issues with the configurator opening out of range of my desktop, specially when connecting over remote desktop to my multi monitor desktop
> I know the configurator saves its position to launch at the same place, but is it possible to check if the save x and y is actually inside the display area of the current monitor setup?

This code does as requested - ensure the coordinates actually exist on the screen before trying to place the window there.


mmosca:
I will test is soon.

mmosca:
Initial test. 

My dual monitor setup:
Primary screen is on the right (screen 2)
image


* [x] Position saved and restored correctly on primary screen
* [ ] Position saved and restored on secondary screen: Not working. Currently opens on primary screen. I suspect we may be dealing with negative window coordinates.
* [x] Position save on secondary screen, opening on remote desktop single screen

So far, in my test case, windows always open on primary monitor. I can certainly live with this, but users may want it to open on save position on monitor 2.

sensei-hacker:
Yeah the API doesn't let us decide which monitor. It's blind to concept of multiple monitors.

What we the PR DOES do is avoid opening it larger than the monitor it's on. Where the secondary monitor is larger, prior to the PR it could open on the primary using size and position it saved from the secondary. The PR ensures the size and position are sane for the desktop it is on.

PR #1848, opened 2023-10-12
fix mixer_profie configurator issue

shota3527:
Requires firmware
https://github.com/iNavFlight/inav/pull/9364

- Fix the issue of the configurator cant not retrieving inactivate mixer_profile`s output mappings, used motor/servos
- Allow the configurator to set motor`s transition mixing by throttle range of -2~-1
- warning will appear when throttle value is out of 0~1 range


shota3527:
image

PR #1850, opened 2023-10-20
Dzikuvx ez tune

DzikuVx:
Zrzut ekranu 2023-10-20 122419

This side of https://github.com/iNavFlight/inav/pull/9354

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -5549,5 +5549,8 @@
     },
     "ezTuneRatePreviewExpo" : {
         "message": "Expo"
+    },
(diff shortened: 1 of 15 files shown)


trailx:
Hey Pawel, from a usability standpoint, it may make sense to group the descriptions with the sliders a little better.  Once you get to the middle of the sliders, around Response and Damping in particular, its a little unclear whether the text description goes with the slider above or below it.  I think what makes it more confusing is that the "title" is below the descriptions.  Typically, titles go above text.  You may want to add some sort of additional separation to assist with this if you want to keep this layout.

PR #1852, opened 2023-10-24
Serialpassthrough button in port tab

error414:
Story: Main goal was that writing whole comand into cli was slow and mainly because I do not remeber whole command. I use serialpassthrough once per month so would be more comfortably to just click to button. Serialpassthrough is not used as often so I decided to hidden buttons into clickable col. 

image

How it works: if you click to >>" col then area with buttons is opened. Buttons is generated only in row where is set some functionality (MSP, GPS, etc,,,, ), it's not showed for RX or USB VCP. Params for  serialpassthrough is taken from settings. Serialpassthrough params can be visualised by toats message from "question mark".

image

If baud rate is not possible determinate then there is select for baudrate selecting. 
image

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -4202,6 +4202,9 @@
     "portsPeripherals": {
         "message": "Peripherals"
     },
+    "serialpassthrough": {
+        "message": "Activate serialpassthrough"
+    },
     "appUpdateNotificationHeader": {
         "message": "New Configurator version available."
     },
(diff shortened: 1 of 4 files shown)


mmosca:
I think this would fit better in the CLI tab than in the ports tab.

For gps, there is a dedicated command: `gpspassthrough`, for others, is there really a common enough use case to add it to the GUI?

error414:
If you would like to have it in CLI tab then maybe would be better improve CLI autocomplete for better hinting for commands params, or at least add there description from "help" commad, I'm not sure now if there is params exmplanation. 

So can close this PR and I will findout how to improve CLI autocomplete.

PR #1877, opened 2023-10-31
preset on pid_iterm_limit_percent for MC

shota3527:
Set pid_iterm_limit_percent=50 for multi-rotor preset.
Default 33 might be better but 50 should be safer for unbalanced CoG/very windy conditions


mmosca:
If you want to include this in 7.0, you need to target the release_7.0.0 branch

Issue #1893, opened 2023-11-15
Bing Map Provider has massive altitude discrepancies in Elevation Data 


b14ckyy:
OSM Maplink Elevation Data differ massively from Bing Maps Elevation Data in configurator. 
OSM seems to be the correct one as it is on par with other providers like GMaps, Airmap and Copernicus within a 5m range or less. 
Depending on the region I checked Bing Maps shows +40-50m in Germany. up to +40m in UK. -40m in the US East coast and +20m on Hawaii. 
These errors are too big to be ignored and are not suitable for a ASL planned mission or automatic landing at all. 

Other Applications like MultiWiiPlanner also use Bing maps as a map provider and they show the correct data so I assume this is a parsing error in INAV Configurator. 

![image](https://github.com/iNavFlight/inav-configurator/assets/33039058/987969de-47ee-4c56-8356-063610e328da)
![image](https://github.com/iNavFlight/inav-configurator/assets/33039058/2ab81510-0d9a-4edb-9566-3b62925f8a08)

stronnag:
The problem it would appear is that the configurator is either using OpenTopo elevations for OSM maps and Bing  elevations with ellipsoid datum for Bing Imagery (or it's using Bing elevations with sealevel datum  for OSM and Bing elevations with ellipsoid datum for imagery.

For each of your locations (so the point moved ... but it makes no difference; the datum is the issue.
For example:

```
$  ./get_elevations -sea-level -bing-key $(<bkey.txt) 51.4499118 11.8806571
Using mapzen topo set
topo : 51.449912, 11.880657, elev = 119.0 in 346.475603ms
bing (sea-level) : 51.449912, 11.880657, elev = 118.0 in 201.815501ms

$ ./get_elevations -bing-key $(<bkey.txt) 51.4499118 11.8806571
Using mapzen topo set
topo : 51.449912, 11.880657, elev = 119.0 in 127.887764ms
bing (ellipsoid) : 51.449912, 11.880657, elev = 163.0 in 78.228758ms
```

and 

```
./get_elevations -bing-key $(<bkey.txt) 51.4498917 11.8805605
Using mapzen topo set
topo : 51.449892, 11.880560, elev = 119.0 in 122.580514ms
bing (ellipsoid) : 51.449892, 11.880560, elev = 163.0 in 239.017463ms


[jrh@eeyore]$ ./get_elevations -sea-level -bing-key $(<bkey.txt) 51.4498917 11.8805605
[...]

PR #1898, opened 2023-11-20
Change default Bing elevation datum to sealevel

breadoven:
Closes https://github.com/iNavFlight/inav-configurator/issues/1893.

Changes default Bing maps elevation datum to Sea Level with Ellipsoid now an optional setting.


MrD-RC:
Maybe the default could be selected in the gear settings. Next to where you’d set up using Bing! for the maps.

PR #1905, opened 2023-11-26
Sh profile fix

### LINKED ISSUE #1900: 7 RC2 Pid profile Issue
On the Pid tuning page and on the pid gains tab, when i switch pid profiles it changes the first time but if i try and change a second time nothing happens, the logs show the profile is changing but the values in the pid gains tabs only change the first time.

shota3527:
fixes #1900
played for about 2 mintues, no issues found


rts18:
@DzikuVx Can we have this fix merged before the 7.0.0 full release. So we can test it?

PR #1910, opened 2023-12-03
Show motor wizard results in mixer preview

sensei-hacker:
The mixer preview  has numbers showing which motor is where. Users expect that when they use the motor wizard to set the positions of the motors, the image should will reflect their changes.  A common support issue is "I used the wizard to set motor 1 to right front, but it didn't change" - because the preview image still shows it as right rear.

This commit updates the preview image with the positions set by the wizard.
Demo here:
https://youtu.be/Jsm5t_u5A0Y


mmosca:
Does it still show the new number positions when you disconnect and connect again?

sensei-hacker:
> Does it still show the new number positions when you disconnect and connect again?

Yes. It reads the existing mixer lines and shows those, unless / until you use the wizard to change them.

Issue #1916, opened 2023-12-07
wrong units in descriptions in "Advanced Tuning"


tippfehlr:
In Advanced Tuning the units in the descriptions don’t always watch the units in the input field.
In Failsafe the deciseconds notice is still there, but it doesn’t claim the value is in that unit.

For example:

![image](https://github.com/iNavFlight/inav-configurator/assets/46620049/6e12e49f-f91c-41ad-a194-063607477ab3)
![image](https://github.com/iNavFlight/inav-configurator/assets/46620049/801e895e-7d7c-4d82-8f74-9252f605b917)
![image](https://github.com/iNavFlight/inav-configurator/assets/46620049/d1cfc809-94ac-4c49-9135-2e3ab1f3d247)
![image](https://github.com/iNavFlight/inav-configurator/assets/46620049/56175b3e-68af-4ba0-b538-f79436391ea6)
![image](https://github.com/iNavFlight/inav-configurator/assets/46620049/58db422a-c42c-4229-a08d-1a809e80e632)
etc.

MrD-RC:
The units shown in the description are the default units. This is expected display.

PR #1917, opened 2023-12-07
fix(messages): correct units in Advanced Tuning and in Failsafe

tippfehlr:
Issue: #1916


MrD-RC:
This is not a valid fix. This is only changing the default units to metric. In Configurator, you can select other unit types, including Imperial and GA. In those cases, and the default units, these descriptions would be wrong. 

This is why the description were left with the default units. As they are true to CLI.

Issue #1921, opened 2023-12-11
7.0 final, unable to initiate a freshly flashed FC


vevere:
Tested on:
- SpeedyBee F405 Wing
- AtomRC F405_NAVI

Step1: flash inav 7.0.0 with Full chip erase (no problem)
Step 2 : after reboot, select "Airplane with a tail" or "Airplane with out a tail" 

then the step of applying the defaults settings never ends:
![image](https://github.com/iNavFlight/inav-configurator/assets/26826853/e4c8e57c-bcf8-4c5b-96ff-f08ca329c5f5)

I can cancel this task with the top white cross but I'm not able to know is any default parameter is missing or not

I can reproduce the problem without flashing, juste hit "reset settings" on the Setup page and the problem is the same

ChristopherBrusie:
I can confirm this same issue with the SoeedyBee F7 v2 target.

eLEcTRiCZiTy:
This is similar to new issue described here: [inav-configurator/issues/2090](https://github.com/iNavFlight/inav-configurator/issues/2090)

Issue #1896, opened 2023-11-20
7 Final release gives issues


0crap:
~~- Battery profiles do not update when switching~~
~~- Settings validated checkmark stays red (even when known good diff all is applied)~~
- After full erase and flash it hangs after apply defaults for wing **- Still a bug in final release. -**

Tested with ~~INAV-Configurator_win64_7.0.0-RC2.zip~~ INAV-Configurator_win64_7.0.0.exe and Matek [H743-Wing V2](http://www.mateksys.com/?portfolio=h743-wing-v2)
None of these issues with Configurator 7 RC1.

RoadyFPV:
same here it shows reboot ez tune tab is there. after u close configurator and connect again ez tab is not there. and settings validated is red. cli status show ez_filter_hz error

RoadyFPV:
same in release version

0crap:
> same in release version
> 
> stuck in: Device - Saving default settings

Confirmed here as well.
That is just sloppy.
No reason not to do a RC3 for the Configurator to see if this was fixed.
Now it's still there in the final release. Nice...

Yury-MonZon:
Same here on linux x64 7.0 version Matek F411WTE

RoadyFPV:
its not a cosmetic error, after u click on x and start configurator again ez tab is missing

RoadyFPV:
thanks for the clarification. in the rc version, ez tab was also displayed for airplane. i will check it again.
and the boot stuck on savig defaults?

0crap:
>.... and the boot stuck on saving defaults?

This is still a bug as mentioned in the opening post.

PR #1922, opened 2023-12-11
Fix Defaults Dialog

Scavanger:
Fixes some issues with the "Default Dialog":

#1896 #1921 

- All default settings are now saved,  it should no longer be stuck in the dialog "Device - Saving default value"
- Removed old unused default value ```failsafe_mission```
- Added matching ```ez_filter_hz``` for Multirotor
- If a setting is not available (e.g. with SITL), the remaining settings are saved anyway.
- Reboot after "Reset Settings"

### DIFF
--- a/js/defaults_dialog.js
+++ b/js/defaults_dialog.js
@@ -192,6 +192,11 @@ helper.defaultsDialog = (function () {
             {
                 key: "failsafe_procedure",
                 value: "DROP"
+            },
+            // Ez Tune
+            {
+                key: "ez_filter_hz",
+                value: 90
             }
         ]
     },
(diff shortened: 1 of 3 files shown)


MrD-RC:
This seems to have broken setting the default settings in the 3 control and possibly the 3 battery profiles.

https://www.facebook.com/groups/INAVFixedWing/posts/2351057105096074/?__cft__[0]=AZVtVMe2XzgZX0AC7sM7pdRM5OOTIOC-vs2Iba2PjDIofjtt0JFBWSRi0Ng8IPbB0o6OMxHNoaymqZkWN1so9UUju7r2whtxEUhaFPfFdfy3qQCSmNKxIOW5WABpWYNnSjY5UbSSYXWzxO6XaHVQMmzo&__tn__=%2CO%2CP-R

PR #1930, opened 2023-12-25
Add github actions for building configurator

mmosca:
Adds builds to prs that generate installers/binaries for linux, windows and mac.

The mac binaries are not notarized.

The windows installer is not signed.

I got some random linux build failures, still investigating those.


DzikuVx:
replaced by #2039

PR #1960, opened 2024-02-10
Fixed Wing Autoland

Scavanger:
Configurator support for: https://github.com/iNavFlight/inav/pull/9713

Includes a overhaul of safehomes in mission control.


stronnag:
Please use lower case for mission files tags and attributes, as is the [INAV convention](https://github.com/iNavFlight/inav/tree/master/docs/development/wp_mission_schema). These is also a [XSD](https://github.com/iNavFlight/inav/blob/master/docs/development/wp_mission_schema/mw-mission.xsd) which supports this.

In order for the boolean values to validate as `xs:boolean` (rather than arbitrary text), the values should also be lower case.

breadoven:
Settings for `FW Approach Alt (m)` and `FW Land Alt (m)` in the settings box don't save. Seems to be due to `vMaxDistSH ` being undefined at https://github.com/iNavFlight/inav-configurator/blob/841b05ccd769aa48cda3d6c311492afc9fa3001e/tabs/mission_control.js#L3122C186-L3122C196.

Also the landing approach path doesn't display on the map seemingly because `settings.fwApproachLength` becomes undefined 
 (same for `settings.fwLoiterRadius`). It's initially correctly set at https://github.com/iNavFlight/inav-configurator/blob/841b05ccd769aa48cda3d6c311492afc9fa3001e/tabs/mission_control.js#L100C21-L100C46 but by the time the map is draw it's become undefined. I can't work out what's causing this.

Not sure why these are defined as Configurator settings when they are FC specific settings surely ?

Has anyone else noticed these problems or is this something specific to my Configurator build ?

breadoven:
Just tried the official 7.1.0 RC1 Configurator release and the landing approach path does display correctly so no idea what's wrong with my custom build, not immediately obvious anything is overwriting the settings that become undefined.

However, the settings box/panel still won't save changed settings no doubt for the reason given above.

PR #1973, opened 2024-02-20
Add platform specific GPS/Baro weight in defaults

MrD-RC:
Specific settings for `inav_w_z_baro_p`, `inav_w_z_gps_p`, and `inav_w_z_gps_v` for multirotors and fixed wing.


mmosca:
I am not convinced we need the profile changes. We reverted to better behaving defaults and in 8.0, with the baro filtering changes and potential gps weight adjustments by hdop, the defaults will be good enough. If we still think users will be changing these often, they should be somewhere in the configurator instead.

DzikuVx:
I think users should not be changing those. It's complex already and let's not make it more complex tha  before.

MrD-RC:
It seems like this is not necessary. So closing the PR :)

PR #1974, opened 2024-02-20
fix OSD bug, no HD variants in select

error414:
from facebook, channel **INAV Fixed Wing Test-Chamber**

Rodrigo:
Anyone else not being able to select the HD OSD (Avatar, HDZero, DJIWTF, BFCOMPAT) on the 7.1.0 configurator?


MrD-RC:
Why only load the elements if 7.1? This needs to be backwards compatible with all 7.x releases. I’m also not experiencing this issue.

MrD-RC (approved):

PR #1984, opened 2024-03-05
[7.1 RC1] Fix saving mission file

Scavanger:


breadoven:
OK so the file now saves ... but it won't reload. Seems `parseBooleans `isn't recognised in `loadMissionFile`.

Scavanger:
Args! 
Sorry, the "requier" somehow got lost.

PR #1990, opened 2024-03-07
Make it impossible to modify USB VCP serial port assignment in the GUI

DzikuVx:
image

### DIFF
--- a/tabs/ports.js
+++ b/tabs/ports.js
@@ -230,66 +230,77 @@ TABS.ports.initialize = function (callback) {
 
             port_configuration_e.data('serialPort', serialPort);
 
-            port_configuration_e.find('select.msp_baudrate').val(serialPort.msp_baudrate);
-            port_configuration_e.find('select.telemetry_baudrate').val(serialPort.telemetry_baudrate);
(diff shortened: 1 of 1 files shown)


stronnag:
Perfect, all the other useful cases require premeditated acts in the CLI (and still need MSP). And if the user is obtuse enough to disable VCP MSP via the CLI, they deserve the consequences.

stronnag:
But why even show VCP?

DzikuVx:
Hmm... good question @stronnag maybe this is the way indeed

YI-BOYANG:
image
Haha, this is really a good PR, I also accidentally turned off the USB VCP once, which made me no longer able to connect to the configurator and had to re-burn the firmware.
At that time I thought, why can't you forbid the user to operate the switch of USB VCP like BF configurator? So I refer to the code of the BF configurator and implement this function on the INAV configurator.
In the end, I didn't submit the PR because I wasn't sure if the INAV configurator left the ability for the player to operate the USB VCP, or what else.

DzikuVx:
@YI-BOYANG I think that @stronnag suggestion to remove VCP from this table completely will be a better idea. I will handle that as soon as I find some time

Issue #1982, opened 2024-03-04
7.1 RC1 Issues


breadoven:
2 minor issues with 7.1 RC1 on windows 10:

1) The usage option setting for the new CLI Auto Complete isn't saved on shutdown, it's always selected ON when restarting Configurator.
2) Configurator window won't remember previous maximised size when restarted.

PR #1992, opened 2024-03-09
Window sizing and CLI auto complete setting fixes

breadoven:
Closes https://github.com/iNavFlight/inav-configurator/issues/1982.

### DIFF
--- a/main.js
+++ b/main.js
@@ -135,8 +135,8 @@ $(document).ready(function () {
                 } else {
                     win.height = result.windowSize.height;
                     win.width = result.windowSize.width;
-                    win.x = result.windowSize.x;
-                    win.y = result.windowSize.y;
+                    win.x = Math.max(result.windowSize.x, window.screen.availLeft);
+                    win.y = Math.max(result.windowSize.y, window.screen.availTop);
                 }
             }
         });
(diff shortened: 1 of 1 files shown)


sensei-hacker:
Out of curiosity, what's the purpose of removing the x and y checks and drawing off the screen?

breadoven:
> Out of curiosity, what's the purpose of removing the x and y checks and drawing off the screen?

Freedom of choice surely ?

I just realised you changed this via https://github.com/iNavFlight/inav-configurator/pull/1847. Unfortunately I can't see how that change works if the saved screen size is bigger than the available screen size the system reports ... it just ends up stuck with the default screen size which is what happens in my case. The saved size is a few pixels bigger than the supposed available size in my case.

I changed this so it will now always align at the top left corner so it can't start off screen (except it doesn't on my screen because there are inconsistencies between the screen size/position data reported ... so you end up with a gap). Not sure if this change works for the original reason for #1847 or not @mmosca ?

mmosca:
My issue was that when remote desktoping to my desktop computer that has 2 screens, sometimes the saved screen positions would be on the secondary display and I was not able to open configurator, as it was open off screen.

So I would often have to nuke the saved config to restore configurator to the first screen, or remember to make sure configurator was always on the first screen.

mmosca:
Remote desktop is fine. It actually always shows up in the primary display, which I think is ok.

PR #2011, opened 2024-04-01
Default Dialog Profile Fix

Scavanger:
Deafults dialogue now sets all profiles correctly.


MrD-RC:
I think there may be a 7.1.1. So it may be worth waiting for that. Or at very least this PR should be cherry picked for it.

PR #2025, opened 2024-04-10
fixed aligment tool

RomanLut:
Fixes https://github.com/iNavFlight/inav-configurator/issues/1955

### DIFF
--- a/tabs/magnetometer.js
+++ b/tabs/magnetometer.js
@@ -600,7 +600,7 @@ TABS.magnetometer.initialize3D = function () {
         var magRotation = new THREE.Euler(-THREE.Math.degToRad(self.alignmentConfig.pitch-180), THREE.Math.degToRad(-180 - self.alignmentConfig.yaw), THREE.Math.degToRad(self.alignmentConfig.roll), 'YXZ'); 
         var matrix = (new THREE.Matrix4()).makeRotationFromEuler(magRotation);
 
-        var boardRotation = new THREE.Euler( THREE.Math.degToRad( -self.boardAlignmentConfig.pitch ), THREE.Math.degToRad( -self.boardAlignmentConfig.yaw ), THREE.Math.degToRad( -self.boardAlignmentConfig.roll ), 'YXZ');
+        var boardRotation = new THREE.Euler( THREE.Math.degToRad( self.boardAlignmentConfig.pitch), THREE.Math.degToRad( -self.boardAlignmentConfig.yaw ), THREE.Math.degToRad( self.boardAlignmentConfig.roll ), 'YXZ');
         var matrix1 = (new THREE.Matrix4()).makeRotationFromEuler(boardRotation);
 
 /*
--- a/tabs/magnetometer.js
+++ b/tabs/magnetometer.js
@@ -198,7 +198,7 @@ TABS.magnetometer.initialize = function (callback) {
         var magRotation = new THREE.Euler(-THREE.Math.degToRad(degree[0]-180), THREE.Math.degToRad(-180 - degree[2]), THREE.Math.degToRad(degree[1]), 'YXZ'); 
         var matrix = (new THREE.Matrix4()).makeRotationFromEuler(magRotation);
 
-        var boardRotation = new THREE.Euler( THREE.Math.degToRad( -self.boardAlignmentConfig.pitch ), THREE.Math.degToRad( -self.boardAlignmentConfig.yaw ), THREE.Math.degToRad( -self.boardAlignmentConfig.roll ), 'YXZ');
+        var boardRotation = new THREE.Euler( THREE.Math.degToRad( self.boardAlignmentConfig.pitch ), THREE.Math.degToRad( -self.boardAlignmentConfig.yaw ), THREE.Math.degToRad( self.boardAlignmentConfig.roll ), 'YXZ');
         var matrix1 = (new THREE.Matrix4()).makeRotationFromEuler(boardRotation);
 
         matrix.premultiply(matrix1);  


Jetrell:
@RomanLut Thanks for your work on this.
I tested the changes, and have found the flight controller mounting orientation now works well at any position. Right way up or flipped. And any yaw offset.

Although the compass _orientation preset_ is only based around `CW270FLIP`. And altering that preset when the hardware requires it, does not work. 
Which means if the user happens to purchase a GNSS module with compass. That has a different orientation preset, based on the magnetometer chip, and its **Arrow** or **Plug faces the rear of the aircraft rear**. Then the arrow in the virtual image of the alignment tool is incorrect.. Which means the setup is not correct either.
And example of this is the Walksnail M181 GPS. Which has an Orientation preset of `CW0FLIP`.
I have one of these units mounted at Yaw 180° in a plane, with the plug facing forward... The alignment tool provides the wrong settings, to what is required for it to work.

I have written more information on the subject of Orientation Preset and GNSS units like the M181. To provide a more in depth explanation. https://github.com/iNavFlight/inav/wiki/GPS-and-Compass-setup#setting-up-the-compass-alignment

samsab1:
many thanks for your gps/magneto wiki
from your experience , can you tell me how to align the HGLRC M100-5883 M10 GPS where no arraw id displayed.
https://fr.aliexpress.com/item/1005006083972371.html

i have a big doubt about orientation, and i do not know how to clear that doubt.
best regards from france

Jetrell:
> many thanks for your gps/magneto wiki from your experience , can you tell me how to align the HGLRC M100-5883 M10 GPS where no arraw id displayed. 
> i have a big doubt about orientation.

It looks to be the same layout as the Walksnail M181.
Its orientation preset would then be `CW0FLIP`. With the arrow facing towards the front of the aircraft, along with the flight controller arrow facing the same way.

If it had an arrow. It would be like this.

gps

RomanLut:
image

Y and Y marking on this picture is incorrect.

Screenshot_20240427_131140_Chrome

There a is dot marking pin 1 of the chip.
Pins are numbered counter clockwise,

X axis is along pins 1,2,3,4
Y Axis is along chip making reading direction.

You also can see this on qmc5883 module:

image

Y axis on matek module should point down.
The arrow placed on Matek module points in the opposite direction to Y axis.


m100

RomanLut:
walksn

Jetrell:
I did make a mistake in the wiki.. But I added this image to help explain it better.

Clockwise arrows

> Y and Y marking on this picture is incorrect.

I agree, that image in not perfect. Although the main issue is caused by the manufacturer mounting the mag chip any direction they want. With respect to the plug being at the rear. And lately not even providing a reference arrow.

Because manufactures like Walksnail and HGLRC are not following the default orientation preset of CW270FLIP. We require that function to work independently of the PRY magnetometer alignment sliders.

Please don't take my statements as criticism. I very much respect your work and efforts towards this complex issue.
I spent ages trying work out how to get the M181 working with its FC last year. And it was confusing.. And after trying practically every mag alignment and orientation preset there is to get it working. This is how I know there is still an issue with the tool.

Jetrell:
@RomanLut  This may help you visualize the issue withe the M181.

These are the only settings I could find to make this plane work correctly. And its now had over 50 flights.
```
align_mag = CW180FLIP
align_mag_pitch = 1800
align_mag_yaw = 1800
align_board_yaw = 1800
```
This is the aircrafts layout. With the FC's arrow facing rearward. And the compass plug facing forward, based on the magnetometer chips orientation.  

mounting

And this is what the tool shows with this PR.. The FC arrow is correct, But the mag arrow is off by 90.. The plug should be facing forward (and the arrow backwards if this unit provided one).

alignment tool

RomanLut:
>Please don't take my statements as criticism. 

Everything is Ok, Testing ang bug reports are appreciated! I am interested in making it work correct, but I do not use theese combined GPS modules.

Now I understand that my assumption that arrow points to the opposite side of axis Y is correct for Matek m10q only.
Moreover two other modules listed above does not have arrow at all.
So basically we should add correct 3d model for each combined GPS module on the market.

This will be problematic. Instead of this I think I should add 3D models of dummy sensor PCBs like:
image
with dot and markings on magnethometer chip clearly visible, for each sensor supported by inav.

Users wil be able to find orientation by matching orientation of the chip on the PCB and the chip on physical setup.
I will do it next week.

sensei-hacker:
> So basically we should add correct 3d model for each combined GPS module on the market.

I have a working proof of concept which dispenses with all of this. It has the user point the aircraft roughly north, then roughly east. From those two readings we can *compute* the orientation.

Similarly for FC / accelerometer orientation. While the aircraft is roughly flat (and pointing north for compass, whichever direction reads about 1G is the side that's pointing down. 

It all works and makes it easy for the user. The one challenge remaining is translating the new mag orientation based on what the new FC orientation will be. A little bit of quaternion math that I'm not particularly comfortable with.

Jetrell:
@sensei-hacker It would make it easier in theory.. Provided the magnetic bearing is not influenced in the slightest by weak magnetic fields emanating from something on the craft.. Otherwise that magnetic deviation would render the derive compass orientation erroneous. 

The benefit of having an orientation preset that is based on physical parameters. Means its one less thing to guess about if the compass calibration is not correct. Or magnetic deviation does occur inflight from something on the copter. 
e.g. high bank angles, causing the mag to temporarily intersect the motor magnetic fields. Or higher current drain emitting a magnetic field disturbance.

sensei-hacker:
> the compass calibration is not correct. Or magnetic deviation does occur inflight from something on the copter.
e.g. high bank angles

You'd still set the compass orientation once, on the bench. Not continuously in flight. Nothing changes about in-flight behavior.

You simply have the user point the aircraft north, then east, and round to the nearest 90°.  (Unless both readings are a very good match for a 45° increment).

RomanLut:
Added magnetometer chips and some GNSS modules:

https://github.com/iNavFlight/inav-configurator/pull/2056

PR #2028, opened 2024-04-11
Copy/Paste of elements in OSD layout

error414:
Here is video to explain how to works: https://www.youtube.com/watch?v=-F0jJvEQK2w 

It's a simple clipboard for layouts, it's easy to use. Two buttons copy and paste. 

Copy: save curent layout to clipboard
Paste: apply layout from clipboard

Super simple.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -4847,6 +4847,9 @@
     "paste": {
         "message": "Paste"
     },
+    "clear": {
+        "message": "Clear"
+    },
     "active": {
         "message": "Active"
     },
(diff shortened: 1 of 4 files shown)


MrD-RC:
This is really cool. I think a **Clear OSD** option would be nice too.

error414:
it's a good idea, just set all "isVisible" to false and then it's variant of "paste" functionality. easy to do. I will add it.

error414:
Clear functionality added, Buttons "paste" and "clear" send changes to FC, but for make changes permanents is still necessary to click to save button. I would say it's ok. 

image

PR #2039, opened 2024-04-24
[Electron] GitHub Actions

Scavanger:
Builds Configurator on PR

Builds:
 - Linux: x64 zip, deb and rpm (Electron does't support 32 bit on Linux, see https://www.electronjs.org/blog/linux-32bit-support
  - MacOS x64 and arm64: zip and dmg
  - Windows x32 and x64: zip and msi installer


DzikuVx:
Awesome!

PR #2041, opened 2024-04-26
[Electron] Fixes and Adjustments

Scavanger:
- MacOS:
   - Nice icon for DMG
   - Fix bug when Configurator was minimized and dock icon is clicked

- Fix (Workaround):  After a MessageBox (alert()) was displayed, all input fields stopped working.
- Fix Blackbox tab (save data flash)
- Removed useless MSC button in blackbox tab
- Adjust Readme
- Add context menu  (copy and paste, undo, redo)
- Fix LED Tab
- Fix Github Actions (What still worked yesterday. )

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,7 +85,7 @@ jobs:
             cache: 'npm'
         # Workaround due to a bug in node-gyp: https://github.com/electron/rebuild/issues/1116
         - name: Install Setuptools
-          run: pip install setuptools
+          run: python3 -m pip install --break-system-packages setuptools
         - name: Install deps
           uses: nick-fields/retry@v2
           with:
--- a/README.md
+++ b/README.md
@@ -20,32 +20,32 @@ everything, the hardware is not working, or you have any other _support_ problem
 
 ## Installation
 
-Depending on the target operating system, _INAV Configurator_ is distributed as a  _standalone_ application or Chrome App.
+ _INAV Configurator_ is distributed as a  _standalone_ application.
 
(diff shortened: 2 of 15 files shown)


Jetrell:
Thank you for the fixes @Scavanger 

I found another issues that I'm not sure if you're aware of. 
When attempting to load a previously saved mission file from a js build.. Or even save a mission file that has been made with the electron build.
This log error occurred. `error reading file (xml2js not found)`

Scavanger:
> Thank you for the fixes @Scavanger
> 
> I found other issues that I'm not sure if you're aware of with the Windows build. When attempting to load a previously saved mission file from a js build.. Or even save a mission file that has been made with the electron build. This log error occurred. `error reading file (xml2js not found)`

Thanks for finding it. Fixed.

mmosca:
Not exactly related, but is it possible to use https://github.com/electron/universal to distribute a single universal binary application, instead of separate x64 and arm64?

Scavanger:
> Not exactly related, but is it possible to use https://github.com/electron/universal to distribute a single universal binary application, instead of separate x64 and arm64?

I have already tried this in the GitHub Actions, but unfortunately it always cancelled with the following error message:

```
✖ Stitching darwin/x64 and darwin/arm64 into a darwin/universal package [FAILED: While trying to merge mach-o files across your apps we found a mismatch, the number of mach-o files is not the same between the arm64 and x64 builds]
```

I'm not a Mac expert if you know how to fix the error, please feel free.

DzikuVx:
let's merge and have incremental improvements

PR #2042, opened 2024-04-26
Fix GitHub Actions for MacOS

Scavanger:
I don't know why, yesterday it still worked, today it doesn't.


Scavanger:
Now in #2041

PR #2043, opened 2024-04-27
Bing maps elevation dem datum fix

breadoven:
It's correct in 7.1.1 but fixing in Master just to avoid any more inadvertent datum swapping.


mmosca:
merging the latest master should fix the build checks. github upgraded the macos-latest image in the action to macos-14 and it broke some of the build. I set it back to macos-13 in the mean time.

PR #1849, opened 2023-10-12
SITL: adjusted SITL tab (built-in serial receiver support)

RomanLut:
Accompanying  submit for https://github.com/iNavFlight/inav/pull/9365

NOTE:  SITL executables in resources folder of inav configurator should be updated from inav repository after merging inav PR!

PR #2044, opened 2024-04-27
Enable SITL for MacOS

mmosca:
Since the sitl binary for MacOS now works, we can enable Demo mode for MacOS as well. :)

Still missing Ser2TCP, but should be fixed by iNavFlight/inav#9365 and #1849


Scavanger:
It's great that it also runs on Macs.

Shouldn't we also activate the SITL tab then?

/tabs/sitl.js: 81
```
 var os = GUI.operating_system;
    if (os != 'Windows' && os != 'Linux') {

        $('.content_wrapper').find('*').remove();
        $('.content_wrapper').append(`<h2>${i18n.getMessage('sitlOSNotSupported')}</h2>`);
        
        GUI.content_ready(callback);
        return;
    }
```

PR #2056, opened 2024-05-01
added magnetometer models

RomanLut:
This PR adds 3D models of various magnetometers and magnethometer chips to Alignment tool tab.

image


samsab1:
sounds good :-)

RomanLut:
>@Jetrell But I did notice the alignment tool would occasionally lock-up the configurator after a save/reboot.

Do you mean that after "save and reboot" configurator may fall into never endling loading cycle? Unfortunatelly it happens with other tabs with current Master. Alignment tool is not exception here.

RomanLut:
--------------------
>@Jetrell All he understands is making the image of the GNSS module in the tool, look the same as it does on his hardware. And hoping that will provide him with the settings he needs to make it work.

Yes he has to do this  way. This is the whole idea of the tool. If he can provide photos of his setup, we may try to understang what is wrong.

--------------------
The are two methods to provide orientation of the mag in inav:
a) with **align_mag** preset
b) with **align_max_pitch, align_max_roll, align_max_yaw** angles.

If all (b) are zero, then **align_mag** is used. if any of (b) is non-zero, then (b) is used and **align_mag** is ignored.

(a) defines rotation relative to FC. (b) defines absolute rotation. 

This is how settings work in inav. This is not going to be changed.

--------------------
image

On this picture tool shows that **align_mag** is ignored  - dropdown is crosed out. Orientation is set with (b) - defined with sliders.
Dropdown can have any value here, it is just a current value of **align_mag** setting from cli which is **ignored**.

--------------------
image

On this picture, tool shows that **align_mag=cw270flip is used** (dropdown is **NOT** crossed out).

At the same time, sliders show values for (b) which would provide the same orientation **as  align_mag=cw270flip with current FC rotation**. If user desides to define orientation using (b), he can touch sliders. Dropdown will became crossed out. This  very usefull if mag is mounted slightly off. User can rotate it a little bit at any time, tool will switch to (b).

--------------------
>The image below shows how the tool does not work as the user would expect.
>With slider align_mag_yaw = 0. The GNSS/mag arrow should be facing forward, as it would be on their aircraft hardware.

image

Arrow is not facing forward because **align_mag** is ignored.

I thought that crossed out dropdown would be enough to understand that preset is not used. If it is not the case - I can add explicit bold text somewhere. 

--------------------
If you want to define orientation with (a) - select preset in dropdown, do not touch sliders, save and reboot.
If you want to define orientation with (b) - setup orientation using sliders. Dropdown is crossed out once sliders are toched.

--------------------
You might wonder what will happend if user set sliders to zero and save setting - will it revert to cw270flip preset and screw up orientation? No, tool will save  align_max_pitch=0, align_max_roll=0, align_mag_yaw=**1**. Orientatino will still be defined by angles consitently.

image

Jetrell:
>The are two methods to provide orientation of the mag in inav:
a) with align_mag preset
b) with align_max_pitch, align_max_roll, align_max_yaw angles.

I agree. And I understand your view point on this matter.
However.. I'm speaking from the perspective of others who will not understand. Those who have never used a magnetometer before.. We often see this on Discord.

>(a) defines rotation relative to FC. (b) defines absolute rotation.

I believe the above explanation is somewhat confusing for new users.
In reality, both (a) and (b) are relative to the flight controller.
While (a) is **also** relative to the magnetometer chips mounting position on the PCB, with respect to a given reference point on the module. (the arrow or plug). Comparing to the front 0° of the aircraft.

I rang the guy I was speaking to about this. He doesn't have a Github account.
He gave me this example. Which did not make sense to him.

basic install

You can see that the front of the aircraft is the reference point. (green arrow and 0°.)
The arrow on the module is facing the rear as he requires.. Which is 180° from the front of the aircraft.
But the YAW slider has to be set to 90° instead of 180°, based on the modules rendered arrow facing rearward.. And to provide the correct settings.
Which all reference back to the Orientation Preset. Because its CW270° minus CW90° = CW180°.

I know this is simple if you understand, as you and I do.
But it makes the tool so confusing for new users from a visual perspective, when things don't line-up with what people see on their hardware.. Especially at the times the tool is required the most for more diverse hardware orientations.

RomanLut:
If FC is mounted with arrow pointing forward:

If matek-m10q is mounted with arrow facing forward, then cw270flip orientation preset should be used.
The same orientation can be set using align_mag_roll =0,  align_mag_pitch =180,  align_mag_yaw=270, as shown by the tool:

image

If matek-m10q is mounted with arrow facing backward, then cw90flip orientation preset should be used.
The same orientation can be set using align_mag_roll =0,  align_mag_pitch =180,  align_mag_yaw=90, as shown by the tool:

image

Sliders do not show rotation which is "additionally applied on top of orientation preset". They show values of align_mag_xxx settings which corresponds to rotation of magnethometer as seen on screen.

Maybe this helps to understand: yaw slider has to be set to 90 because CW90flip means "rotation by 90 on yaw".

rts18:
@RomanLut I've been watching this pull request along with #2025.
I was one of the people who pointed out the alignment tool had issues.

I would like to send a thank you for attempting to make this tool work correctly. 
However, with all due respect. After reading the comments in #2025 and here. And being a long term iNAV user. I have to agree with @Jetrell .
I too understand your perspective as he does. But I don't see how that will make this tool useful for new users.

> If matek-m10q is mounted with arrow facing backward, then cw90flip orientation preset should be used.

We may understand. But I guarantee you that know one else will. If I had to go to that much trouble to decipher what the tool is telling me, I wouldn't bother with it. As I don't. But this isn't about me.  
Newbies need someone with your skills to make this tool easy to interpret. Most do not have the ability to work out the spatial axis offset.
They require a tool which allows them to replicate their models hardware layout by moving the sliders to a setting that makes **sense**. And provides the correct compass and flight controller settings, based on what the GUI is showing them.

RomanLut:
User in NOT required to undestand how aling* settings work. User is NOT required to know preset name for default mounting of mag. User is NOT requred to know order or rotations and directions of euler angles. In extreme case, I could hide all text, numbers and dropdown, and display just three nameless sliders. All user is required to do - is to rotate mag model to match physical orientaton and click "save and reboot". Values of align_mag_* settings makes not sense for the user? It is expected. In order to settings to make sense, user has to understand everything mentioned above.

Jetrell:
@RomanLut 
The fact you have the tool working correctly. But its just the degrees numbering on the sliders and in the boxes that are not aligning with what the users sees on their hardware.

> In extreme case, I could hide all text, numbers and dropdown, and display just three nameless sliders

I think that idea would be best if the user has selected a combine compass model from the drop-down menu. 
That way they can just use the rendered image as a reference. Which is what most will do, if their compass model is available in the menu.

Capture

But having all the degrees numbering will be ESSENTIAL if a combine compass model is NOT loaded with its orientation preset, or a compass model is not available for there hardware.

mmosca:
I was just looking at the alignment tool in 7.x and noticed that the default model arrow does not match the gps axix, which certainly does not help. Having these many models with correct alignment would be a great step forward, but I worry a bit about keeping it up to date in the long term.

RomanLut:
@mmosca Y Axis shoud not point to the north, it should point in direction of magnethometer Y axis with current mounitng. In the long term, aligment is possible by alligning magnetomer chip orientation or magnetometer axes.
https://github.com/iNavFlight/inav-configurator/issues/2061

Jetrell:
@RomanLut  Do you have anything against changing this https://github.com/iNavFlight/inav-configurator/pull/2056#issuecomment-2094536297 ? ONLY when one of the compass preset models are selected.

And adding your version of this [chart](https://github.com/iNavFlight/inav-configurator/pull/2025#issuecomment-2080651934) for when there isn't a compass preset available in the drop-down. So the user can determine their modules orientation preset.. 

>Y Axis shoud not point to the north, it should point in direction of magnethometer Y axis with current mounitng. In the long term, aligment is possible by alligning magnetomer chip orientation or magnetometer axes.

This proves the XYZ render image is going to be confusing to most users, because it depends on the position of the mag chip on the specific hardware.

Other than that, your algorithm works fine from my testing.. Its just the interface that needs a little refinement for the users that don't think like an electrical engineer ;-)

Choolet:
I was going to report a problem with the compass side of the alignment tool in iNav 7.1. But its already being worked on. Thanks @RomanLut 
 I read through the comments. At the risk of sounding like another broken record, it leaves me wondering why you have not zeroed the magnetometer sliders once the orientation preset is applied.
This is what orientation preset is for. To reset them to the default position of CW0FLIP with the plug facing the opposite direction the model travels. 
That's why its called a PRESET. It effectively clears the offset made by manufactures that use any other magnetometer chip position other than CW0FLIP. 
And why does it have align_mag_pitch = 180 once the preset is applied. That's what flip does.

>Sliders do not show rotation which is "additionally applied on top of orientation preset". 

That is not correct. Additional align mag slider setting should be zero after the orientation preset 90, 180 or 270 are applied.
Another way of explaining this. Is if every magnetometer chip was mounted equivalent to CW0FLIP with its plug facing the rear as default, there would be no need of an orientation preset.

As for the XY coordinates. I fail to see the benefit of them being added to a module or to the tool. How are people to know if its X positive / negative, or Y positive / negative.

RomanLut:
@Choolet >why you have not zeroed the magnetometer sliders once the orientation preset is applied

The are two methods to provide orientation of the mag in inav:
a) with **align_mag** preset
b) with **align_max_pitch, align_max_roll, align_max_yaw** angles.

If all (b) are zero, then **align_mag** is used. if any of (b) is non-zero, then (b) is used and **align_mag** is ignored.

This is stated in the title "2. Select a preset (align_mag) **or** create a custom configuration using the sliders
(align_mag_roll, align_mag_pitch, align_mag_yaw)"

So yes, in order for preset(**align_mag**) to take effect, **align_max_pitch, align_max_roll, align_max_yaw** settings should be zeroed.

At the same time, it is usefull to show **align_max_pitch, align_max_roll, align_max_yaw** settings which would provide the same orientation **as  selected preset with current FC rotation**. If user desides to define orientation using (b), he can touch sliders. 

For example: I am experienced user, I know that my FC is not rotated, my matek magnehometer is mounted with arrow forward, so I select CW270FLIP. Then I notice that mag is actually slightly off by 20 degrees, so I use sliders to slightly adjust orientation. Tool switches from preset to align_max_* settings once sliders are touched.

In this PP, I hope I made enough changes in interface to indicate which settings take effect.

If preset takes effect, then **align_max_pitch, align_max_roll, align_max_yaw** are striked out, and CLI string below shows "align_max_pitch=0, align_max_roll=0, align_max_yaw=0" for thouse who understand how settings work:

image

If **align_max_pitch, align_max_roll, align_max_yaw** takes effect, then preset dropdown is striked out:

image


>As for the XY coordinates. I fail to see the benefit of them being added to a module or to the tool. How are people to know if its X positive / negative, or Y positive / negative.

Actually, XYZ model is the most important model in the tool. This single model is enough to setup any magnetometer. 

inav does not care about magnetometer or magnetometer chip manufacturer, arrow marking, default preset etc. All inav needs to know is how magnethometer local frame axes are oriented in the physical setup. 

If magnethometer has XY axis markings, settings can be easily found in tool by aligning XYZ model to physical setup. So in future, if some magnetometer 3D model is not included in tool, setings could be found by alligning axes. 

image

image

Choolet:
>The are two methods to provide orientation of the mag in inav:
a) with align_mag preset
OR
b) with align_mag_pitch, align_mag_roll, align_mag_yaw angles.

 I don't agree with that assessment.
Both (a)`align_mag` and (b)`align_mag_roll, align_mag_pitch, align_mag_yaw` may work on a similar premise. 
**But they are purposely different settings.** 

(a) Should be applied first. `align_mag` is purposely meant to account for the manufactures mounting orientation of the magnetometer chip. With reference to a known marker on the module. Either the plug or arrow.
You practically have it working. But it baffles me why you can't seem to grasp the individual purpose of these two settings.

When any one of the orientation presets are applied. **All the (b) sliders should be reset to ZERO.**  So they can be used to reference mounting position changes by the user. 

> If all (b) are zero, then align_mag is used. if any of (b) is non-zero, then (b) is used and align_mag is ignored.

(a) stands by itself. It should not be rendered void when the sliders are adjusted.
It can be likened to resetting the Trip meter in your automobile. It provides a new distance reference point. But it doesn't alter the Odometer distance, its the base reference.

RomanLut:
**align_mag_roll, align_mag_pitch, align_mag_yaw** are not applied on top of preset (**allign mag**) in inav. 
There are no settings in inav to apply additional rotation on top of preset.

If **CW270FLIP** preset takes effect, then settings are:
**align_mag=CW270FLIP  align_mag_rol=0  align_mag_pitch=0  align_mag_yaw=0**

If magnethometer is off by 1 degree, then settings should be:
**align_mag=[any value because it is ignored]  align_mag_rol=0  align_mag_pitch=1800  align_mag_yaw=2710**

While **CW270FLIP** preset takes effect, I prefer to set sliders in positions **0 180 270** and strke out "align_mag_xxx" title to indicate that sliders values do not have effect. This way, when user touches sliders, magnethometer keep orientation and is not flipped to 0,0,0 (CW0).

DzikuVx:
@RomanLut shall we merge? It looks awesome enought :)

PR #2064, opened 2024-05-10
CLI saved filename fix

breadoven:
Reinstates auto generated filename when saving CLI output to file.


breadoven:
Couldn't get `FC.CONFIG `to work with `generateFilename ` when called from `helpers.js`, just returns undefined. Any ideas why ?

DzikuVx:
@breadoven I allowed myself to do it in a different way in #2066

PR #2067, opened 2024-05-11
Remove modes tab as not used and replaced by auxiliary years ago

DzikuVx:


MrD-RC:
Can we rename auxiliary to modes now. So it matches the actual UI 🤣

DzikuVx:
Later :)

PR #2069, opened 2024-05-11
Allow LED pin to be reassigned as MOTOR or SERVO

mmosca:
Configurator counterpart to iNavFlight/inav#10022

This change requires corresponding firmware.

* [x] Add LED to override options
* [x] LED override option displayed correctly on loading
* [x] LED override option saved correcly 
* [x] Show LED pad as possible output
* [x] Use special label info from MSP to identify original LED pin

Led pin is marked as Sn/LED
image


mmosca:
Output display in configurator does not match fc, when first timer is set to let

PR #2074, opened 2024-05-13
Changing DECORATION to DIRECTION to match firmware define.

bishko:
This PR is linked to the following [PR](https://github.com/iNavFlight/inav/pull/10031)

I think there is a small error in the naming and I have corrected this according to what is used in the firmware files in [`osd_symbols.h`](https://github.com/iNavFlight/inav/blob/master/src/main/drivers/osd_symbols.h).

As metioned in the other PR, I don't know which terminilogy is correct so feel free to change it if needed.


MrD-RC:
It should actually be decoration.

PR #2075, opened 2024-05-15
ADD APM32 MCU USB_DFU

Alden-wrk:
By modifying port_handler.js and manifest.json, you can add the VendorID and ProductID corresponding to APM32 so that it can directly support the firmware download of APM32 series MCU，https://global.geehy.com/


Alden-wrk:
By modifying port_handler.js and manifest.json, you can add the VendorID and ProductID corresponding to APM32 so that it can directly support the firmware download of APM32 series MCU

DzikuVx (approved):

PR #2084, opened 2024-05-18
Hopefully this fixes crash on GPS tab

### LINKED ISSUE #2083: [master] Configurator crashes to front page in GPS tab.
## Description

The 8.0.0 electron configurator crashes frequently in the GPS tab. This seems to be exacerbated with INAV 8 firmware and many 'features' around the GPS location. 

A JS neophyte might suspect the `MSP2_ADSL_VEHICLE_LIST` code getting confused / being broken about  'features' in the map view, for example the a scenario with FW Approach and ADSB vehicles active causes the GPS tab to crash very quickly / without zooming.

`2024-05-18 @ 10:17:06 -- Unexpected error: Cannot read properties of undefined (reading 'lat')`. With fewer (or zero) features active, it is necessary to zoom the map rapidly. 

## Crashtastic scenario

FW Approach and GA (ADSL) provoke crash in short order.

Untitled

## Debug view

(Earlier example, zooming only).

Screenshot from 2024-05-18 09-36-49

## Caveat

JS neophyte. Above is speculation / observation not analysis.

DzikuVx:
fixes #2083

### DIFF
--- a/tabs/gps.js
+++ b/tabs/gps.js
@@ -229,8 +229,8 @@ TABS.gps.initialize = function (callback) {
             var feature = mapHandler.forEachFeatureAtPixel(mapHandler.getEventPixel(evt.originalEvent), function(feature, layer) {
                 return feature;
             });
-
-            if (feature) {
+            
+            if (feature && feature.get('data') && feature.get('name')) {
                 TABS.gps.toolboxAdsbVehicle.setContent(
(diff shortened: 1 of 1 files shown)


stronnag:
Tested on Linux x86_64 using the same "crash quick" scenario (safehome, fwapproach, ADSB) as before.

Did not crash this time.

No ADSB symbols, but that's a different issue.

PR #2089, opened 2024-05-20
fix display of ADSB contacts

stronnag:
* Enable display of aircraft west of prime meridian and south of the equator
* Use `Callsign` to label `callsign` field (vice `icao` which is something different).


stronnag:
Screenshot from 2024-05-20 16-26-58

PR #2102, opened 2024-06-01
Remove GPS serial configuration note

JulioCesarMatias:
Now that we can configure the GPS serial port in the GPS tab itself, I think this message is not very useful...
Captura de Tela 2024-06-01 às 19 11 37


MrD-RC:
It could also be changed to put a red ring around the serial port box if one is not selected. But the GPS has been enabled. This is done on Advanced Tuning for things like linear descent selected but no home altitude set; in the RTH settings.

PR #2106, opened 2024-06-06
Serial gimbal support

mmosca:
Configurator counterpart for iNavFlight/inav#10109

Currently only include the changes to the ports tab.

May eventually add support for the configuration settings.


mmosca:
While adding the modes, I grouped  some items together to make them easier to spot.

image

mmosca:
This is an example of the setup on the ports tab.
image

In this example, UART6 can be used if headtracker type is set to MSP on the configuration tab and UART4 can be used if type is SERIAL.

mmosca:
Current configuration tab options, subject to change:
image

shannonbaker:
[AUTO] selection of input is not working for Gimbal modes.

PR #2124, opened 2024-06-23
Get latest release instead of list of all releases in version checker

bfmvsa:
It is possible to retrieve the desired release directly instead of the approach when you get a whole list of releases and then get the latest one from the sorted list.

Currently, the number of releases retrieved from `/repos/iNavFlight/inav-configurator/releases` equals to 29. That's approximately **600KB** of data retrieved every check in json.
After this update the size of loaded json through `/repos/iNavFlight/inav-configurator/releases/latest` is about **24KB**

### DIFF
--- a/js/appUpdater.js
+++ b/js/appUpdater.js
@@ -10,12 +10,11 @@ var appUpdater = appUpdater || {};
 
 appUpdater.checkRelease = function (currVersion) {
     var modalStart;
-    $.get('https://api.github.com/repos/iNavFlight/inav-configurator/releases', function (releaseData) {
+    $.get('https://api.github.com/repos/iNavFlight/inav-configurator/releases/latest', function (releaseData) {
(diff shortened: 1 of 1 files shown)


bfmvsa:
Sorry guys for the random pull requests, but I'm starting to research the repo and as a side effect trying to be useful if it is possible. In this specific PR I figured out that we don't need to retrieve all list of inav releases to check if current version is outdated. After this fix we will save to inav users about half of megabyte every check.

mmosca:
I am not sure this particular one is a good idea. Sometimes INAV has bugs, or someone wants to check if an issues happened on an older release. Fetching only the latest release, while faster, does not allow for these use cases. It also breaks showing pre-release versions.

bfmvsa:
Yes I agree, the title of the pr without checking the changed files is unclear. Changed a little bit the title.

PR #2144, opened 2024-07-11
calibration.html: Make width match (full)

sensei-hacker:
Finishes #2116 by @JulioCesarMatias


sensei-hacker:
Original layout. One section is 3/4 width, the other 1/4 next to it.
screenshot-34-14

#2116 leaves one box at 3/4 width and removes the 1/4 width box, leaving an empty hole.
screenshot-34-full

This PR allows both sections to go full width, filling in the hole.
screenshot-full-full

PR #2146, opened 2024-07-15
clear separation between mixer preset and wizard

sensei-hacker:
Moves the mixer wizard button out of the mixer preset section because of user confusion.
Users expected the mixer preset stuff like "load and apply" to be tied to mixer wizard because of the way the elements were positioned on screen.  They would try to "load and apply" the results of the wizard, etc.

Now:
image

### DIFF
--- a/locale/en/messages.json
+++ b/locale/en/messages.json
@@ -2096,7 +2096,7 @@
         "message": "Mixer <span style=\"color: #37a8db\">saved</span>"
     },
     "mixerWizard": {
-        "message": "Mixer wizard"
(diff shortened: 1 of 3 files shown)


mmosca:
Should it be renamed to Motor Mixer Wizard?

At least gives users a hint it is going to do something with the motors.

sensei-hacker:
> Should it be renamed to Motor Mixer Wizard?
> 
> At least gives users a hint it is going to do something with the motors.

Your wish is my .. well it's a pretty good suggestion. Done.

PR #2154, opened 2024-07-24
Add ability to add GEO files to mission planner map.

paulwrath1223:
I basically just stole the example from openlayers, and then fixed a bug where the custom drag handler for waypoints was interfering with panning over imported zones.

This feature is the bare minimum, but it does work, and mostly importantly it doesn't change any functionality. The only way a user would even find out this change was made is if they happen to drag a KML or other geo-file over the map.

Why did I add this? 
I was doing WP mission over a massive plot of land and had trouble identifying landmarks, as well as making sure to stay on 'my' side of the property line. I have the property lines and trails and waypoints saved as KML files for a different project, and figured it would also be useful for staying out of restricted airspace. The advantages for surveying would also be numerous.

"Isn't this just budget GeoZones?"
Yes. But also no. GeoZones will be priceless for staying out of certain zones, but I believe there is an advantage to keeping some features only visible locally in the configurator. While I am trying to work with the developer working on GeoZones to add support for importing them from other geo-files, like KML, the fact that this feature is so unobtrusive makes me think that even if I am the only person in the world who will ever use it, it's still worth it. GeoZones are also limited by the on-board EEPROM, meaning my example of the property line wouldn't work because it has too many vertices.

### DIFF
--- a/tabs/mission_control.html
+++ b/tabs/mission_control.html
@@ -469,7 +469,7 @@
             </div>
         </div>
         <div class="cf_column threefourth_left" style="height: 100%;">
-            <div id="geo_info">&nbsp;</div>
+            <div id="geo_info" style="display: block; text-overflow: ellipsis; white-space: nowrap; overflow: hidden;">&nbsp;</div>
             <div id="missionMap"></div>
             <div id="missionPlannerElevation" class="gui_box grey" style="display: none">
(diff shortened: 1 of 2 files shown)


paulwrath1223:
I'm going to take this back while I add a way to toggle layer visibility

PR #2183, opened 2024-09-10
fix copy OSD layout for electron

error414:
It's fix for eletron, it seems approach to promises is not same in NWJS and electron.

### DIFF
--- a/tabs/osd.js
+++ b/tabs/osd.js
@@ -3160,22 +3160,26 @@ OSD.GUI.updateAll = function() {
             }
         });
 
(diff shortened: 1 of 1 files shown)


Jetrell:
I can confirm this function was broken for Windows.. It wouldn't save the pasted OSD page.
However this commit has now fixed the issue. 👍

MrD-RC:
I've just come across this error. If @Jetrell has tested and it is fixed. I'm happy with that.

MrD-RC (commented):
All good

PR #2194, opened 2024-09-19
Fix issue with saving firmware files to local machine

bfmvsa:
Looks like we had an issue saving downloaded firmware files to a local machine (Because of outdated chrome api). Fixed and tested on MacOS, if it is possible, please check on the other platforms before merging.
Also fixed filename of the files from always "inav.hex" to specific for downloaded firmware.

screenshot 2024-09-19 at 09 20 05

### DIFF
--- a/tabs/firmware_flasher.js
+++ b/tabs/firmware_flasher.js
@@ -32,7 +32,8 @@ TABS.firmware_flasher.initialize = function (callback) {
     }
 
     var intel_hex = false, // standard intel hex in string format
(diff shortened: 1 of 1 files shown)


bfmvsa:
Maybe my description of the bug was unclear, so I'd like to explain it again.

When you load firmware by using "Load Firmware [Online]" you can download the .hex file to your local machine by clicking on the link in the progress bar (highlighted in the screenshot), it is not working because of outdated Chrome API, in this PR I've fixed it by using `fs` library.

PR #2198, opened 2024-09-21
Update the CSS to avoid hiding the save and reboot button

MrD-RC:
Before
INAV Configurator - before

After
INAV Configurator - after

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -1001,7 +1001,7 @@ li.active .ic_mag {
 #content {
     margin-top: 0;
     padding: 0;
-    height: calc(100% - 145px);
+    height: calc(100% - 150px);
     /*  (port picker 105px, log CLOSED 25px, status bar: 20px + padding) - was: calc(100% - 171px)*/
     background-color: white;
     overflow-x: hidden;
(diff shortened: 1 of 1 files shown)


Jetrell:
@MrD-RC I'm not sure if its effects some monitor screen differently. 
This is what I had **before** this commit.

before


And **after**. Which has moved the Save reboot button higher and left a blank space below it. Shrinking the dialog size.. 
It effects all tabs with Save reboot button, except the Configuration tab.

after

MrD-RC:
Are you on mac or windows? Which browser?

Jetrell:
Windows... I use Firefox.
I loaded Chrome as default for a test. And that returned it to what it was.

MrD-RC:
Damn browsers and their different implementations. I'm using Brave, which is based on Chromium. Actually, I don't get why that impacts the application. The whole thing used to be running as. Chrome app. I guess moving to Electron had changed that?

Jetrell:
>Damn browsers and their different implementations

It seems that way. I also tried it with Opera and Edge browsers.. Unfortunately both of them had the same issue as Firefox.

MrD-RC:
I might have to try some nasty browser specific css 😕

Jetrell:
I gave it a go with Brave as you use. And it worked.
I also thought Opera should work. Because it has a Chromium license.. And after hammering the _Flush Setting Cache_ button a few times. I got it to work... But I still couldn't get Firefox and Edge to work when doing the same. Even after an FC and PC reboot.

MrD-RC:
Reverting as the issue seems to only affect some pages, such as configuration.

PR #2200, opened 2024-09-23
Fit width of sensor according to width of text

bfmvsa:
This PR truncates text in the sensors block.

Before:
screenshot 2024-09-22 at 11 11 06

After:
screenshot 2024-09-22 at 11 17 38

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -277,7 +277,7 @@ input[type="number"]::-webkit-inner-spin-button {
 #sensor-status li {
     float: left;
     height: 67px;
-    width: 35px;
+    min-width: 33px;
     line-height: 18px;
     text-align: center;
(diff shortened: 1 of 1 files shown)


MrD-RC:
I don't think this is the best way to get around that problem. I think that there should be alternative language css files which can change the size of elements. IMHO the sensor boxes should be made a little wider. So that the text can fit. Truncating isn't readable.

sensei-hacker:
> @MrD-RC No problem, please check this 

This is the way, thank you!
The browser **already knows** how much space the text needs. The right thing is for the CSS to NOT force a size.  "Don't prevent the browser from laying out the boxes so things fit". 

This is so much smarter than piling on more CSS trying to manually do what the browser was made to do, and will do by default if you let it. :)

PR #2202, opened 2024-10-01
vtx: fetch powerlevel count from fc

bkleiner:
instead of duplicating the logic to determine the number of power levels, we fetch them from the fc
needs https://github.com/iNavFlight/inav/pull/10395

### DIFF
--- a/js/fc.js
+++ b/js/fc.js
@@ -354,6 +354,9 @@ var FC = {
             power: 0,
             pitmode: 0,
             low_power_disarm: 0,
+            band_count: 0,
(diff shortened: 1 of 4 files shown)


sensei-hacker:
There is a merge conflict due to power min=0 in some cases. Can you please resolve or comment?

sensei-hacker:
@bkleiner  ?

sensei-hacker:
Superseded by #2486 which resolves the merge conflict and adds power_min support with backward compatibility.

PR #2204, opened 2024-10-07
Change gimbal sensitivity from programming framework

error414:
https://github.com/iNavFlight/inav/pull/10405


MrD-RC:
Please separate these out in to 2 PRs

Also, I think the gimbal sensitivity adjustment would be better as an adjustment.

error414:
1) fixed 
2) I'm not sure I understand

MrD-RC:
What do you mean? 

Gimbal sensitivity adjustments would be better suited as "adjustments" rather than in the programming framework. I put more detail in a PR you made on the firmware side.

Issue #2208, opened 2024-10-15
Configurator crashes after "heads-up settings"


error414:
Inav configurator crashes if you change anything in OSD tab, in "Heads-up Display settings". It's casued by https://github.com/iNavFlight/inav-configurator/blob/master/js/gui.js#L129 .

These elements have set class "toggle" https://github.com/iNavFlight/inav-configurator/blob/master/tabs/osd.html#L394  but I would say that it's wrong, because inputs are types "number", not checkbox. 

It seems that if we incialize switchery for "number" input type, then switchery is not inicialized properly and some propertirs are missing inside switchery
![wrong_toogle](https://github.com/user-attachments/assets/02c6542c-acde-4c0a-9ff8-2f32acc0f0f2)
And mainly it does not make sense to inizialized switchery for input type number. 

I would fix it, but changes around switchery are old and it works for inav 7.xx, so maybe it's connected with electron.

PR #2210, opened 2024-10-16
dont init switchery for non checkbox inputs

error414:
it's fix for issue https://github.com/iNavFlight/inav-configurator/issues/2208


MrD-RC:
Works fine. Just need the mac errors resolved before merging.

error414:
I would say that osx build more times does not work than works. If you look at error, it's probably not caused by my changes.

error414:
ok, fixed :), I did not change anything and it passed :D

MrD-RC:
It can be a pain in the ass. I re-ran it too. Just in case it was a glitch. Bloody macs.

Issue #2209, opened 2024-10-15
Loading OSD tab is so slow


error414:
Loading OSD tab is incredibly slow. It's caused by calling updateOSDCustomElementsDisplay. 

Loading custom OSD elements fire "change" event for each filled input in custom element block.  So if four custom elements are used. then it means 4 * 4 = 16, 16 calling of updateOSDCustomElementsDisplay, in updateOSDCustomElementsDisplay is called OSD.GUI.updatePreviews(); what is incredibly slow if it's called in the loop. 

So for future be careful about calling OSD.GUI.updatePreviews();. 

![slow_osd](https://github.com/user-attachments/assets/986920a4-a0d2-4586-a604-959992a94713)

MrD-RC:
Looks like it would be good if change events could be blocked until the page has finished loading. Then call the update previews once the page has finished loading.

PR #2211, opened 2024-10-16
speed up OSD tab loading if there are too many custom OSD elements.

error414:
Fix for https://github.com/iNavFlight/inav-configurator/issues/2209


MrD-RC:
Works fine. If the error on the mac build can be fixed. We can get this merged.

error414:
uff, it passed as well. you are right, it's better to see all green

PR #2228, opened 2024-11-03
Fix missing scrollbars on a couple of pages

MrD-RC:
@Jetrell Hopefully this is the last of these changes. All pages have scrollbars. No double scrollbars, except for the OSD page, if the screen is smaller than the OSD preview.


Jetrell:
Looks good now.. Thanks mate.

PR #2229, opened 2024-11-04
Fix OSD previews for diatance precision

MrD-RC:
Fixed OSD previews for
- Home distance
- Remaining distance
- Trip distance


Jetrell:
Thanks again.

PR #2235, opened 2024-11-07
Edit osd_camera_uptilt_help for clarity

chasdabigone:
Added the exact variable osd_ahi_camera_uptilt_comp in the help message so it is clear how to enable it


sensei-hacker:
Thanks for this!
While you're at it, in that message could you change "degres" to "degrees"?

PR #2237, opened 2024-11-08
Geozones

Scavanger:
Configurator part for https://github.com/iNavFlight/inav/pull/10459


stronnag:
Crashes when the Mission Control tab is opened when connected to a (SITL) FC holding valid Geozones. 
Looks to these untrained eyes that the helper for MSP2_GEOZONE is broken (since the update to add AMSL option).

stronnag:
On Linux at least, it is not possible to create a polygon with more than four points.
image

Scavanger:
> On Linux at least, it is not possible to create a polygon with more than four points.

Try clicking on the line between two vertices.

PR #2241, opened 2024-11-14
Add AUTOHD mode

mmosca:
Add new option to negotiate OSD canvas size with MSP_OSD_CANVAS messages.

Basic support. Canvas is not resized with FC info yet, but it auto adjusts  if a message is received.

Counter part to iNavFlight/inav#10465

TODO:
* [ ] Send MSP_OSD_CANVAS message to FC and interrogate canvas size.


sensei-hacker:
I'm not sure what this means:

> Canvas is not resized with FC info yet, but it auto adjusts if a message is received.
> TODO:
>   Send MSP_OSD_CANVAS message to FC and interrogate canvas size.

is this something that would be useful to merge as it currently sits?

Issue #1940, opened 2024-01-12
GCC 13.2 and profile/mixer, increasing settings load lockup occurrence


Jetrell:
The matter of DIFF/DUMP introduced settings errors, or locking up the compiler when loading, is not new, when the DIFF ALL has multiple `PID` and `Battery` Profiles applied.. This has been evident with F411's, since back in the 3.0 releases.

But this is also now experienced in 7.0, with F405's, since the addition of the Profile mixer switch.
And now that the compiler has been upgraded for 8.0. Its even starting to effect F722 controllers when loading settings.

In my experience it does not occur every time. So it may not be seen by all users.. But for someone as myself who is often flashing fw and loading settings for testing purposes. Its has become far more noticeable. With the only workaround, being to load the DIFF ALL in two parts.

It would appear there is a fixed transfer rate between the Configurator and the flight control ? Due to it effecting processors in order of their clock speeds, with the introduction of more demanding background tasks being placed upon them.

stronnag:
To add a bit more context, following some discussion with @Jetrell 

* The hack in the configurator to use arbitrary delays is an inadequate solution, masking the real issue.
* The problem is serial overrun due to the configurator not "pacing" the delivery of data to that the FC can process.
* For example, `profile X` takes c. 1 second to return to a CLI prompt. The configurator can swamp the FC during this time.
* mwp's `fcset` tool does not use a delay, instead it awaits the FC sending the prompt "\n# " before sending the next line, so it can't overrun the FC's serial processing.
* @jetrell has reported that a further configurator hacked to increase the delays does alleviate the immediate issues. 
* While increasing the extant delays from 50ms to 100ms for normal line delay and from 200ms to 1000ms for `profile` delay did alleviate problem, it would, in this reporter's view, be preferable to adopt an "await prompt" solution. 
* The "await prompt" solution has the additional benefit of increasing throughput as it does not incur delay when it is not necessary.

rts18:
I concur with your observations gentlemen. I too have experienced the matters. Both loading 2 or 3 profiles. And more so of late, since the GCC13.2 tool chain commit.

stronnag:
Just for "fun", I instrumented mwptool's `fcset` to measure the time elapsed from writing a line command to receiving the new CLI prompt "\n# ".  This is on a F405 bench test FC,  recent INAV master (8.0), demonstrating that the current delays are an inadequate provision against overrun.

| Index | Time(s) | Command |
[...]

PR #2247, opened 2024-11-16
Update name change from `profile` to `control_profile`

MrD-RC:
Fix for RC2


stronnag:
> Just thinking. I know these delays aren't the best way to do this. But shouldn't there also be a delay when switching the `battery_profile` or `mixer_profile` too?
> 
> @stronnag @Scavanger

Yes, they should. All these `*profile*` items are more expensive than average. I re-ran the #1940 instrumentation, and the figures are pretty much the same for RC1.

Issue #2246, opened 2024-11-16
8.0 - RC1 Mission Control not functioning if Geozones enabled on unsupported Flight Controller


b14ckyy:
To reproduce: 
- Flash INAV on F722 or F411 FC that is excluded from Geozones Feature due to lack of flash and ram
- Turn on Geozones in the Configuration panel (cannot be removed if unsupported but should not have an effect)
- Save settings
- Open Mission Control panel

Result: Configurator will stuck at "Loading Data" 
![image](https://github.com/user-attachments/assets/1db6b96b-c079-49fb-8e03-5d1876bc740e)

Workaround: Keep Geozone feature disabled on unsupported FCs.

b14ckyy:
@Scavanger

PR #2248, opened 2024-11-16
Geozones RC2 fixes

Scavanger:
- Fix for #2246 (see also https://github.com/iNavFlight/inav/pull/10471)
- Add sanity cheks


b14ckyy:
Tested and confirmed, the empty Geozone settings box is not there anymore. 👍

PR #2250, opened 2024-11-17
Localization: Add Japanese Translation

ToshihiroMakuuchi:
This is a sudden pull request, but I wanted to enable Japanese localization, so I created the dictionary file messages.json.
Additionally, I have added Japanese as a selectable option alongside en and uk.

In the current version (8.0.0 RC1), it seems that Demo mode is not working properly. However, in version 7.1.2, the Japanese localization appears to be functioning as expected.

One issue is that the font for Japanese differs from the one used in Betaflight Configurator, making the appearance less readable. I am unsure where to modify the code to match the font with Betaflight Configurator. If you know how to adjust this, I would appreciate your guidance.

Thank you!


b14ckyy:
Thanks a lot for that. I can't give you any info regarding the font unfortunately. Maybe a developer can answer that. Afaik its hard coded and can't be switched based on the localization right now. 

It might take a bit until a localization is added as it needs to be translated back and doublecheck by someone else before we can merge it. Appreciate your work.

ToshihiroMakuuchi:
@b14ckyy 

Thank you for your previous response and the guidance regarding localization.

I have created a Japanese version of the messages.json file based on the latest version as of November 24th. I carefully reviewed all keys to ensure consistency with the English version and made adjustments where necessary. Additionally, I have revisited the translations to ensure that they are accurate and natural in Japanese.

I understand that the localization will need to be reviewed and double-checked before it can be added. If there are any further steps required, such as addressing specific localization concerns or providing additional context, please let me know. I am happy to assist in any way I can.

It is an honor to contribute to this project. I look forward to receiving feedback from the team.

YI-BOYANG:
You need to resolve the language switching option in the configurator before you can add Japanese files. Otherwise, when a computer using a Japanese system opens the configurator, it will always display Japanese and cannot be switched to English or any other language.

sensei-hacker:
The language should also be added to line #13 of js//ocalization.js like so:

const availableLanguages = ['en', 'ja' 'uk'];

That will update the language switched drop-down.

The proper font can be set by adding to the CSS:

*:lang(ja) {
...
}

sensei-hacker:
In #2271 I did some small changes to support having language-specific CSS< such as setting a font appropriate to the language.

I also did a PR against Toshihiro's branch with three changes - add ja to the list of supported languages, add the name of the language, and set the font.

https://github.com/ToshihiroMakuuchi/inav-configurator/pull/1

sensei-hacker:
> You need to resolve the language switching option in the configurator before you can add Japanese files. Otherwise, when a computer using a Japanese system opens the configurator, it will always display Japanese and cannot be switched to English or any other language.

Are you referring to adding it to the switcher as I have here?
https://github.com/iNavFlight/inav-configurator/commit/3891ba5811d6dcc4a2197e551b3efcd3520b4910

mmosca:
Can you fix the conflicts? They are probably a result of merging other languages

sensei-hacker:
I did a merge back from master to fix the conflicts from the addition of Chinese.

sensei-hacker:
I merged master back to fix the conflicts.

PR #2258, opened 2024-11-20
... Various minor bugfixes

Scavanger:
#2253 was merged a bit quickly. :D

Here it goes on...


b14ckyy:
Looking good so far. 
Parachute icon works, Mode Channel detection is fine as well. No new issues found. 

Only remainder is, that the popup for Clearing a mission, still triggers the Input field focus issue. Its fine for the Geozone popup but still happens after the Clear Mission Popup. 
Switching application focus solves it as a workaround.

Issue #2262, opened 2024-11-23
8.0-20241120-5019b12 Missing Motor Numbers in Outputs tab for MR


b14ckyy:
The Outputs Tab misses the numbering on the Motor Graphic to identify motor positions and test order: 

![image](https://github.com/user-attachments/assets/60f43516-7a3f-4079-b6e7-58c27c1ed1b0)
![image](https://github.com/user-attachments/assets/562a7682-c885-4b48-9b84-8cc23f26bfea)

mmosca:
Did it have that before? I think it never did. It can be helpful though.

mmosca:
> I thought it did at some point. Without them, identifying the motors can be hard as I need to screenshot or not down from the mixer page.

The motor numbers are usually marked on the esc and the mixer is needed to understand where they are in the quad, in case someone had it mixed up by rotating the esc.

I do think it is a good idea overall, but it needs to match what the mixer thinks it is, and potentially, have the motor order wizard here as well

b14ckyy:
Motor wizard is a neat feature but should get an update. Its a long time since I did with BF but I think there it twitches a motor with short pulses of 10% throttle and you click on a graphic what motor is spinning. This way the mixer is set up.

sensei-hacker:
Fixed in https://github.com/iNavFlight/inav-configurator/pull/2263


> I do think it is a good idea overall, but it needs to match what the mixer thinks it is, 

It does now. It used to have a static image with the default positions from the preset.
Then it had none. Now it has them correct, matching the mixer.

Caused by https://github.com/iNavFlight/inav-configurator/pull/1910
The preview image from the mixer was reused on the outputs tab.

PR #2263, opened 2024-11-23
Add back motor numbers on Outputs tab preview diagram

sensei-hacker:
Addresses https://github.com/iNavFlight/inav-configurator/issues/2262 "Missing Motor Numbers in Outputs tab for MR"

### DIFF
--- a/src/css/tabs/motors.css
+++ b/src/css/tabs/motors.css
@@ -472,6 +472,7 @@
 .tab-motors .motorNumber {
     position: absolute;
     font-size: 1.4em;
+    visibility: hidden;
 }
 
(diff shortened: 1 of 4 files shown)


sensei-hacker:
It occurred to me this may need to first check whether there are four motors. So that it doesn't do anything if it's not a quadcopter.
I did this PR when I was tired and only tested that it did what it was supposed to do. I forgot to test that it doesn't do anything when it's not supposed to.  So I'll try to update it in the next couple days.

sensei-hacker:
The concern I mentioned in my last comment has been addressed. It should be good to merge now.

Issue #2265, opened 2024-11-26
Total mess in configirator on small screens


and-sh:
1024x768 resolution, Debian. Gnome, cannot reach save button, can not reach CLI.
Large font (accessibility) lead to worst results.
![Screenshot From 2024-11-26 19-00-06](https://github.com/user-attachments/assets/cd6d2b24-8c66-4b49-81dd-5107435ca405)

and-sh:
I found some solution. If I press Tab several times it switch  focus to buttons and they become partially visible. same with cli.

btw Linux is becoming noticeably more convenient than Windows. For example, I use different desktops for different tasks, which allows me to switch between them quite quickly.
![Screenshot From 2024-11-26 20-55-12](https://github.com/user-attachments/assets/71a07c3b-0aae-41d9-a704-c7bf384bf8f6)

MrD-RC:
I believe this has been fixed in 8.0.

and-sh:
This is 8

I guess that some minor parameters of window are incorrect. All content are in places. I can switch between them by Tab. I can not scroll by sliders only.

Found a new solution. Just stretch the window a little.
![Screenshot From 2024-11-27 07-30-01](https://github.com/user-attachments/assets/1d760537-dbb4-4d96-bc82-df76df437346)
![Screenshot From 2024-11-27 07-30-49](https://github.com/user-attachments/assets/d3220b1b-4ef2-4475-ba5c-83871d2dc5f9)

MrD-RC:
With a resolution that size, you should run Configurator maximised.

MrD-RC:
@and-sh Could you try using this configurator https://github.com/iNavFlight/inav-configurator/actions/runs/12065115088?pr=2266

and-sh:
On Debian, Gnome work fine, thank you very much
![Screenshot From 2024-11-29 11-28-16](https://github.com/user-attachments/assets/9d42fc11-1de8-4b34-956a-2b9306ac58d9)

PR #2266, opened 2024-11-27
Rearrange top bar to fit on 1024px screens

MrD-RC:
There is an issue on 1024px wide screens where the dataflash box spills over on to a new row. This hides the save and reboot button. This change has a slight rearrangement, stacking the profile selection boxes vertically. This considerably reduces the width taken up by this. I have also made it so that the dataflash box will hide if the screen is too narrow (less than 1060px).

This needs testing with a flight controller with dataflash connected. Just to make sure the dataflash area is ok when populated.

Fixes https://github.com/iNavFlight/inav-configurator/issues/2265

Current layout on large screen
image

Current layout on 1024 x 768 screen
image
Notice missing save and reboot button

New layout on large screen
image

New layout on 1024 x 769 screen
image

Technically a bug fix. So should be fine for the next 8.0 RC if the change is wanted.

### DIFF
--- a/index.html
+++ b/index.html
@@ -88,13 +88,7 @@
                 </div>
             </div>
             <div class="header-wrapper">
-                <div id="dataflash_wrapper_global">
-                    <div class="noflash_global" align="center" i18n="sensorDataFlashNotFound"></div>
-                    <ul class="dataflash-contents_global">
-                        <li class="dataflash-free_global">
-                            <div class="legend" i18n="sensorDataFlashFreeSpace"></div>
-                        </li>
-                    </ul>
+                <div id="profiles_wrapper_global">
                     <div id="profile_change">
                         <div class="dropdown dropdown-dark">
                             <form name="profile-change" id="profile-change">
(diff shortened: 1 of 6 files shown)


Jetrell:
This was with my screen scaled down to 1024 x 768. 

1024 x 768

And this with the default monitor size on the PC I'm using ATM. With 1366 x 768.

1366 x 768

While it looked like this before these changes. I also rebooted the PC to be sure.

1366 x 768 before these changes

Jetrell:
This was with my screen scaled down to 1024 x 768. 

1024 x 768

And this with the default monitor size on the PC I'm using atm.. It is 1366 x 768.

1366 x 768

While it looked like this before these changes. I also rebooted the PC to be sure.

1366 x 768 before these changes

The Save/Reboot buttons appear fine in the tabs at the lower resolution.
But I can't get the OSD tab to populate at all with this build, at either resolution.

MrD-RC:
That's as expected, other than the visible html on the dataflash box. I will get that sorted tonight. 

At 1024px wide. The dataflash area should not be visible. There is only so much space. And that seemed like the least important part to keep. The sensors, profiles, and failsafe/warning areas seem more important to me. This stop the top bar from spilling over and hiding the save and reboot button.

MrD-RC:
After latest change

Full screen
image

1024 x 768
image

PR #2271, opened 2024-12-02
support per-language CSS / fonts (for Japanese etc)

sensei-hacker:
Some language translations need different CSS, such as an appropriate font.

That's done by setting the current language in the root html tag, such as:
[html lang="en"]

Or
[html lang="ja"]

This PR adds a case to allow:
[html i18n_lang="currentLanguage"]

This will allow an appropriate font to be used for Japanese ( #2250 ), and any other language-specific CSS.
To have an example of usage, I have added one line to the CSS to set an appropriate font when Japanese is selected.


sensei-hacker:
> Needs to add currentLanguage to recently merged chinese translation

That has been added.

PR #2272, opened 2024-12-05
Add Simplified Chinese language to the configurator

YI-BOYANG:
**Glad to see multilingual support added to the 8.0.0-RC1 program. Since INAV configurator has only been available in English for a long time, but there are a lot of Chinese players using INAV, and the English version of the configurator is very inconvenient to use. Therefore, starting from the 4.0 version in 2022, I continue to produce the Chinese version of the INAV configurator for Chinese players, and the total number of downloads distributed by various platforms is expected to reach more than 20,000 people. To prove that there is a very large group of players using INAV in China, this is the download data of some platforms:**
image
image

****

When making the Chinese version of the INAV program, I found that there were many strings that had not been localized, and I had to manually modify these characters to Chinese every time a new version was released. I then submitted three PRS to add these strings to the localization file. At that time, I thought about submitting the Chinese localization file to PR, but because the INAV configurator was forced to select the language version according to the computer system language, which could not be manually modified, I did not have the ability to add the function of manually switching the language to the configurator for the time being. After communicating with the developer, I did not submit PR, but continued to make the Chinese version of the configurator to distribute to players.

****

Now someone has finally added the language switching function, thank you for his contribution! I think it's time to end the manual production of Chinese INAV configurator and let the official project also add Chinese language support!

****

Although the Chinese language review is a bit cumbersome and may not be merged as quickly, at least the work of enabling the configurator to support Chinese language is a big step forward.

### DIFF
--- a/js/localization.js
+++ b/js/localization.js
@@ -10,7 +10,7 @@ const i18nextXHRBackend = require('i18next-xhr-backend');
 const Store = require('electron-store');
 const store = new Store();
 
-const availableLanguages = ['en', 'uk'];
+const availableLanguages = ['en', 'uk','zh_CN'];
 
 const i18n = {};
   
--- a/locale/en/messages.json
+++ b/locale/en/messages.json
@@ -14,6 +14,10 @@
         "message": "українська",
         "_comment": "Don't translate!"
     },
+    "language_zh_CN": {
+        "message": "简体中文",
+        "_comment": "Don't translate!"
+    },
     "language": {
         "message": "Language"
     },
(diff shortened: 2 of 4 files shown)


YI-BOYANG:
https://github.com/iNavFlight/inav-configurator/pull/1756

YI-BOYANG:
image

mmosca:
Thanks! You can verify it with the next nightly build. We will also need help to keep it up to date in the future.

YI-BOYANG:
> Thanks! You can verify it with the next nightly build. We will also need help to keep it up to date in the future.

I'll keep an eye on it. Thank you!

PR #2307, opened 2025-01-17
First attempt at adding arm64 linux sitl binaries

mmosca:
Hopefully, this will work on recent raspberry pi computers.


mmosca:
@stronnag and @b14ckyy this should hopefully get the SITL bits workings on the pi as well.

PR #2321, opened 2025-01-22
allow options tab to scroll

sensei-hacker:
As requested by @b14ckyy 

Old:

image

New:

image

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -512,6 +512,8 @@ input[type="number"]::-webkit-inner-spin-button {
     background-color: white;
     border-radius: 5px;
     z-index: 10000;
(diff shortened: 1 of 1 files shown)


src/css/main.css
```diff
@@ -1993,7 +1995,6 @@ dialog {
         height: 24px;
         padding-bottom: 0;
         margin-bottom: 5px;
-        float: left;
```
mmosca, line 1996: Was this change intended? Or accidental?

sensei-hacker: That is intentional. Unfortunately I can't check every possible scenario with every possible label for side effects, but it does only apply on the narrow screen view.

Issue #2324, opened 2025-01-24
[INAV 8.0.0] Mixer speed value overflow


davidk-ro:
In the Mixer tab, setting speed larger than 127us/s causes overflow (e.g. set 128us/s will cause it to display -127us/s)
The value should be an unsigned char 8bit or at least some check so it can't overflow.

sensei-hacker:
Fixed in https://github.com/iNavFlight/inav-configurator/pull/2326 , assuming that gets merged.

PR #2326, opened 2025-01-25
servo mixer send uint values

sensei-hacker:
In the servo mixer, the following are **UN**signed ints:

Servo number
Input source
Speed

Because they populate a servoMixer_t:
https://github.com/iNavFlight/inav/blob/d1ef85e82d8aa5bb8b85e518893c8e4f6ab61d6e/src/main/flight/servos.h#L128

They are read as unsigned values:
https://github.com/iNavFlight/inav/blob/d1ef85e82d8aa5bb8b85e518893c8e4f6ab61d6e/src/main/fc/fc_msp.c#L2271

See:
https://github.com/iNavFlight/inav-configurator/issues/2324


MrD-RC:
Could this be rebased on the 8.x.x branch. This should be in the next release.

PR #2328, opened 2025-01-26
Group nightly releases and limit to 10 nightlies per major

mmosca:
Should ensure we can see both 8.x.x nitghtlies and 9.x.x nightlies.

May still have issues if we have more than 50 releases in a row for 8.x.x or 9.x.x


mmosca:
@b14ckyy this is what I planned to do, on listing the releases. It is also easy to make sure it does not show nightlies older than configurator release, but that has not been implemented.

PR #2346, opened 2025-02-19
smartaudio 8 levels

sensei-hacker:
Companion to https://github.com/iNavFlight/inav/pull/10261 now that TBS is making some VTX units with 8 power levels.


bkleiner:
competing fix here https://github.com/iNavFlight/inav-configurator/pull/2202 might prefer ours because its so much simpler.

sensei-hacker:
Superseded by #2486 which dynamically reads power level count from firmware, making the hardcoded 8-level change unnecessary.

PR #2354, opened 2025-03-07
Add vertical throttle gauge

MrD-RC:
- Added functionality to configurator
- Updated analogue fonts
- Updated digital fonts
- Added new font symbols to `INAV Character Map.md`

Goes with https://github.com/iNavFlight/inav/pull/10725


antlestxp:
I don't see the new graphic with DJI air units. Is DJI missing the character set to display it?

Issue #2369, opened 2025-04-20
Geozones causing issues with missions


breadoven:
Trying to add a multimission in Mission Control without FC connected. Works OK for a while then the following errors start appearing which causes the whole thing to crash.

2025-04-20 @ 11:07:44 -- Unexpected error: Cannot read properties of null (reading 'geozoneCount')

2025-04-20 @ 11:07:52 -- Unexpected error: Cannot read properties of null (reading 'geozoneCount')

2025-04-20 @ 11:07:54 -- Unexpected error: Cannot read properties of null (reading 'geozoneCount')

2025-04-20 @ 11:08:00 -- Unexpected error: Cannot read properties of undefined (reading 'getLonMap')

2025-04-20 @ 11:08:03 -- Unexpected error: Cannot read properties of null (reading 'geozoneCount')

2025-04-20 @ 11:08:06 -- Unexpected error: Cannot read properties of undefined (reading 'getLonMap')

2025-04-20 @ 11:08:11 -- Unexpected error: Cannot read properties of null (reading 'geozoneCount')

2025-04-20 @ 11:08:19 -- Unexpected error: Cannot read properties of undefined (reading 'getLonMap')

2025-04-20 @ 11:08:25 -- Unexpected error: Cannot read properties of null (reading 'geozoneCount')

2025-04-20 @ 11:08:30 -- Unexpected error: Cannot read properties of undefined (reading 'getLonMap')

2025-04-20 @ 11:08:34 -- Unexpected error: Cannot read properties of undefined (reading 'getLonMap')

2025-04-20 @ 11:08:38 -- Unexpected error: Cannot read properties of undefined (reading 'getLonMap')

2025-04-20 @ 11:08:40 -- Unexpected error: Cannot read properties of undefined (reading 'getLonMap')

2025-04-20 @ 11:08:45 -- Unexpected error: Cannot read properties of undefined (reading 'getLonMap')

2025-04-20 @ 11:09:16 -- Unexpected error: Cannot read properties of undefined (reading 'getMultiMissionIdx')

2025-04-20 @ 11:09:17 -- Unexpected error: Cannot read properties of undefined (reading 'getMultiMissionIdx')

It's also noticeable that adding a mission is erratic. Several mouse clicks are required to do things such as adding WPs, selecting WPs etc. All very sluggish and erratic.

I'm guessing this is related to Geozones.

Edit:
Checking again the missed mouse clicks are caused by:
[...]

PR #2371, opened 2025-04-25
Mission control WP options and Geozones fixes

breadoven:
Fix for https://github.com/iNavFlight/inav/issues/10821.

Also fixes https://github.com/iNavFlight/inav-configurator/issues/2369.

Additional fix added for WP Jump P1 value. This is initially set as the geo WP number to Jump to but when the Jump WP was selected again it showed the WP index which may be different to the geo WP number ... confusing. The Jump P1 field now shows the geo WP number when the Jump WP is selected (this isn't strictly correct as the P1 value will be different to the geo WP number but it makes more sense from a user point of view).

### DIFF
--- a/js/waypointCollection.js
+++ b/js/waypointCollection.js
@@ -320,6 +320,16 @@ let WaypointCollection = function () {
         return outputNumber;
     }
 
+    self.convertWaypointToJumpNumber = function(jumpId) {
+        let outputNumber = 0;
+        self.getNonAttachedList().forEach(function (element) {
+            if (element.getNumber() == jumpId) {
+                outputNumber = element.getLayerNumber();
+            }
+        });
+        return outputNumber;
+    }
+
     self.isJumpTargetAttached = function(waypoint) {
         let lJumptTargetAttached = [];
         data.forEach(function (element) {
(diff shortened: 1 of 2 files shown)


rts18:
Thx for the fix.
It no longer returns P1 to 0 from 1 after saving mission to eeprom.

However can I make a suggestion. It now defaults RTH P1 to 0. Requiring us to select 1 to enable landing after RTH.
In my reasoning, I'd think most people that select WP RTH would expect it to land, rather than hover above home.
How does it seem to you if the selection for RTH P1 action defaults to 1 as before?

breadoven:
> Thx for the fix. It no longer returns P1 to 0 from 1 after saving mission to eeprom.
> 
> However can I make a suggestion. It now defaults RTH P1 to 0. Requiring us to select 1 to enable landing after RTH. In my reasoning, I'd think most people that select WP RTH would expect it to land, rather than hover above home. How does it seem to you if the selection for RTH P1 action defaults to 1 as before?

Problem is that for fixed wing you probably wouldn't want it to land given FW landing is usually a lottery unless landing on a large open flat field. Not sure what the solution is here. Could make it platform specific I guess which will work so long as Configurator is connected to the FC.

breadoven:
Changed so this now defaults to RTH land for multirotors.

rts18:
> Changed so this now defaults to RTH land for multirotors.

Appreciate that. 👍

PR #2381, opened 2025-05-30
adding Ukrainian translations from crowdin

smsnvc:
Adding Ukrainian translations to keep Ukrainian version 100% translated


sensei-hacker:
Is there anyone who can independently take a look at this, double-check it?

bfmvsa:
Looks good for me. But I am not sure we need to translate 
"currentLanguage": {
        "message": "укр"
 },
 In other files we do not translate this. Probably we should add a comment there "Don't translate!"

sensei-hacker:
Yeah currentLanguage should be "uk".
Maybe I should have called that currentLanguageCode or similar.

smsnvc:
@bfmvsa , thanks a lot!

@sensei-hacker , I've changed currentLanguage to uk.

PR #2385, opened 2025-06-11
Replace Bing Maps with Esri World Imagery

Scavanger:
Bing Maps is out of service.

Esri World Imagery (+ World Boundaries and Transportation) is not quite as detailed but can be used free of charge and does not require an API key.


sensei-hacker:
Thank you so much, Scavanger!

Linux zip available for testing here:
https://github.com/iNavFlight/inav-configurator/actions/runs/15574766724/artifacts/3301919165

b14ckyy:
We should make a 8.0.2 configurator for that soon don't you think?

MrD-RC:
I'm thinking 8.1 and add the change Breadoven has recently made a PR through, to allow the old altitude estimator use. But, it should be well tested.

stronnag (approved):
Appears to address everything, including elevation API 
+1

Jetrell:
The only thing I could mention. Is that it shows _Map data not yet available_, when zoomed beyond what the imagery provides.
Not a deal breaker by any means. But it may be raised by some after the release, as being a problem.

>I'm thinking 8.1 and add the change Breadoven has recently made a PR through, to allow the old altitude estimator use. But, it should be well tested.

@MrD-RC I have yesterdays build loaded and ready. And intend to test and tune it on a couple of the planes I had troubles with. But the weather down here is miserable. So I'm looking for a day that I can spend a good amount of time trying different adjustments.

stronnag:
This is an issue with the data set. If you live in Western Europe or the USA, the data is available at all advertised scales.
If you live the the Australian outback or other less populated area, the data is not available.

INAV cannot fix this. We might mention it in the release note.

If this bothers you, can add a more comprehesive source via the "Map Proxy"  option. 

You may know this place ....

## MapBox

image

## Google

image


## Bing (RIP)

image


## Esri (alas)

image

b14ckyy:
yeah that makes sense. In my area where the max level is available, I can zoom in to a full screen single pixel as the map drawer just does not load more and zooms what it has. 
But when I go into some outback or desert, then the server delivers a "no map data" time instead. Configurator cannot know that this is no actual map. 

But overall looks good to me and I think we can merge.

Jetrell:
@sensei-hacker @stronnag That makes sense.

>If this bothers you, you can add a more comprehensive source via the "Map Proxy" option.

My answer to that earlier was - _Not a deal breaker by any means.

PR #2476, opened 2025-12-14
fix serial port open

error414:
**This PR needs review by someone who knows JS well.** 

I noticed that opening serial port is slow. The reasson is that serial port is not opened and configurator tries to write data serial port. It causes restart MSP quering and delay for opening configurator. 

The flow is:
```
Connecting to: COM5
connection.js:88  Connection opened with ID: 1, Baud: 115200

connectionSerial.js:121  Serial write error: Invalid serial port or port closed <-- IT'S NOT CORRECT
serial_queue.js:285      Force free hard lock
serial_queue.js:146      MSP data request timed-out: 1

intervals.js:114  Killing all intervals except:
  - global_data_refresh
  - msp-load-update
  - ltm-connection-check

fc.js:205  Generating AUX_CONFIG
```

I would say it's caused by calling **resolve** before port is fully opened, right aproach call resolve when port is fully opened. See pull request code. 

After my patch is connection to FC much faster and log look more clean
 ```
Connecting to: COM5
connection.js:88  Connection opened with ID: 1, Baud: 115200

intervals.js:114  Killing all intervals except:
  - global_data_refresh
  - msp-load-update
  - ltm-connection-check

fc.js:205  Generating AUX_CONFIG

intervals.js:114  Killing all intervals except:
  - global_data_refresh
  - msp-load-update
  - ltm-connection-check
  ```

### DIFF
--- a/js/main/serial.js
+++ b/js/main/serial.js
@@ -12,10 +12,16 @@ const serial = {
     connect: function(path, options, window) {
         return new Promise(resolve => {
             try {
+                var openPortResolved = false;
                 this._serialport = new SerialPortStream({binding, path: path, baudRate: options.bitrate, autoOpen: true});
                 this._serialport.on('error', error => {
                     if (!window.isDestroyed()) {
                         window.webContents.send('serialError', error); 
+                    }
+
+                    if(!openPortResolved) {
+                        openPortResolved = true;
+                        resolve({error: false, id: this._id++});
                     }
                 });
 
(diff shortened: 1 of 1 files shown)


sensei-hacker:
/ask qodo, do you have comments on the above discussion?

sensei-hacker:
I've found qodo is right about half the time, so the comments from the bot shouldn't be accepted at face value, but are things to consider.

PR #2512, opened 2025-12-31
Implement robust WebGL fallback for 3D rendering

sensei-hacker:
## Summary

Adds automatic fallback from hardware to software WebGL rendering with graceful degradation when WebGL is completely unavailable.

## Problem

The configurator had WebGL detection code but crashed when WebGL was unavailable:
- When `renderer.getContext('webgl')` failed, `renderer` stayed `undefined`
- Code then tried to call `renderer.setSize()` → **crash: Cannot read property 'setSize' of undefined**
- The existing "fallback" model was never used because the app crashed before loading it

## Solution

Implemented a 3-tier progressive fallback:

1. **Tier 1**: Try hardware-accelerated WebGL (best performance)
2. **Tier 2**: Try software-rendered WebGL with `failIfMajorPerformanceCaveat: false` (slower but compatible)
3. **Tier 3**: Show user-friendly message and provide no-op functions (graceful degradation)

## Changes

- Add `tryCreateWebGLContext()` helper with 3-tier fallback
- Try hardware WebGL, then software WebGL, then graceful disable
- Show user-friendly error messages when 3D unavailable with actionable troubleshooting steps
- Provide no-op `render3D()` and `resize3D()` functions to prevent crashes
- Add console logging for WebGL rendering method detection

## Testing Completed

**Linux testing:**
- ✅ Normal operation (hardware WebGL available) - PASSED
- ✅ WebGL completely unavailable (forced fail) - PASSED, shows fallback message

**WIndows testing**
user tested with older Windows system that doesn't support WebGL

## Benefits

- Prevents crashes when WebGL unavailable in Setup and Magnetometer tabs
- Improves UX for users with GPU driver issues
- Better support for VM environments and Remote Desktop
- Clear guidance for users on how to resolve issues

## Files Modified

- `tabs/setup.js`: Robust WebGL detection for 3D aircraft model
- `tabs/magnetometer.js`: Robust WebGL detection for compass visualization

### DIFF
--- a/tabs/setup.js
+++ b/tabs/setup.js
@@ -225,14 +225,84 @@ TABS.setup.initialize3D = function () {
     canvas = $('.model-and-info #canvas');
     wrapper = $('.model-and-info #canvas_wrapper');
 
-    // webgl capability detector
-    // it would seem the webgl "enabling" through advanced settings will be ignored in the future
-    // and webgl will be supported if gpu supports it by default (canary 40.0.2175.0), keep an eye on this one
-    var detector_canvas = document.createElement('canvas');
-    if (window.WebGLRenderingContext && (detector_canvas.getContext('webgl') || detector_canvas.getContext('experimental-webgl'))) {
-        renderer = new THREE.WebGLRenderer({canvas: canvas.get(0), alpha: true, antialias: true});
-        useWebGlRenderer = true;
+    // Robust WebGL capability detection with fallback
+    function tryCreateWebGLContext() {
+        if (!window.WebGLRenderingContext) {
+            return null;
+        }
+
+        const detector_canvas = document.createElement('canvas');
+        let gl = null;
+        let renderMethod = null;
+
+        // Try 1: Hardware-accelerated WebGL (best performance)
+        try {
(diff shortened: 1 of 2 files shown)


tabs/setup.js
```diff
@@ -225,14 +225,84 @@ TABS.setup.initialize3D = function () {
...
     wrapper = $('.model-and-info #canvas_wrapper');
 
-    // webgl capability detector
-    // it would seem the webgl "enabling" through advanced settings will be ignored in the future
-    // and webgl will be supported if gpu supports it by default (canary 40.0.2175.0), keep an eye on this one
-    var detector_canvas = document.createElement('canvas');
-    if (window.WebGLRenderingContext && (detector_canvas.getContext('webgl') || detector_canvas.getContext('experimental-webgl'))) {
-        renderer = new THREE.WebGLRenderer({canvas: canvas.get(0), alpha: true, antialias: true});
-        useWebGlRenderer = true;
+    // Robust WebGL capability detection with fallback
```
Scavanger: This is the same code as in magnetometer.js.
This code should only appear once in another file (GUI.js would be a candidate).
Edit:
 I mean, of course, the tryCreateWebGLContext() function.

PR #2549, opened 2026-01-31
Fix mission file save error caused by fs module usage

sensei-hacker:
## Summary
Fixes the "fs is not defined" ReferenceError that occurs when trying to save mission files from the Mission Control tab.

## Problem
Users encountered a JavaScript error when attempting to save mission files:
```
mission_control.js:4050 Uncaught (in promise) ReferenceError: fs is not defined
```

The legacy code was using Node.js `fs.writeFile()` directly from the renderer process, but the `fs` module is not accessible in Electron's renderer process due to context isolation.

## Changes
- Replaced `fs.writeFile()` with `window.electronAPI.writeFile()` 
- Updated from callback pattern to Promise-based API (`.then()`)
- Moved success messages inside `.then()` block to ensure they execute after file write completes

## Testing
- [x] Mission file save dialog opens correctly  
- [x] Files save successfully to selected location
- [x] No console errors during save operation
- [x] Tested with live configurator (yarn start)

This follows the same pattern already used in `cli.js` and other tabs that were previously updated.

### DIFF
--- a/tabs/mission_control.js
+++ b/tabs/mission_control.js
@@ -4057,15 +4057,15 @@ function iconKey(filename) {
         var builder = new xml2js.Builder({ 'rootName': 'mission', 'renderOpts': { 'pretty': true, 'indent': '\t', 'newline': '\n' } });
         var xml = builder.buildObject(data);
         xml = xml.replace(/missionitem mission/g, 'meta mission');
-        fs.writeFile(filename, xml, (err) => {
+        window.electronAPI.writeFile(filename, xml).then(err => {
             if (err) {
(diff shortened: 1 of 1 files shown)


Jetrell:
I tested this with Windows. When the 'save eeprom mission' button is pressed, the tab will go blank and then reappear with the mission WP's gone.
But if I load it, then save it again. It will then save fine. And every time after that.

I also noticed that other 9.0 builds before this fix will also work once a mission is saved via this build. i.e. before the PC is shutdown and cache is dumped.

Issue #2552, opened 2026-02-07
9.0 Custom OSD Visablity only shows LC 0


Jetrell:
When choosing to display a custom OSD message.  It can be displayed as the result of a logic condition being met.
When ever that field is entered and saved, and you leave the OSD tab and return. All the fields show Logic 0. Although they are actually saved with the required condition, shown in a CLI under `# OSD: custom elements`.

<img width="369" height="365" alt="Image" src="https://github.com/user-attachments/assets/20478ad7-4db5-459b-b625-a11f92f54d77" />

error414:
could you share you dump (settings)?

Jetrell:
If any more than 20 logic conditions are entered into the programming framework, regardless of whether they are in a numerical sequence from 1 to 20. Or just scattered throughout the 63 lines... This causes all the Custom OSD _visibility_ logic drivers to become LC 0 in place of the conditions they where given.

The bug appears to have turned up after the RC2 configurator release.

error414:
could you provide config?

Jetrell:
@error414 There is nothing special about the diff setting file. I tested this on multiple boards. Other than it has 3 custom text fields that use most of the 16 characters.
This diff has 20 logic line active with no operands or functions being checked... If I activate a 21st. That is when the visibility logic driver become LC 0.

Another odd occurrence is the `mah drawn` element in the OSD tab. When ever a custom OSD text letter or number is added to the text filed. The numerical value of `mah drawn` changes in preview. 

[INAV_9.0.0_cli.txt](https://github.com/user-attachments/files/25170492/INAV_9.0.0_cli.txt)

MrD-RC:
The mAh drawn value is a random number to the correct length.

error414:
@Jetrell I did't think something wrong is in your config, just it's much faster to set my FC and reproduce the bug :). Thanks for the dump, I will look at that.

error414:
It's fixed, https://github.com/iNavFlight/inav-configurator/pull/2558 , Could you test it please?

PR #2558, opened 2026-02-09
Fix filling visibility LC of OSD custom elements

error414:
It's fix for https://github.com/iNavFlight/inav-configurator/issues/2552


Jetrell:
Thanks Petr. I can confirm it now works correctly again.

sensei-hacker:
Thanks for that. We now have an MSP message that tells us which LCs are configured, with just one small MSP message.
loadLogicConditions() can be up to 64 MSP round trips.

https://github.com/iNavFlight/inav-configurator/pull/2559

error414:
so we can close this PR, because core of issue was calling getLCOptions from wrong place 

https://github.com/iNavFlight/inav-configurator/pull/2559/changes#diff-cd81907a0c9960d9561f9ca8a2eb5c95069f69456485620e79877ce2576043f7L3561

PR #2559, opened 2026-02-09
OSD: use LC configured mask

### LINKED ISSUE #2552: 9.0 Custom OSD Visablity only shows LC 0

sensei-hacker:
## Summary
- The OSD tab only needs to know *which* Logic Conditions are enabled (for dropdown selectors) — it never inspects LC content
- Replaces the heavyweight `loadLogicConditions()` call (which fetches up to 64 individual LCs via MSP) with a single `MSP2_INAV_LOGIC_CONDITIONS_CONFIGURED` request that returns an 8-byte bitmask
- Moves the fetch into the `OSD.reload` callback chain so the mask is available before `createCustomElements()` builds the UI

This is an alternative fix for #2552 — same root cause (LC data not loaded when custom element dropdowns are built), but instead of loading all 64 LCs, we only fetch the configured mask since that's all the OSD tab needs.

## Testing
Tested on real hardware (AOCODARCF4V3_SD running INAV 9.0.0):
1. Enabled LC 4 and LC 6 in Programming tab, saved to EEPROM
2. Navigated to OSD tab — LC dropdowns correctly show only LC 4 and LC 6
3. Set custom element 0 visibility to "Logic Condition" -> LC 6, saved
4. Left OSD tab (Status), returned to OSD tab
5. Verified: dropdown shows LC 6 (not LC 0) — bug #2552 is fixed
6. Verified FC data model: {type: 2, value: 6} matches UI
7. Zero console errors throughout

Also tested with no LCs configured — dropdowns are correctly empty.

## Code Review
Reviewed with inav-code-review agent — no critical or important issues found.

Fixes #2552

### DIFF
--- a/tabs/osd.js
+++ b/tabs/osd.js
@@ -2435,10 +2435,12 @@ OSD.reload = function(callback) {
                     OSD.data.supported = true;
                     OSD.msp.decodePreferences(resp);
                     
-                    MSP.promise(MSPCodes.MSP2_INAV_CUSTOM_OSD_ELEMENTS).then(() => { 
+                    MSP.promise(MSPCodes.MSP2_INAV_CUSTOM_OSD_ELEMENTS).then(() => {
                         mspHelper.loadOsdCustomElements(() => {
-                            createCustomElements();
-                            done();
+                            MSP.send_message(MSPCodes.MSP2_INAV_LOGIC_CONDITIONS_CONFIGURED, false, false, function() {
+                                createCustomElements();
+                                done();
+                            });
                         });
                     });
                 });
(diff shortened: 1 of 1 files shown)


Jetrell:
I've been using a build with this in it. So far I haven't seen any issues.
In fact, I started using 9.0.1 configurator and noticed this hadn't been added. So back to this build I went.

9.0.2 maybe ?

PR #2560, opened 2026-02-10
OSD: Make custom elements UI more intuitive

sensei-hacker:
Old: Four mystery boxes, each a flat list of 64 combinations of sources and formats:

image

image

Also with the old style you had to do the custom element in two places - setting it up on the right doesn't mean it'll appear.

New: six sources to choose from. Then separately, twelve formatting options:

image

While we're at it, instead of typing the number of an icon, 1 -256, this may be 784% better?

image

### DIFF
--- a/src/css/tabs/osd.css
+++ b/src/css/tabs/osd.css
@@ -822,6 +822,100 @@
     width: 100% !important;
 }
 
+/* Icon picker popup */
+.ce-icon-picker-overlay {
+    position: fixed;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    background: rgba(0, 0, 0, 0.5);
+    z-index: 10000;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
+
+.ce-icon-picker-popup {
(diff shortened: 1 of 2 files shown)


Jetrell:
I fear we may have a few issues with this.
* I tested this on a smaller laptop screen (1366 x 768). And with a HD OSD.  Which has a wider preview. 
As can be seen. The boxes overhang the columns.
This was also an issue that required fixing in past releases, on the current implementation as well.

With HD and small screen


* The next issue is the "Custom element decimal index". And it's "driver LC condition selector'" ( Not the active state condition ! )

The image below shows the current method. 
As can be seen. It contains the "Custom element decimal index" at the top in green. And the "LC condition driver selector'" in blue at the bottom.

Current UI

While the new one has the 'Custom element decimal index'.  But I doesn't show the LC  driver selector drop-down.


New UI

But with PAL, it does show up.
Pal screen

sensei-hacker:
Thanks Jetrell!

https://github.com/iNavFlight/inav-configurator/pull/2636

PR #2569, opened 2026-02-18
* add ADSB areas to GPS map

error414:
#PR related to https://github.com/iNavFlight/inav/pull/11346

New:
1) new table for adsb contacts
2) marked plane warning or alert is coloured 

image


sensei-hacker:
Resolved the merge conflict with \`maintenance-10.x\` and pushed directly to this branch (thanks for having maintainer edit enabled).

**What was conflicting:** \`tabs/gps.js\` had one conflict zone — the tab-modules refactor (PR #2458) renamed \`TABS.gps\` → \`gpsTab\` in the same area where this PR improved the ADSB tooltip content with i18n keys.

**Resolution:** Applied the \`gpsTab\` rename to the PR's improved tooltip content so both sets of changes are preserved. All other files auto-merged cleanly.

This branch should now be conflict-free against \`maintenance-10.x\`.

sensei-hacker:
Looking at the Sonarqube items, the following might be worth considering. I haven't analyzed it closely because you know the code better than I do, but perhaps worth looking at:

  Issue 4 — Stale warning ICAO when vehicles disappear:

  When vehiclesCount drops to 0, get_adsb_warning() skips the MSP fetch and calls update_adsb_ui() directly. But update_adsb_ui() gates on vehiclePacketCount > 
  0, which is cumulative and never resets. Result: the warning ICAO row stays visible showing the last aircraft's ICAO even after it's gone. A user would see a
  persistent "warning" with a stale ICAO number when sky is clear.


ADSB_VEHICLE_TYPE[emitterType] unguarded: LOW RISK

  The table covers 0–19; ADSB types 20–23 exist in the spec but are reserved/unassigned. An out-of-range value would throw TypeError inside the vehicle
  rendering loop, breaking that map update cycle. Practically, real aircraft don't broadcast types 20–23, but it's still a crash-on-bad-data issue worth a
  one-line guard.

error414:
Thank you very much for checking.  Issue 4 is valid. I will look at that. 

I will fix "ADSB_VEHICLE_TYPE[emitterType] unguarded: LOW RISK" as well, it would be nice to add guard there.

PR #2573, opened 2026-02-25
Add bulgarian translation

Pechenikov1:
This PR adds full Bulgarian language support.

Created locale/bg.json with all interface strings translated.

Registered the Bulgarian language in the language selector.

The translation has been tested locally and is working correctly.


sensei-hacker (changes requested):
Thanks for submitting this!


It appears you ran `yarn install` locally in a newer environment, which updated package versions. Please remove `yarn.lock` from the commit.  Also, you can't .gitignore a tracked file. One would need to either a) just don;t commit changes to the file or perhaps b) .git/info/exclude


**Action required:** Please rebase/reset so that `yarn.lock` and .gitignore are not modified in this PR

Pechenikov1:
"I have reverted yarn.lock changes as requested. The PR now contains only the translation files and language registration. Ready for review."

PR #2595, opened 2026-03-19
Allow motor test mode when reversible motors (3D) enabled

sensei-hacker:
Needs firmware support for reverse to work

## Summary

Fixes the Outputs tab motor test checkbox being disabled when the REVERSIBLE_MOTORS feature (3D mode / bidirectional ESCs) is enabled.

## Problem

`feature3DSupported` was initialized to `false` and never set to `true` anywhere. This caused the guard `if (self.feature3DEnabled && !self.feature3DSupported)` to always block test mode for 3D users. The variable was a leftover from a removed API version check (commit ef831a3 from 2016).

Additionally, a typo (`FC.EVERSIBLE_MOTORS` instead of `FC.REVERSIBLE_MOTORS`) in the neutral value sanity check would have caused a runtime TypeError if the code path had been reachable.

## Changes

- Removed `feature3DSupported` variable and the guard that blocked test mode
- Simplified `feature3DEnabled && feature3DSupported` checks to just `feature3DEnabled`
- Fixed typo `FC.EVERSIBLE_MOTORS` → `FC.REVERSIBLE_MOTORS`
- Improved comment on neutral value sanity check bounds
- Slider initialize to center when reversible mode is on

## Testing

- Verified motor test checkbox enables correctly with REVERSIBLE_MOTORS feature active
- Sliders initialize to neutral point (1500) in 3D mode
- Motor values display raw PWM in 3D mode
- Normal (non-3D) motor testing unaffected

### DIFF
--- a/tabs/outputs.js
+++ b/tabs/outputs.js
@@ -16,14 +16,12 @@ import interval from './../js/intervals';
 
 TABS.outputs = {
     allowTestMode: false,
-    feature3DEnabled: false,
-    feature3DSupported: false
+    feature3DEnabled: false
 };
 TABS.outputs.initialize = function (callback) {
     var self = this;
 
     self.armed = false;
-    self.feature3DSupported = false;
     self.allowTestMode = true;
 
     var $motorsEnableTestMode;
@@ -402,10 +400,6 @@ TABS.outputs.initialize = function (callback) {
     function process_motors() {
         $motorsEnableTestMode = $('#motorsEnableTestMode');
 
-        if (self.feature3DEnabled && !self.feature3DSupported) {
-            self.allowTestMode = false;
-        }
-
         $motorsEnableTestMode.prop('checked', false);
         $motorsEnableTestMode.prop('disabled', true);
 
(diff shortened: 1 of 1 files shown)


Pikkuboo:
tested on a speedybee f7 v3 fc and motor testing now works when reversible motor is on, but only forwards.

sensei-hacker:
> tested on a speedybee f7 v3 fc and motor testing now works when reversible motor is on, but only forwards.

@Pikkuboo  that should be fixed by https://github.com/iNavFlight/inav/pull/11847 
I don't suppose you could test it?

Raffi1202:
Tested the configurator side of this PR (head 660060dc) against the bundled 9.1 SITL, driven over CDP:

- Quad X mixer applied, `REVERSIBLE_MOTORS` enabled in the Outputs tab, saved and rebooted.
- After the reconnect the motor test checkbox is enabled (`disabled = false`), the four sliders sit at the FC's neutral value (1460 from `3d_neutral`) with range 1000-2000 and unlock when test mode is switched on.
- Dragging motor 1 to 1400, 1600 and back to 1500: each value goes out in `MSP_SET_MOTOR` (bytes 0-1 = 1400 / 1600 / 1500) and `MSP_MOTOR` reads the same value back; the value label under the slider shows the raw number as expected in 3D mode.
- No renderer exceptions from the tab (only the unrelated dev-mode `appGetVersion` one on maintenance-9.x).

So the checkbox/slider fix behaves as described. Whether the ESC actually spins backwards below neutral is firmware territory (#11847), which a SITL cannot show.

PR #2599, opened 2026-03-26
feat: Add Map Generator tab for offline map tile downloading

MartinovEm:
Add a new Map Generator tab that allows downloading offline map tiles for OSD map widgets on radios. Works without a flight controller connection.

Hero

Supported output targets: b14ckyy ETHOS Mapping Widget, Yaapu Telemetry (ETHOS and EdgeTX sub-targets). Map providers: OpenStreetMap, ESRI, Google (Street/Satellite/Hybrid/Terrain). Features: draw rectangle to select download area with live measurements, configurable zoom level range (1-20), sync tiles directly to mounted SD card, export tiles as ZIP archive, eject SD card from the app (Windows/macOS/Linux), local tile cache for faster repeated downloads, save/restore settings persistence, geocoder search, measure tool, right-click context menu.

Drawing

Sync


MartinovEm:
## Update: Added INAV Terrain (SRTM) target

The Map Generator now supports generating INAV-compatible `.DAT` terrain files directly from the configurator — no external tools needed.

### What it does

Select "INAV Terrain (SRTM)" as the target, draw a rectangle over your flying area, and the configurator downloads NASA SRTM 30m elevation data, converts it to `.DAT` files, and writes them directly to the flight controller's SD card (or exports as ZIP).

### How it works

1. Downloads SRTM1 HGT tiles from AWS (public domain NASA data)
2. Decompresses gzip, performs bilinear interpolation at each grid point
3. Packs into 2048-byte blocks with CRC16-XMODEM checksums
4. Outputs one `.DAT` file per 1°×1° degree square (INAV-compatible format)

### Features

- **1°×1° terrain grid overlay** with `.DAT` filename labels on the map
- **IndexedDB HGT cache** — downloaded SRTM tiles persist across sessions, no re-download on repeat generation
- **Live altitude readout** — hover the mouse to see SRTM elevation after generation (from cached data, no server requests)
- **SD card sync** with FREESPAC.E auto-deletion, write progress with speed/ETA
- **ZIP export** when no SD card is linked
- **SRTM coverage warning** for selections beyond 60°N / 56°S
- **Completion screen** shows the CLI command to enable terrain (`set terrain_enabled = ON`)
- Ocean depths clamped to 0m (matches INAV firmware behavior)

### Validation

The terrain math has been validated against 133M+ data points across 13 tiles (flat terrain to 2900m mountains). Flight-tested on real hardware (SpeedyBee F405 Wing, INAV 9.0.1) — sub-meter mean error vs source data.

### Screenshots

**Terrain mode — grid overlay and area selection:**

selecting_area


**Generating terrain files — sync modal with progress:**

generating_tiles


**Sync complete — files written to SD card:**

terrain_synced

MartinovEm:
**Fix: OSM tiles returning 403 in packaged builds (commit 97e9739c)**

While working on the Map Generator tab, I noticed that the packaged/release builds of the configurator get 403 errors from OSM tile servers (affects GPS tab, Mission Control — any tab that loads OSM tiles). Dev mode works fine.

**Root cause:** The packaged build loads the app using the local file protocol (`mainWindow.loadFile(...)`), so tile requests carry an invalid or empty `Referer` header and Chromium's generic `User-Agent`. OSM rejects these per their [Tile Usage Policy](https://operations.osmfoundation.org/policies/tiles/) (sections 3.1 and 3.4), which requires:
- A unique `User-Agent` identifying the application (not a library/browser default)
- A valid `Referer` header from web contexts

In dev mode, Vite serves via `http://localhost`, giving a valid Referer — so the issue only appears in packaged builds.

**Fix:** Added `session.webRequest.onBeforeSendHeaders` in `js/main/main.js` to set a proper `Referer` and `User-Agent` for OSM tile requests. 6 lines, no other code affected. Tested in a packaged win64 build — zero 403s.

This brings the configurator into compliance with OSM's tile usage requirements.

sensei-hacker:
> **Fix: OSM tiles returning 403 in packaged builds (commit [97e9739](https://github.com/iNavFlight/inav-configurator/commit/97e9739ceb6a3e35de761d4812a475bd3a3161d7))**
> 
> While working on the Map Generator tab, I noticed that the packaged/release builds of the configurator get 403 errors from OSM tile servers (affects GPS tab, Mission Control — any tab that loads OSM tiles). Dev mode works fine.
> 
> **Root cause:** The packaged build loads the app using the local file protocol (`mainWindow.loadFile(...)`), so tile requests carry an invalid or empty `Referer` header and Chromium's generic `User-Agent`. OSM rejects these per their [Tile Usage Policy](https://operations.osmfoundation.org/policies/tiles/) (sections 3.1 and 3.4), which requires:
> 
> * A unique `User-Agent` identifying the application (not a library/browser default)
> * A valid `Referer` header from web contexts
> 
> In dev mode, Vite serves via `http://localhost`, giving a valid Referer — so the issue only appears in packaged builds.
> 
> **Fix:** Added `session.webRequest.onBeforeSendHeaders` in `js/main/main.js` to set a proper `Referer` and `User-Agent` for OSM tile requests. 6 lines, no other code affected. Tested in a packaged win64 build — zero 403s.
> 
> This brings the configurator into compliance with OSM's tile usage requirements.

That's really interesting. I have confirmed it does set the User-Agent as required -- but I checked in dev mode.
They say it doesn't have to set a referer since it's not a web site, but it doesn't hurt to send one.

Speaking of their TOS, I see also:
```
You must not:

Bulk download (“scrape”) tiles or offer prefetch features.
```

Do you have thoughts on that?  I think the spirit of it is they don't want people downloading tiles they don't end up using, to reduce wasteful load on their servers.  Would this feature, if merged, potentially cause that?

MartinovEm:
@sensei-hacker Thanks for pointing this out!

I agree with your interpretation of the OSM Standard tile policy regarding offline/bulk download. Before I change behavior in this PR, I want to confirm the intended direction with you:

1. Keep OSM available for normal interactive map viewing, so the Map Generator tab follows the rest of the Configurator tabs, where OSM is used for previews and quick orientation.
2. Restrict offline export flows (Sync to SD / ZIP) when OSM is selected.
3. Show a clear in-app message explaining that OSM Standard does not permit offline bulk export, and ask users to choose another provider for export.

If you agree with this direction, I will implement it in this PR.

sensei-hacker:
> @sensei-hacker Thanks for pointing this out!
> 
> I agree with your interpretation of the OSM Standard tile policy regarding offline/bulk download. Before I change behavior in this PR, I want to confirm the intended direction with you:
> 
> 1. Keep OSM available for normal interactive map viewing, so the Map Generator tab follows the rest of the Configurator tabs, where OSM is used for previews and quick orientation.
> 2. Restrict offline export flows (Sync to SD / ZIP) when OSM is selected.
> 3. Show a clear in-app message explaining that OSM Standard does not permit offline bulk export, and ask users to choose another provider for export.
> 
> If you agree with this direction, I will implement it in this PR.

Seems reasonable to me. I see that for example MapTiler offers up to 5,000 sessions per month with their free tier, using the same API.  Would it make sense to have a link to a documentation page that has a couple such providers?
perhaps even add a commercial provider or two (such as Maptiler) directly int he drop-down? These would be providers that are free for any reasonable number of requests.

MartinovEm:
@sensei-hacker Thank you for the suggestions! I've implemented the first part:

- **OSM is now preview-only** -- the Sync to SD and Export as ZIP buttons are disabled when OSM Standard is selected, with an inline notice directing users to choose ESRI or Google for offline export. OSM still works for interactive map viewing, consistent with how it's used in other configurator tabs.

Regarding MapTiler: I'm starting work on adding it as a provider directly in the dropdown. Since it requires a personal API key, the UI will include a key input field (shown only when MapTiler is selected) with a short inline message guiding users to get a free key at maptiler.com -- no separate documentation page needed.

sensei-hacker:
> Since it requires a personal API key, the UI will include a key input field (shown only when MapTiler is selected) with a short inline message guiding users to get a free key at maptiler.com -- no separate documentation page needed.

That will probably be more convenient than the existing API key field that is under the gear icon at top right of Configurator, won't it.

I guess it can be the same setting under the hood, but just put the input field right where people need it?

MartinovEm:
@sensei-hacker Thanks for the suggestion! Just to clarify — there isn't an existing map API key field under the gear icon. The gear settings only have the map provider for the Mission/GPS tabs (OSM / ESRI / MapProxy with URL + layer), which is separate from the Map Generator. So the key is new and lives right in the Map Generator tab where it's most useful, stored persistently in `electron-store` (`mapgen_maptiler_key`) so users only need to enter it once.

Here's a summary of what's been added:

**New provider: MapTiler** with 6 map types:

| Type | Notes |
|---|---|
| Satellite | Aerial imagery (JPEG) |
| Hybrid | Satellite + labels (JPEG) |
| Street | MapTiler streets style (PNG) |
| OSM Style | OpenStreetMap-rendered tiles — a compliant alternative for bulk export |
| Outdoor | Topographic + outdoor features (PNG) |
| Topo | Clean topo lines (PNG) |

The key input shows only when MapTiler is selected, with an inline "Get a free key at maptiler.com" link. A help icon next to the Provider dropdown links directly to the README setup section. MapTiler's free tier covers 5,000 sessions/month which is plenty for personal use.

MapTiler_Key


Full setup guide with registration steps: [README → MapTiler Setup](https://github.com/MartinovEm/inav-configurator/blob/feature/map-generator/README.md#maptiler-setup)

MartinovEm:
Rebased onto `maintenance-10.x` (was 9.x). The diff is now clean and feature-only (24 files), with the commit history preserved. INAV Terrain is now the default target.

All six platform builds pass on my fork: https://github.com/MartinovEm/inav-configurator/actions/runs/28181247774

MartinovEm:
@sensei-hacker

The latest push resolves the Reliability finding (added an `aria-label` to the MapTiler key input) and 2 of the Security findings (hoisted the drive-letter sanitizer in `ejectDrive` so the macOS/Linux `execFile` calls get the sanitized value too).

The remaining 3 `S2083` "path from user-controlled data" findings are on the `writeFile` / `rm` / `pathExists` IPC handlers. These are false positives in the Electron desktop context: the "user-controlled" path is the user's own chosen SD-card folder / save-dialog destination — writing tiles there is the feature's entire purpose, so the path can't be constrained to a fixed root. `writeFile` and `rm` are also pre-existing handlers already on `maintenance-10.x` (and shared by the backup/SITL features). Could you review and mark these as Safe? I'll adjust it if you'd prefer a different approach.

MartinovEm:
@sensei-hacker Pushed two small Map Generator fixes (Yaapu tiles now JPEG; terrain Sync/ZIP buttons work on cold start). All 7 builds are green. The SonarCloud failure is unchanged and expected — these fixes don't touch the flagged code, so it's still just the same 3 `S2083` path findings on the `writeFile`/`rm`/`pathExists` handlers, which only need to be manually marked **Safe**. Whatever you think is the correct way to handle them, I'll go with it.

MartinovEm:
## Terrain: Copernicus GLO-30 is now the default elevation source (SRTM kept as fallback)

The Terrain target can now build `.DAT` files from **Copernicus GLO-30** instead of SRTM1. SRTM stays selectable, and the `.DAT` format is unchanged — only the height values differ.

### Why

I validated four public 30 m DEMs against **ICESat-2 ATL08** laser ground returns (bare earth, orthometric EGM2008, cloud-free / snow-free / uncertainty < 10 m).

**Pooled study** — n = 10,777 paired points, block-bootstrap 95% CI, Wilcoxon signed-rank:

| Source | MAE vs ICESat-2 | 95% CI |
|---|---|---|
| **Copernicus GLO-30** | **1.07 m** | [0.89, 1.31] |
| Re:Earth (Copernicus-derived) | 1.10 m | [0.91, 1.34] |
| SRTM / Mapzen *(current)* | 2.01 m | [1.68, 2.53] |
| JAXA ALOS AW3D30 | 3.06 m | [2.77, 3.41] |

Copernicus vs the current SRTM path: **ΔMAE −0.94 m [−1.23, −0.75], p < 0.001**.

stageB_stats_forest


**Denser single-tile check** — same product and same QA filters, but seven summers (2019–2025) of ground tracks over one tile, n = 16,962: Copernicus **1.08 m**, Re:Earth 1.13 m, SRTM **2.29 m**, JAXA 3.68 m. Copernicus is unchanged with 6× more truth data and the ranking holds. SRTM also carries a **+1.77 m high bias** against the laser, versus +0.47 m for Copernicus.

stageB_errormap_N42E023


Spatially, how each candidate differs from the SRTM baseline shipped today (source rasters, identical sampling grid):

stageA_quad_N42E023


Copernicus and Re:Earth are nearly identical, as expected — Re:Earth is Copernicus-derived. JAXA is visibly noisier and biased high.

Two honest caveats: at n ≈ 11k almost any real difference is "significant", so the number that matters is the effect size in metres, not the p-value — all four are sub-3 m DEMs. And the DEMs are fixed-epoch (SRTM 2000, Copernicus 2010–2014) while the laser data is 2019–2025, so real terrain change adds noise — but equally to all four, so the comparison stays fair. The practical wins are bare-earth accuracy, **coverage to 84°N** (SRTM stops at 60°N), and fresher data.

### How it works

Copernicus ships Cloud-Optimized GeoTIFF, so `geotiff` (MIT) is added as a dependency. Its 3600×3600 pixel-is-point grid sits on the same whole-arc-second graticule as SRTM `.hgt`, so filling the 3601×3601 grid the converter already expects is a **direct pixel copy — no resampling, no half-pixel shift**. The south and east edge lines come from the neighbouring tiles; the N/E/NE neighbours are fetched as thin strips to cover the block grid's overshoot past the 1° square.

**The `.DAT` encoder is untouched** — block layout, CRC-16/XMODEM, geodesy constants and filenames are identical.

- Copernicus and SRTM grids are cached under separate keys, so switching source can't mix them.
- If the Copernicus service is unreachable, the user is **asked** before falling back to SRTM — never switched silently.
- A missing tile (ocean / outside coverage) is treated as no data, not an error.

copernicus_inav_cinfig


### Licence

Copernicus WorldDEM-30 permits redistribution of derived products, including commercially, conditional on two verbatim notices. These are shown in the tab and written into the exported ZIP as `NOTICE.txt` (the licence ties them to distribution, so a direct write to the user's own SD card doesn't carry the file). No endorsement by the EU, ESA, Airbus or DLR is implied.

### Verification

On `N42E023` (15,500 blocks / 31,744,000 bytes — byte-layout identical to the SRTM output):

- CRC-16 valid on **all 15,500 blocks**; magic, version and spacing correct
- sampled heights match the source GeoTIFF exactly, including the south edge (cross-checked against the neighbouring tile)
- **no edge holes** — 0 zero-heights out of 13,888,000
- Copernicus vs SRTM **`.DAT` output**: mean |Δ| 2.72 m
- **tested on a real flight controller**: the files load, and flying from a Copernicus tile into a neighbouring tile generated earlier from SRTM is seamless

PR #2601, opened 2026-04-01
fixed usb descriptor len mismatch on some chips

Yuri-Sharapov:
Some DFU implementations (e.g. STM32 ROM bootloader on certain devices like stm32f427)
may report incorrect bLength in USB string descriptors.

Clamp descriptor length to actual buffer size to avoid out-of-bounds reads.


sensei-hacker:
Thanks, Yuri!

PR #2615, opened 2026-05-12
fix(cli): replace fixed delay with prompt-based flow control

sensei-hacker:
## Summary

Replaces the fixed 50ms delay between CLI commands with prompt-based flow control. Previously `executeCommands` used arbitrary delays (`lineDelayMs: 50`, `profileSwitchDelayMs: 100`) that caused commands to be silently dropped when the FC needed more time to process them (profile switches, flash writes, heavy settings loads).

## Changes

- Remove `lineDelayMs` and `profileSwitchDelayMs` — no longer needed
- `executeCommands` now waits for the FC's `# ` prompt before sending the next command
- Sliding window of 2: seeds with 2 commands immediately, then sends one more per prompt received — hides round-trip latency without overflowing the FC's serial buffer
- `TABS.cli.read()` fires `promptCallback` only when `cliBuffer.endsWith('# ')` (the actual FC prompt), not on every incoming data chunk
- 5s per-command timeout fallback for FC disconnect/crash mid-send, with generation counter to prevent double-counting if a real prompt arrives just after the timeout fires
- Empty lines filtered from batch sends — trailing newlines in loaded files no longer cause a 5s stall

## Testing

- Interactive typing: typed `version`, response received, no JS errors
- Diff All (635-line response): received correctly, no JS errors
- 9-command batch: all 9 commands applied, exactly 9 prompt fires confirmed (96–128ms each, 896ms total), `cliBuffer = "# "` on every fire — proving the guard condition works
- Full 27-tab sweep with FC connected: zero console errors

## Code Review

Reviewed with inav-code-review agent. Three critical issues found and fixed before PR:
1. Missing `endsWith('# ')` guard — would have fired callback on every data chunk
2. Empty-line stall from trailing newlines in loaded files
3. Timeout race condition — generation counter prevents double-counting

Fixes silent command dropping when loading large diffs over CLI.

### DIFF
--- a/tabs/cli.js
+++ b/tabs/cli.js
@@ -16,10 +16,10 @@ import { generateFilename } from './../js/helpers';
 import dialog from '../js/dialog';
 
 TABS.cli = {
-    lineDelayMs: 50,
-    profileSwitchDelayMs: 100,
     outputHistory: "",
     cliBuffer: "",
+    promptCallback: null,
+    promptTimeoutId: null,
     GUI: {
         snippetPreviewWindow: null,
     },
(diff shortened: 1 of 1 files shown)


sensei-hacker:
@Jetrell  will like this.
I can't find the related issue to close it.

PR #2621, opened 2026-05-16
output assignment: firmware-authoritative MSP2 READ/QUERY API for Configurator

sensei-hacker:
## Problem

The Mixer/Outputs tab re-implements the firmware's output assignment algorithm in JavaScript (`getTimerMap()` in `outputMapping.js`). Any version mismatch between Configurator and firmware causes the wrong pin assignment preview, and on targets with explicit timer overrides (`OUTPUT_MODE_MOTORS`/`OUTPUT_MODE_SERVOS`) the assignment silently changes at firmware upgrade.

## Solution

Make the firmware the sole authority on output assignments. On Mixer tab load, the Configurator requests the firmware's actual post-boot assignment. When the user changes a timer override, the Configurator queries the firmware for what the assignment *would be*, without save+reboot.

Old firmware returns an error; the Configurator automatically falls back to the existing `getTimerMap()` JS algorithm — no behaviour change for users on older firmware.

## Changes

- **`js/msp/MSPCodes.js`**: add `MSP2_INAV_OUTPUT_ASSIGNMENT` (0x210E) and `MSP2_INAV_QUERY_OUTPUT_ASSIGNMENT` (0x210F)
- **`js/outputMapping.js`**: add `directAssignments` storage; `setDirectAssignment()`, `hasDirectAssignment()`, `flushDirectAssignment()`, `getOutputTableDirect()` methods
- **`js/msp/MSPHelper.js`**: decoders for both new messages; `loadOutputAssignment()` and `queryOutputAssignment()` helpers
- **`tabs/mixer.js`**: load direct assignment on tab open; prefer `getOutputTableDirect()` when available; send QUERY on timer-override dropdown change for live preview

## Testing

- Flashed firmware PR companion to BROTHERHOBBYF405V3
- `0x210E` READ: returns 8 valid `(outputIndex, type, number)` tuples — Motor 1–4 on outputs 0–3, Servo 1–4 on outputs 4–7
- `0x210F` QUERY: result with current overrides matches READ exactly
- Mixer tab: `hasDirectAssignment()` returns true, `getOutputTableDirect()` produces correct `["Motor 1", "Motor 2", ..., "Servo 1", ...]` table (9 outputs confirmed via CDP)
- Fallback: tested against old firmware — no response causes automatic fallback to `getTimerMap()`

## Code Review

Code review performed — no critical issues found.

Companion firmware PR: iNavFlight/inav#XXXX


sensei-hacker:
/review
-i

PR #2625, opened 2026-05-20
fix: reopen correct tab after Save and Reboot

sensei-hacker:
## Summary

Fixes two bugs related to saving settings and rebooting.

---

### Bug 1: Wrong tab shown after Save and Reboot (Welcome tab error)

After clicking \"Save and Reboot\" from Outputs, Mixer, or most other tabs, the configurator showed \"You need to upgrade your firmware before you can use the Welcome tab\" and failed to navigate back to the correct tab.

**Root cause:** `handleReconnect()` stored `reopenTab` inconsistently — the `boolean=true` path captured the `<li>` element, while most tab callers passed the `<a>` element directly. `onValidFirmware` then called `$('a', reopenTab).trigger('click')`, which calls `$(anchor).find('a')` when `reopenTab` is already an `<a>`, finding nothing. No tab navigation happened after reconnect, leaving `li.tab_landing` still marked `active` from the disconnect phase. `GUI.updateActivatedTab()` (called by the periodic profile-change MSP callback) then re-triggered the landing tab click — but `allowedTabs` was in connected mode (no `landing`), producing the misleading error.

**Fix:** Normalise `reopenTab` to always store the `<a>` element at assignment time in `handleReconnect()`, for all input variants: `boolean=true`, jQuery `<a>` (most tab callers), jQuery `<li>` (CLI exit path), and `false`/null. `onValidFirmware` then calls `.trigger('click')` directly.

---

### Bug 2: Crash in Advanced Tuning Save and Reboot (race condition)

Clicking Save and Reboot quickly after opening Advanced Tuning could crash with:
`TypeError: Cannot read properties of null (reading 'setting')`

**Root cause:** `Settings.configureInputs()` loads settings asynchronously after the tab renders. If the user saves before it finishes, some inputs have no `data-setting-info` yet. `processInput()` returns `null` for these, and `pickAndSaveSingleInput` crashed accessing `null.setting`. (Once `configureInputs` completes, inputs for unknown settings are removed from the DOM and never reached.)

**Fix:** Null-guard in `pickAndSaveSingleInput` — skip inputs that return `null` from `processInput` rather than crashing.

---

## Changes

- `js/serial_backend.js`: Normalise `reopenTab` to always be `<a>` in `handleReconnect()`; simplify trigger in `onValidFirmware`
- `js/settings.js`: Skip null `settingPair` in `pickAndSaveSingleInput`

## Testing

Tested Save and Reboot from the following tabs — each correctly reopened its own tab after reconnect with no error in the log:

- Outputs ✅
- Mixer ✅
- Ports ✅
- Configuration ✅
- PID Tuning boolean path — `$('#tabs > ul li.active a')` resolves correctly ✅

## Code Review

Reviewed with inav-code-review agent. Review caught the CLI exit `<li>` path in the `else` branch of `handleReconnect` — addressed by normalising all non-boolean inputs.

### DIFF
--- a/js/msp/MSPHelper.js
+++ b/js/msp/MSPHelper.js
@@ -86,13 +86,13 @@ var mspHelper = (function () {
 
                 let profile_byte = data.getUint8(offset++)
                 let profile = profile_byte & 0x0F;
-                if (profile !== FC.CONFIG.profile) {
+                if (profile !== FC.CONFIG.profile && FC.CONFIG.profile !== -1) {
                     profile_changed |= GUI.PROFILES_CHANGED.CONTROL;
                 }
                 FC.CONFIG.profile = profile;
 
                 let battery_profile = (profile_byte & 0xF0) >> 4;
-                if (battery_profile !== FC.CONFIG.battery_profile) {
+                if (battery_profile !== FC.CONFIG.battery_profile && FC.CONFIG.battery_profile !== -1) {
                     profile_changed |= GUI.PROFILES_CHANGED.BATTERY;
                 }
                 FC.CONFIG.battery_profile = battery_profile;
@@ -104,7 +104,7 @@ var mspHelper = (function () {
                 //read mixer profile as the last byte in the the message
                 profile_byte = data.getUint8(dataHandler.message_length_expected - 1);
                 let mixer_profile = profile_byte & 0x0F;
-                if (mixer_profile !== FC.CONFIG.mixer_profile) {
+                if (mixer_profile !== FC.CONFIG.mixer_profile && FC.CONFIG.mixer_profile !== -1) {
                     profile_changed |= GUI.PROFILES_CHANGED.MIXER;
                 }
                 FC.CONFIG.mixer_profile = mixer_profile;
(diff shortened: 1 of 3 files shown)


Jetrell:
It fixes the Welcome tab issue and the firmware mismatch, 

However the issue concerning the Advanced Tuning tab is still there.
e.g. If I make a change in the Advanced Tuning tab, then save and reboot. The tab sometimes doesn't repopulates.  While other times it does. But in my case. I had a MC configured FC repopulate the Advanced Tuning tab with fixedwing auto launch, fixedwing navigation and fixedwing landing settings at the top of the page. And MC settings at the bottom of the tab.

Jetrell:
Sorry to complain. I tried these latest commits with Windows 64. And unfortunately it still shows the top half of the Advanced tab as FW and the bottom half as MC. Even though its a MC platform.

AT mc and fw

sensei-hacker:
> Sorry to complain. I tried these latest commits with Windows 64. And unfortunately it still shows the top half of the Advanced tab as FW and the bottom half as MC. Even though its a MC platform.
> 
> AT mc and fw


So it's trying to tell you to make a VTOL? 😁

sensei-hacker:
@Jetrell  
Thanks so much for the heads up!  This fix should make it stop trying to force you to build a VTOL:
https://github.com/iNavFlight/inav-configurator/pull/2626


> Sorry to complain

I figure hearing it from you is 7392% better than hearing it from a bunch of people on Facebook after the release. Thanks!

PR #2626, opened 2026-05-23
fix(advanced-tuning): load mixer config before rendering tab

sensei-hacker:
## Summary

Fixes a regression introduced alongside progressive settings loading (PR #2625): the Advanced Tuning tab shows both fixed-wing and multirotor sections simultaneously, regardless of craft type.

## Root Cause

`processHtml()` calls `FC.isAirplane()` / `FC.isMultirotor()`, both of which read `FC.MIXER_CONFIG.platformType`. The tab never fetched MIXER_CONFIG itself — it relied on it being pre-loaded by another tab. If `platformType` was still `-1` (the `resetState()` default), both methods return false and the `else` branch renders both sections at once.

## Changes

- `tabs/advanced_tuning.js`: Add MSPChainer to fetch `loadMixerConfig` before calling `load_html()`, ensuring `platformType` holds the real FC value when `processHtml()` applies show/hide logic. Matches the established pattern in `pid_tuning.js` and `mixer.js`.
- `tabs/advanced_tuning.js`: Remove pre-existing duplicate semicolon (`i18n.localize();;`).

## Testing

- Connected to SITL (multirotor, platformType = 0)
- Advanced Tuning tab: multirotor section visible, airplane section hidden — correct
- No JavaScript errors in console
- Code reviewed with inav-code-review agent — no critical issues found

## Related

Complements PR #2625 (progressive settings loading)

### DIFF
--- a/tabs/advanced_tuning.js
+++ b/tabs/advanced_tuning.js
@@ -17,7 +17,7 @@ advancedTuningTab.initialize = function (callback) {
         GUI.active_tab = this;
     }
 
-    var loadChainer = new MSPChainerClass();
+    const loadChainer = new MSPChainerClass();
     loadChainer.setChain([mspHelper.loadMixerConfig]);
     loadChainer.setExitPoint(load_html);
     loadChainer.execute();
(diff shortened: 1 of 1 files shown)


Jetrell:
I can confirm it now works correctly again in tests.

PR #2632, opened 2026-05-24
Error414/fix/open tab after reconnect

error414:
How to reproduce problem:
* connect to FC
* go to port tab
* hit "save and reboot" button
* wait for reconnecting

What happen: 
no tab is opened
image

What should happen: 
open last tab

this issue is fixed by commit https://github.com/iNavFlight/inav-configurator/commit/9b5c24cccf7aac7890328cadf5d90310ca6ea348 

Second commit is fix double loading page, there was race condition with loading profile.

### DIFF
--- a/js/msp/MSPHelper.js
+++ b/js/msp/MSPHelper.js
@@ -86,13 +86,13 @@ var mspHelper = (function () {
 
                 let profile_byte = data.getUint8(offset++)
                 let profile = profile_byte & 0x0F;
-                if (profile !== FC.CONFIG.profile) {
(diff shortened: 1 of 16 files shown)


sensei-hacker:
Looks familiar 🙂

https://github.com/iNavFlight/inav-configurator/pull/2625

error414:
merget 15 min before, you forgot to that PR :D. 

I noticed that bug when I fixing CPA, my OCD can't live with that :D. Next time I should do more research in pull requests.

error414:
clossing, [sensei-hacker](https://github.com/sensei-hacker) already fixed it

PR #2636, opened 2026-05-25
Fix OSD custom element layout on narrow/HD and wide screens

sensei-hacker:
## Summary

- **Bug 1 (overflow on narrow screens with HD OSD):** Removed `width: auto !important; flex-shrink: 0` from `.ce-source-select` and `.ce-format-select`, which was preventing the slot rows from fitting in the narrow side panels of HD OSD types (HDZero, DJI HD, etc.)
- **Bug 2 (LC condition driver selector missing in HD mode):** Moved the format select into a dedicated sub-row (`$fmtRow`) that appears only when source is GV or LC, freeing the main row to show the LC/GV driver selector at a usable width in all OSD modes
- **Responsive improvement:** Added `@media (max-width: 1439px)` rule reducing `content_wrapper` padding from 20px to 10px, giving each side panel ~10px more width on narrower screens
- **Wide-screen improvement:** Added `max-width: 300px` to `.osd_group` to prevent card content rows from stretching excessively at 1920px+

## Test plan

- [ ] At 1366×768 with HD OSD (HDZero/DJI/Avatar/BFHDCompat): slot rows fit within side panels without overflow
- [ ] HD OSD mode: LC condition driver selector dropdown is visible
- [ ] PAL/NTSC mode: LC condition driver selector still visible (no regression)
- [ ] Selecting GV or LC as source shows the format sub-row; other sources hide it
- [ ] At 1920×1080: card content rows are capped at a reasonable width, not stretching edge-to-edge
- [ ] At <1440px width: side panels are slightly wider due to reduced padding

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -1972,6 +1972,15 @@ dialog {
     border-bottom-right-radius: 5px;
 }
 
+@media only screen and (max-width: 1439px) {
+
+    .content_toolbar .btn a {
+        padding: 0 5px;
+        line-height: 22px;
+    }
+
+}
+
 @media only screen and (max-width: 1055px) , only screen and (max-device-width: 1055px) {
 
     .content_wrapper {
--- a/src/css/tabs/osd.css
+++ b/src/css/tabs/osd.css
@@ -693,6 +693,12 @@
     padding-bottom: 0 i !important;
 }
 
+@media only screen and (max-width: 1439px) {
+    .tab-osd .content_wrapper {
+        padding: 10px;
+    }
+}
+
 @media only screen and (max-width: 1055px) , only screen and (max-device-width: 1055px) {
     .tab-osd .content_wrapper {
         height: calc(100% - 30px);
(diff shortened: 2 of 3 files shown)


Jetrell:
Thanks for the fix.
The custom font box width UI issue is now fine.

The only other thing I see that has occurred. Is the screen width compression has now placed SAVE button on top of the Analog Font Manager button, and increased its box height. Instead of them being side by side.

Fix

Jetrell:
@sensei-hacker Beside the previous post. There are also various other tabs that are now **cutoff** at the bottom. Due the Save and Save - Reboot buttons being so large. I wasn't sure if this button size increase was by design or accident. It just seems to take up a lot of the space on each tab.

Tuning filter 
Tuning filters

Tuning Mechanics
Tuning mechanics

Blackbox tab
Blackbox

CLI
CLI

sensei-hacker:
thanks, Jetrell!

Jetrell:
I checked all tabs, along with the OSD tab. And it now appears to work as it should. Thank you

PR #2641, opened 2026-05-27
Fixed wing Auto speed support

breadoven:
Support for https://github.com/iNavFlight/inav/pull/11595.

### DIFF
--- a/js/fc.js
+++ b/js/fc.js
@@ -281,7 +281,7 @@ var FC = {
             packetCount: 0,
             hwVersion: 0
         };
-        
+
         this.ADSB_VEHICLES = {
             vehiclesCount: 0,
             callsignLength: 0,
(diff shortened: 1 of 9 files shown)


Jetrell:
BTW the Auto Speed element doesn't show in the OSD tab.. By the look of it. It should be between 3Dspeed and GPS max speed.

breadoven:
> BTW the Auto Speed element doesn't show in the OSD tab.. By the look of it. It should be between 3Dspeed and GPS max speed.

Hmmm, that's because it's limited to V 10.0. Not sure what the answer is here other than to remove the version limit for now and add it when it's merged. I'll change this PR anyway because the OSD text is wrong.

Jetrell:
What do you think about adding a `SET SPD` symbol to the OSD font table ? As @dcan999 recommended.
It would provide better explanation of the function.
I know A and G still look a bit hack.  And I guess they could be replaced with 151 and 222.. But for the sake of useless DJI not updating their font package.. Then I guess the SetSpd symbol with A or G following would still look better for Analog, Avatar and HDzero. With DJI just having to miss out on the SetSpd symbol.

setspd

[223 SetSpd (12x18) analog.zip](https://github.com/user-attachments/files/28788609/223.SetSpd.12x18.analog.zip)

PR #2643, opened 2026-05-31
OSD custom elements: add 16-bit icon support

error414:
The icon input for static custom element icons was restricted to values 0–255. This change raises the limit to 65535 and updates the icon picker to display the full font character set (up to 512 characters), so users can select and save 16-bit icon values that the firmware
 

image

Inav PR: https://github.com/iNavFlight/inav/pull/11608

### DIFF
--- a/tabs/osd.js
+++ b/tabs/osd.js
@@ -3793,7 +3793,8 @@ function openIconPicker($targetInput) {
     var $grid = $('<div>').addClass('ce-icon-picker-grid');
     var currentVal = parseInt($targetInput.val()) || 0;
 
-    for (var c = 1; c <= 255; c++) {
+    var maxFontChar = (FONT.data && FONT.data.character_image_urls.length > 0) ? FONT.data.character_image_urls.length - 1 : 511;
+    for (var c = 1; c <= maxFontChar; c++) {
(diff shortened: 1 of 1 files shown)


Jetrell:
@sensei-hacker I can confirm this works, if you want to add it with https://github.com/iNavFlight/inav-configurator/pull/2560 in 9.1.
Just don't forget to merge the firmware too. As I made the mistake of not loading to begin with.

sensei-hacker:
Thanks for the test, Jetrell.

"Just don't forget to merge the firmware too."  That sounds like the definition of next major version, INAV 10?

error414:
firmware is ok, there are no any breaking changes, but in configurator is  beraking change, I would preffer to add it to next major version, INAV 10.

error414:
BTW: I have to say that I like how new OSD custom elements look like. Nice work @sensei-hacker

Jetrell:
> BTW: I have to say that I like how new OSD custom elements look like. Nice work @sensei-hacker

This is why I suggested it would be good to add your work here, along with 2560. Because it would effectively make the Custom OSD improvements complete for 9.1.

error414:
yeah, but then configurator for INAV 9.1 would not fully work for FW INAV 9.0. I think @sensei-hacker does not want breaking changes in minor versions.

PR #2644, opened 2026-05-31
fix(vtol): sync control profile when mixer profile changes

sensei-hacker:
## Summary

When a VTOL aircraft switches mixer profiles (MC↔FW), the control profile should automatically switch to match, if mixer_control_profile_linking is turned on.

## Testing

Tested on VTOL (AEDROXH7, INAV 9.0.1) with the dev configurator:

- **Before fix:** Switching Mixer Profile 1→2 left Control Profile at 1 after reboot
- **After fix:** MC→FW (Mixer 1→2): both Control Profile and Mixer Profile show 2 ✓
- **After fix:** FW→MC (Mixer 2→1): both Control Profile and Mixer Profile show 1 ✓

## Related Issues

Regression introduced in commit 02cc9287b5 (PR: "Refactor tab handling to improve tab switching logic and add confirmation dialog for mixer profile changes")


Jetrell:
This has partially fixed the issue. It now switches in the configurator via the RC channel switch.

But pressing the Mixer Profile button in the top status bar is still broken.
It will switch from Mixer and Control profile 1 to Mixer/Control profile 2 (MC is 2).
But when the Mixer Profile button is pressed again, to change back from Mixer/Control profile 2 to Mixer/Control profile 1. The Mixer Profile will change. But the Control Profile is stuck in profile 1... And from that point on-wards, when I press the Mixer Profile button. It tries to change. But after the reboot, the preview image and the profiles stay perpetually lock in Mixer Profile 2 and Control Profile 1, with a MC image preview.

I always flush the cache before running tests. Just so you know.
I also checked with earlier builds, before mid last month. And they work correctly.

Jetrell:
@sensei-hacker It is now working as it should after the last changes. Thanks

PR #2645, opened 2026-06-01
fix: remove duplicate event handlers introduced by fdd4fcb8

daijoubu:
## Summary

- Commit `fdd4fcb8` re-added the accordion navigation block and the `disable_3d_acceleration` options handler that were already present in `configurator_main.js`, creating duplicate registrations of each
- The doubled accordion handlers caused every group-header click to toggle `.active` twice (on then immediately off), making it impossible to expand or collapse navigation groups
- The doubled `disable_3d_acceleration` handlers caused the change event to fire twice and the checkbox state to be initialised twice on options open

## Changes

- Remove duplicate accordion `.group-header` click/keydown handlers, `updateToggleAllButton()`, `#toggleAllGroups` click handler, and initialisation block (77 lines)
- Remove duplicate `disable_3d_acceleration` checkbox init and change handler (9 lines)

## Test plan

- [x] Navigation groups expand and collapse on single click
- [x] "Expand All / Collapse All" toggle works correctly
- [x] Options dialog `disable_3d_acceleration` checkbox loads and saves correctly
- [x] No regressions in other tab navigation behaviour

### DIFF
--- a/js/configurator_main.js
+++ b/js/configurator_main.js
@@ -361,83 +361,6 @@ $(function() {
         updateToggleAllButton();
 
 
-        // Accordion Navigation Groups
-        $('.group-header').on('click', function(e) {
-            e.stopPropagation(); // Prevent triggering tab click
-            const header = $(this);
-            const items = header.next('.group-items');
-
-            // Toggle this group
-            header.toggleClass('active');
-            items.toggleClass('expanded');
-
-            // Update aria-expanded for accessibility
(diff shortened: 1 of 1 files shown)


sensei-hacker:
Thanks for this. I had included this fix in the PR right before this one, in #2645

daijoubu:
No worries, it was my first stumbling block in trying to figure out how to extend configurator.  I'd assumed my changes broke it for a while.

PR #2661, opened 2026-06-16
Add LIS2MDL magnetometer visualizer option

bkleiner:


sensei-hacker:

PR #2664, opened 2026-06-21
Sync Alignment Tool sliders live as Value fields are edited

### LINKED ISSUE #2589: The Alignment Tool's sliders are not linked to the Value fields
In INAV-Configurator 9.0.0 the Alignment Tool's sliders are not updated if you manually edit the Value fields.

See this image:
Image

the YAW Value of the magnetometer was modified to 90, but the slider stays at position 0.

It seems that this is an issue for all 6 sliders on the Alignment Tool page

Archie-Redford:
## Summary

Fixes #2589 - the Alignment Tool sliders on the Magnetometer tab (board roll/pitch/yaw and mag roll/pitch/yaw) appeared to not update when manually editing the corresponding numeric Value field.

## Root cause

The Value field handlers were bound to the `change` event only, which for `<input type="number">` fires on blur, not on keystroke. So while a user is actively typing a new value, the field's displayed text updates immediately (native browser behavior), but the slider doesn't move until the field loses focus - looking like the slider is stuck/not linked, exactly as described in the issue.

## Fix

- Bind the six Value field handlers to `input change` instead of just `change`, so the slider tracks the field live as you type, matching the expected behavior.
- Added a `NaN` guard in `clamp()` (and a `null` check in each `updateXAxis` function, mirroring the existing guard already present in `updateBoardRollAxis`) so a momentarily empty/invalid field (e.g. while backspacing before retyping) doesn't move the slider to a garbage position or throw.

## Test plan

Verified manually via SITL "Demo mode" (Playwright-driven Electron session) on the Magnetometer > Alignment Tool tab:
- [x] Before fix: typing a new Yaw value while the field still has focus leaves the slider's internal value unchanged until blur (reproduces the report).
- [x] After fix: the slider's value updates immediately as each keystroke lands, matching the field.
- [x] Clearing the field entirely mid-edit no longer throws and leaves the slider at its last valid position.
- [x] Normal blur-commit flow (tabbing away) still works and clamps to the field's min/max as before.
- [x] Verified for all 6 sliders (board roll/pitch/yaw, mag roll/pitch/yaw) via shared code path.


Archie-Redford:
Closing - reclaiming this contribution for now.

PR #2669, opened 2026-06-30
Fix BLE device chooser: keep the discovered device list stable

julpel8:
### Problem

The BLE device chooser rebuilt its entire list on every scan update: all rows were removed, the devices re-sorted alphabetically, then re-inserted. Since advertisements keep arriving while the scan runs, the list constantly reshuffles:

- a row can move (or be replaced by another device) between the moment you aim at it and the moment you click it;
- a device whose name resolves after its first advertisement jumps to a different position;
- typing in the search box re-sorted and re-rendered everything as well.

### Fix

`bt-device-chooser-renderer.js` now renders incrementally, in discovery order:

- each device gets its DOM row once; rows are kept and never re-created, new devices are appended just above the Cancel button;
- when a name resolves later, the existing row's label is updated in place;
- the search box only toggles row visibility (`display`), it no longer reorders or rebuilds the list.

Net effect: the list only grows, rows stay where they are, and clicks land on the device the user actually aimed at.

### How to test

1. Open the BLE connection dialog with several BLE devices powered on nearby.
2. Watch the list while the scan runs: previously discovered rows must stay in place while new ones are appended at the bottom, and names that resolve late must update without moving the row.
3. Type in the search field: non-matching rows are hidden, the remaining ones keep their order and position.
4. Click a device: the correct device is selected and the window closes.

### Notes

UI-only change in the chooser window, no MSP/protocol impact, so no firmware/configurator compatibility implications — targeted at `maintenance-9.x` per the branch-targeting guidance.

### DIFF
--- a/js/libraries/bluetooth-device-chooser/bt-device-chooser-renderer.js
+++ b/js/libraries/bluetooth-device-chooser/bt-device-chooser-renderer.js
@@ -1,56 +1,64 @@
 import './bt-device-chooser-style.css'
 
 document.addEventListener("DOMContentLoaded", () => {
-    // Store all discovered devices
+    const MAX_DEVICE_NAME_LENGTH = 45;
+
+    // Store all discovered devices (Map preserves discovery order)
     const devices = new Map();
+    // Track the DOM row created for each device so we can update it in place
+    const deviceElements = new Map();
 
     // Get DOM elements
     const searchInput = document.getElementById('search');
     const listElement = document.getElementById('list');
     const cancelElement = document.getElementById('cancel');
 
-    // Render devices in sorted and filtered order
-    function renderDevices() {
+    function deviceLabel(device) {
+        const text = `${device.deviceName} (${device.deviceId})`;
+        return text.length > MAX_DEVICE_NAME_LENGTH ? text.substring(0, MAX_DEVICE_NAME_LENGTH) : text;
+    }
+
+    // Show/hide rows according to the search text, without moving or recreating them
+    function applyFilter() {
         const searchText = searchInput.value.toLowerCase().trim();
-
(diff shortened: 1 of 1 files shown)


sensei-hacker:

sensei-hacker:
I understand the goal you're shooting for here and it could be valuable.
On the other hand, it would undo another valuable thing. Sometimes there are LOT of devices detected and it's much easier to select the one you always use if they are ordered. Especially after you develop a bit of muscle memory for going to approximately the right place in the list.

Maybe the best of both worlds would be something like doing the initial discovery scan and after X seconds, fix the order.  Then additional devices discovered AFTER that initial scan and and ordering get placed without resorting the whole list. Or something like that.

sensei-hacker:
Following up on my earlier comment — I've been thinking about the stable-vs-sorted tension, and I think the best of both worlds is a **pinned/known-devices split** rather than a time-based freeze:

- **Known devices first:** devices the user has connected to before (persist the last-connected device IDs, e.g. in localStorage) are shown at the top, sorted alphabetically, and always in the same position.
- **Subtle visual separation:** a light divider line or color shift between the known section and the rest — no big UI.
- **New devices below, in discovery order:** everything else renders exactly like this PR does now (incremental, append-only, stable rows), so the click-safety improvement stays intact.

Why this beats a freeze-timer: muscle memory works because your *usual* device is always in the same spot regardless of when it advertises or whether it resolves its name late — a timer-based sort can't promise that. And for people with lots of devices, the list stays readable without ever reshuffling rows under the cursor.

Does that direction make sense to you? If you'd rather keep this PR focused on the incremental-rendering fix, the pinned section could be a small follow-up — but it feels like a natural fit for this same file.

PR #2677, opened 2026-07-09
Fix OSD preview if filter is active

error414:
There are two issue in OSD tab

1) is not possible to move OSD element in preview if OSD element is hide by filter
2) broken design of bottom buttons

Before fix:
image

After fix:
image


sensei-hacker:
I think the button stacking was already addressed in 9.1? At least, I can;t reproduce it anymore. Can you please confirm?

error414:
Ahaa, good catch

I thought and have no idea why I thought that after release inav 9.1 is maintanence-9.x merged to maintanence-10.x. 

But it's not still correctly fixed 

Original 9.1 configurator
---------------------------------
* write to filter "thr"
* make configurator window on full screen on 2k monitor (or at least window must be big enoight not to active reduce size of buttons)
* scrool down

image
The scroolbar is under button toolbar and last OSD element is not visible. 


In this case would be probably better if I fix it it in maintanence-9.x branch

error414:
I reverted changes about buttons which are partly fixed in maintanence-9.x. This PR is only about fixing

PR #2680, opened 2026-07-10
outputs tab allow 18 servos

sensei-hacker:
Update Configurator outputs tab to handle 18 (or more) servos, since the firmware now does.
Previously, it displayed an empty "servo 0" which was internally servo # -2 because it subtracts from the "maximum" of 16.

Should fix the issue brought up in #2679

### DIFF
--- a/src/css/tabs/motors.css
+++ b/src/css/tabs/motors.css
@@ -198,7 +198,7 @@
 
 .tab-motors .servos .titles li {
     float: right;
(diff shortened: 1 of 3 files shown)


sensei-hacker:
@daijoubu  how does this work for you?

daijoubu:
This fixes the issue of servo 1 being labeled as servo 2 for the limits and trim adjustment and servo graphics also being off by one.

PR #2694, opened 2026-07-25
Dual Receiver support

xznhj8129:
# Dual Receiver support

Configurator support for INAV Dual RX.

This exposes the second receiver configuration, allows each receiver to be assigned to its own serial port, shows the state of both links live, and provides manual/Programming Framework handover controls.

Companion firmware PR: iNavFlight/inav#11748.

### Dual RX configuration

Dual RX can be enabled from the Configurator and exposes separate receiver configuration for:

* RX1 receiver type
* RX2 receiver type
* RX1 serial provider
* RX2 serial provider
* inversion
* half-duplex
* RX2 SBUS sync interval where applicable

RX1 remains the normal existing receiver configuration. RX2 is an additional receiver configuration used when Dual RX is enabled.

The existing channel map is currently **global**. For example, an `AETR` map is applied to both RX1 and RX2 rather than being configured separately per receiver.

### Ports

Adds the secondary serial receiver function:

```text
RX_SERIAL_SECONDARY
```

using firmware serial-function bit 13:

```text
FUNCTION_RX_SERIAL_SECONDARY = 1 << 13
```

A typical serial Dual RX setup therefore has:

```text
UART A -> RX_SERIAL
UART B -> RX_SERIAL_SECONDARY
```

The two receivers may use the same supported protocol or different supported protocols.

### Receiver status

The Receiver tab displays live Dual RX state, including:

* active link
* RX1 state
* RX2 state
* per-link RSSI/LQ/SNR when the receiver protocol supplies them
* overall Dual RX status
* last handover reason and time
* manual handover controls

The status is obtained using:

```text
MSP2_INAV_GET_LINK_STATS = 0x2103
```

The returned status includes configured, initialized and valid-link state separately, so the UI can distinguish cases such as:

```text
OK
Lost
Not configured
Initialization error
```

rather than treating every unavailable receiver identically.

Per-link statistics are only displayed when the firmware reports both the link statistics and the corresponding statistic fields as valid.

### Manual handover

The Receiver tab provides:

```text
Select RX1
Select RX2
```

controls using:

```text
MSP2_INAV_SET_RX_LINK = 0x2232
```

The firmware only accepts a handover to a currently valid receiver.

The button for the active receiver is disabled, as is a button for a receiver which is currently unavailable.

This is an explicit handover event, not a persistent receiver preference.

### Programming Framework

Adds the Dual RX handover Logic Condition operation:

```text
64: RX handover
```

matching the current firmware enum:

```text
LOGIC_CONDITION_ACTIVATE_LANDING = 63
LOGIC_CONDITION_RX_HANDOVER      = 64
```

Operand A selects the requested receiver:

```text
1 = RX1
2 = RX2
```

This allows a Logic Condition to request an explicit receiver handover.

### OSD

Adds Configurator support for the firmware Dual RX OSD items:

```text
ACTIVE_RX_LINK
RX1_LINK_STATS
RX2_LINK_STATS
```

These allow the active receiver and individual link state/statistics to be shown in the OSD.

---

## How to configure it

A serial Dual RX setup currently involves three parts of Configurator.

### 1. Enable Dual RX

Enable Dual RX in the Configuration tab and save/reboot.

### 2. Assign the receiver ports

In the Ports tab:

```text
RX1 UART -> Serial RX
RX2 UART -> Secondary Serial RX
```

Each receiver must have its own appropriate serial port.

If a receiver protocol also requires another serial function on the same port, such as MAVLink telemetry, configure that as required by that protocol.

### 3. Configure both receivers

In the Receiver tab, configure the primary and secondary receiver independently.

For example:

```text
RX1:
Receiver type: SERIAL
Provider: CRSF

RX2:
Receiver type: SERIAL
Provider: SBUS
```

or:

```text
RX1: CRSF
RX2: CRSF
```

Save and reboot.

The Dual RX Status section should then show both configured links and which one is currently active.

---

## Runtime behavior

RX1 is the initial active receiver.

When both links are healthy, receiving valid data on RX2 does not automatically replace RX1.

If the currently active receiver becomes invalid and the other receiver is valid, firmware transfers control to the surviving link.

Recovery of the previously lost receiver does not automatically take control back.

For example:

```text
Boot:
RX1 active
RX2 standby

RX1 lost:
RX2 becomes active

RX1 recovers:
RX2 remains active

RX2 later lost:
RX1 becomes active
```

The Configurator reflects this firmware state rather than implementing receiver-selection policy itself.

Manual handover through the Receiver tab or Programming Framework can explicitly select the other currently valid receiver.

---

## Supported combinations

The firmware decides whether a receiver pair is supported.

Same-driver Dual RX is supported for drivers which have independent per-link state, including:

```text
CRSF + CRSF
SBUS-family + SBUS-family
MAVLink + MAVLink
```

Mixed receiver types such as:

```text
CRSF + SBUS
CRSF + MAVLink
CRSF + MSP
```

are also supported where the corresponding firmware drivers can operate independently.

Some same-driver combinations remain intentionally unsupported where the underlying receiver/telemetry implementation still uses shared singleton state.

Configurator reports these cases through the firmware Dual RX status instead of silently hiding the second receiver.

---

## Firmware compatibility

This Configurator change targets the current INAV 10 Dual RX implementation.

Important parts of the current firmware contract are:

```text
FUNCTION_RX_SERIAL_SECONDARY = 1 << 13

MSP2_INAV_GET_LINK_STATS = 0x2103
MSP2_INAV_SET_RX_LINK    = 0x2232

LOGIC_CONDITION_RX_HANDOVER = 64
```

Dual RX configuration is carried by the extended RX configuration data/settings. It does **not** use an `activeRX` byte appended to `MSP2_INAV_MISC`.

The active receiver is runtime state reported by `MSP2_INAV_GET_LINK_STATS`.

---

## Testing

The Configurator has been exercised against current Dual RX firmware on real hardware.

Confirmed configurations include:

```text
CRSF + CRSF
CRSF + SBUS
```

with live receiver state visible in Configurator and receiver failover functioning on hardware.

The Configurator-side status display has also been verified with real CRSF link statistics.

Firmware-side SITL coverage separately exercises receiver selection, failure/recovery behavior, mixed receiver configurations and explicit handover.

---

## Current UI/workflow limitations

The functionality works, but the setup workflow can still be improved.

At present configuration is spread across:

```text
Configuration tab -> enable Dual RX
Ports tab         -> assign RX1/RX2 ports
Receiver tab      -> configure both receivers and inspect status
```

That is functional but not especially discoverable.

There are also some smaller presentation issues worth improving:

* the channel bars show the currently published/active RC stream but do not clearly label which RX is providing it
* the channel map is global to both receivers, which is not obvious from the UI
* the last-switch timestamp is currently displayed in milliseconds, which is unnecessarily fine-grained for the user-facing display
* the status UI could identify the receiver involved in a link-loss handover more explicitly
* secondary-receiver protocol-specific configuration could be presented more consistently

These are UI/workflow improvements rather than blockers for the current Dual RX functionality.

The current goal of this PR is to expose the firmware feature correctly and make Dual RX usable in Configurator without inventing a separate receiver-selection model in the UI.


sensei-hacker:

xznhj8129:
body updated

sensei-hacker:
Confirmed still accurate — neither of the last two commits touched this. `cfa5bb8a` (Aug 29) is what added the `64: { name: "RX handover", ... }` entry to `js/logicConditionOperators.js`, and it's still 64 as of the current head; `ae33d65` (Aug 31) only touched `tabs/osd.js`.

On the firmware side (`xznhj8129/feature/dualrx`), the operation is `LOGIC_CONDITION_RX_SELECT_OVERRIDE = 63`, so the configurator is still serializing the wrong ID.

One extra wrinkle: that firmware branch is stale vs. current `maintenance-10.x` — it's missing `LOGIC_CONDITION_DISABLE_AUTOSPEED_AIRSPEED` (61 upstream), so `ACTIVATE_RTH`/`ACTIVATE_LANDING` currently sit at 61/62 there instead of 62/63. After rebasing onto `maintenance-10.x`, 63 will already be taken by `ACTIVATE_LANDING`, so `RX_SELECT_OVERRIDE` will likely need to become 64 there too. Please rebase the firmware branch first, then set the configurator's key to match whatever `RX_SELECT_OVERRIDE` ends up as — not just flip it to 63.

PR #2699, opened 2026-08-03
fix(map-generator): handle all-sea Copernicus neighbour tiles + tab cleanup

MartinovEm:
Two fixes for the Map Generator terrain source, both found after #2599 merged (thanks to @error414 for testing).

**Coastal areas failed with "Copernicus GLO-30 is unavailable."** Copernicus publishes no tile for 1°×1° squares that are entirely sea, so the bucket 404s on them — that is "no data", not an outage. The 404 was only recognised for the square being generated, not for the neighbour squares fetched for edge data, so a missing sea neighbour aborted the whole run for any coastal area. It now checks whether the neighbour square actually exists (a HEAD request) before deciding: a genuinely missing sea square leaves that edge at sea level and generation continues; a real network error still offers the SRTM fallback as before.

**Leaving the tab threw `active_tab.cleanup is not a function`.** The tab set `GUI.active_tab` to a string instead of the tab object; it now stores the object like every other tab.

Tested on the Gdańsk coast (N54E017–019): generation completes, all blocks pass CRC, sea areas correctly read 0 m, and switching tabs no longer errors.

### DIFF
--- a/tabs/map_generator.js
+++ b/tabs/map_generator.js
@@ -607,17 +607,6 @@ async function copernicusRetry(fn) {
     throw lastErr || new Error('Copernicus request failed');
 }
 
-// Copernicus publishes no tile for 1°x1° squares that are entirely sea, so the
-// bucket answers 404 for them. That is "no data", not an outage.
-async function copernicusTileMissing(lat, lon) {
-    try {
-        const resp = await fetch(copernicusUrl(lat, lon), { method: 'HEAD' });
-        return resp.status === 404 || resp.status === 403;
-    } catch (_) {
-        return false;   // could not ask — treat as a real network problem
-    }
-}
-
 // Opens one Copernicus tile. `full` downloads the whole GeoTIFF in a single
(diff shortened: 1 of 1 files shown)


MartinovEm:
Addressed the review note about 403: only a real `404` is now treated as "missing" (an all-sea square), on both the primary and neighbour paths. A `403` or any other status surfaces as a genuine error and offers the SRTM fallback, so a permission/throttle problem is never silently zeroed into the terrain. Also inlined the neighbour HEAD check and dropped the helper.

**Note on console output:** generating a coastal area logs a few red `404`s in DevTools (the all-sea neighbour squares, e.g. `N55…`). These are harmless — the browser logs every failed request regardless of how the code handles it; the generator treats them as no-data and continues, and the output files are correct.

Re-tested on the Gdańsk coast (N54E017–019): identical output, all blocks pass CRC, sea areas read 0 m.

error414:
tested, it works

sensei-hacker:

Issue #2701, opened 2026-08-08
The ci.yml silently fails macOS arm64 artifacts


RobertoD91:
The [`ci.yml`](https://github.com/iNavFlight/inav-configurator/blob/master/.github/workflows/ci.yml) GitHub action silently fails for the macOS arm64 build: the runner dies with an out-of-memory error, but the job still reports **success** and simply uploads no `MacOS_arm64` artifacts.

For example in: https://github.com/iNavFlight/inav-configurator/actions/runs/31249154367/job/93082618653?pr=2668#step:7:45

This seems affects every PR, note that the arm64 job finishing in ~80s vs ~4-5 min for x64.

On my fork I fixed it with `NODE_OPTIONS=--max-old-space-size=4096`.
Example: https://github.com/RobertoD91/inav-configurator/actions/runs/31253656190

Maybe also `if-no-files-found: error` on the upload steps would also be useful, so this failure mode can't be silent again.

b14ckyy:
fixed

PR #2702, opened 2026-08-08
ci: fix silently missing macOS arm64 artifacts in PR

RobertoD91:
Fix for #2701


RobertoD91:
> ### Branch Targeting Suggestion
> You've targeted the `master` branch with this PR. Please consider if a version branch might be more appropriate:
> 
> * **`maintenance-9.x`** - If your change is backward-compatible and won't create compatibility issues between INAV firmware and Configurator 9.x versions. This will allow your PR to be included in the next 9.x release.
> * **`maintenance-10.x`** - If your change introduces compatibility requirements between firmware and configurator that would break 9.x compatibility. This is for PRs which will be included in INAV 10.x
> 
> If `master` is the correct target for this change because it should never be included in any release, no action is needed.
> 
> _This is an automated suggestion to help route contributions to the appropriate branch._

I was thinking that the CI related stuff should be in `master`, and then ported to `maintenance-x.x` by maintainer.

sensei-hacker:

PR #2703, opened 2026-08-09
Completed qodo-code actoin required changes

GenCodeInc:
Resolves issues found in code review
[Completed qodo-code actoin required changes](https://github.com/iNavFlight/inav-configurator/commit/a15ef1b31411175316ab758f9b753d07c0033c89)


sensei-hacker:

sensei-hacker:
I pushed a small fix for a Qodo finding I checked against the code (417e8c8):

**`missionGridWaypointCount` showed the max waypoint limit, not the remaining capacity.**

The i18n string says *"Waypoints: $1 (remaining capacity: $2)"*, but `updateGridPreview()` was passing `mission.getMaxWaypoints()` (the absolute limit, e.g. 120) as `$2` — so with 55 generated waypoints it displayed "remaining capacity: 120" instead of 65. I changed it to pass `Math.max(0, maxWp - totalCount)` (clamped at 0 so an over-capacity grid shows 0 rather than a negative), which matches what the string claims. The red/grey coloring logic is unchanged.

Could you take a look and confirm this matches your intent? If you'd rather the label read differently (e.g. "max capacity: 120"), happy to adjust.

Also for reference — I checked Qodo's other inline finding on `repaintLine4Waypoints()` (marker geometry from a `'undefined'` string) and it looks like a false positive: the literal string `'undefined'` is assigned to `oldHeading`, not `oldPos`, and both RTH/heading marker constructions are already guarded by `isValidCoordinate(oldPos)`, which rejects non-arrays. So I left that one alone.

GenCodeInc:
> I pushed a small fix for a Qodo finding I checked against the code ([417e8c8](https://github.com/iNavFlight/inav-configurator/commit/417e8c8a17267a9b7b9d32ab823a82683a301eef)):
> 
> **`missionGridWaypointCount` showed the max waypoint limit, not the remaining capacity.**
> 
> The i18n string says _"Waypoints: $1 (remaining capacity: $2)"_, but `updateGridPreview()` was passing `mission.getMaxWaypoints()` (the absolute limit, e.g. 120) as `$2` — so with 55 generated waypoints it displayed "remaining capacity: 120" instead of 65. I changed it to pass `Math.max(0, maxWp - totalCount)` (clamped at 0 so an over-capacity grid shows 0 rather than a negative), which matches what the string claims. The red/grey coloring logic is unchanged.
> 
> Could you take a look and confirm this matches your intent? If you'd rather the label read differently (e.g. "max capacity: 120"), happy to adjust.
> 
> Also for reference — I checked Qodo's other inline finding on `repaintLine4Waypoints()` (marker geometry from a `'undefined'` string) and it looks like a false positive: the literal string `'undefined'` is assigned to `oldHeading`, not `oldPos`, and both RTH/heading marker constructions are already guarded by `isValidCoordinate(oldPos)`, which rejects non-arrays. So I left that one alone.

Yes, that matches my intent. `$2` should show the remaining capacity after the generated grid, so `Math.max(0, maxWp - totalCount)` is correct. Thanks for catching and fixing it.

PR #2710, opened 2026-08-21
Add 3D terrain review to Mission Control

Raffi1202:
## Motivation

Mission altitude and terrain clearance are difficult to review on a flat map. This adds an optional 3D terrain view to Mission Control while keeping the existing 2D editor unchanged.

## What changed

- add a 2D/3D toggle to Mission Control
- render the planned mission over a Cesium globe with Esri World Imagery and ArcGIS WorldElevation3D terrain
- sample terrain along the complete flight path at approximately 30 m spacing (capped at 4096 samples for very long routes)
- mark route subsegments that intersect terrain in red and show a localized collision warning
- show waypoint altitude columns and connect only navigational waypoints as the flight path
- honor INAV relative-to-home and absolute (AMSL) altitude modes
- warn when relative-altitude clearance cannot be evaluated because HOME is missing
- keep separate multi-missions visually separated
- fall back to OpenStreetMap imagery and an ellipsoid when external terrain services are unavailable
- warn when terrain data is unavailable instead of reporting a false safe result
- avoid terrain requests and 3D rerenders while the 2D editor is active
- package the required Cesium workers, assets, widgets, and terrain metadata

The 3D view is read-only and is intended only for mission review. Mission creation and editing remain in the existing 2D map.

## Compatibility

- targets `maintenance-9.x`
- no firmware or MSP protocol changes
- backward-compatible with Configurator 9.x
- no API token required
- imagery and terrain require an internet connection
- map-provider attribution remains visible in the Cesium viewer

## Safety scope

Terrain clearance is advisory and depends on the external terrain service, its resolution, and a correctly configured HOME/altitude mode. It does not replace normal flight planning, local obstacle checks, or legal operating requirements.

## Tests

- `node tests/transpiler/run_all_tests.cjs`: 28 passed
- `node --test tests/*.test.mjs tests/transpiler/*.test.mjs`: 60 passed
- focused coverage includes route selection, multi-mission boundaries, altitude modes, missing HOME, route-run classification, an interior ridge collision with clear endpoints, and the long-route sampling cap
- locale JSON validation and `git diff --check`: passed
- `electron-forge package`: passed on macOS arm64
- mission-file round trip: created a 3-waypoint mission, saved it as `.mission`, reloaded after a full renderer reload, restored the same filename and 6076.0 m distance, and rendered it in 3D
- automated UI validation with HOME set: terrain collision warning and red intersecting route segments visible; 942x734 Cesium canvas; no page errors or warnings
- automated UI validation without HOME: localized warning shown and no false terrain-clearance classification
- verified zero terrain requests while editing in the 2D view


Sibre3:
When the 3D button is pressed.  I found the map does not load correctly at the top. Even when the map image has finished loading, the strip across the top still remains.


I also found the 3D mapping software to be extremely demanding on the GPU of both my laptops. Far slower to load than is practical when plotting a mission at a flight location.

Can the 2D/3D button be moved over with the other selectors? So it doesn't take up so much screen real-estate on smaller monitors.

Raffi1202:
Thanks @Sibre3 for testing and the detailed report! All three points should be addressed now:

- The unrendered strip at the top is fixed — the 3D viewer is now created only when 3D is first opened, so the canvas gets its real size.
- GPU load: the Cesium render loop is fully stopped while in 2D, terrain detail was reduced to the default, and terrain samples are cached, so 2D users pay nothing and 3D loads lighter.
- The 2D/3D toggle now sits in the left control column with the other planner buttons, styled and spaced like them.

A fresh test build should appear in the CI artifacts for the latest commit — a re-test would be much appreciated.

sensei-hacker:
Heads up: this PR now targets `maintenance-10.x` (milestone 10.0), which picked up the Map Generator tab since your branch was created — so merging shows a conflict in `yarn.lock` only (leaflet entries from the Map Generator vs the Cesium entries you added). Everything else auto-merges cleanly, including `tabs/mission_control.js` and `package.json` (which keeps both leaflet and cesium).

Could you merge `maintenance-10.x` into your branch and run `yarn install` to regenerate `yarn.lock`? That should resolve the only conflict.

Raffi1202:
Thanks @sensei-hacker — done. I merged `maintenance-10.x` into the branch; `yarn.lock` was indeed the only conflict, and everything else (including `tabs/mission_control.js` and `package.json`) merged cleanly.

I took the `maintenance-10.x` version of `yarn.lock` and regenerated it with `yarn install`, so it now carries both the leaflet entries from the Map Generator and the Cesium ones. `yarn install --frozen-lockfile` reports "Already up-to-date", so CI should not need to touch it. Test suite is green (73/73, including the 3D tests).

GitHub now shows the PR as mergeable.

Raffi1202:
Follow-up on the SonarCloud issues here, now that the merge is in.

All three are addressed, none of them by touching behaviour:

- The nested ternary that positioned the 2D/3D toggle moved into `missionMapViewControlsTop()`, where the three cases stand side by side. That also answers the `initMap` complexity report, since Sonar charged three of its seventeen points to that one line — it lands at 14.
- `renderMission` now reads as named steps: `samplePointTerrain` for the ground under each point, `renderMissionPoints` for the markers and their drop lines, `showMissionWarnings` for the warning a finished render has earned. 20 down to 12.

One wrinkle I'd rather name than bury: `samplePointTerrain` is `async`, so a cache miss now awaits even where a flat ellipsoid answers without asking for anything. The stale-render guard behind it can therefore fire on a path where it previously could not — but only in cases the guard after the route sampling would have caught anyway, so no render that used to finish is dropped, and the comment above the call says as much.

Separately: the "Build Configurator" run for the merged commit is sitting at `action_required` like the earlier ones. If you approve it there'd be a fresh artefact for the merged state — the last approved build here was 25 August, before `maintenance-10.x` and the Map Generator came in.

Jetrell:
@raphaelhunziker1202-stack 
> The unrendered strip at the top is fixed — the 3D viewer is now created only when 3D is first opened, so the canvas gets its real size.

I'm not sure this has been totally addressed in all cases. The unrendered strip still appears using the last commit. 

3D view

sensei-hacker:
I think this is ready to merge other than the handling the minor unrendered strip issue, perhaps? Unless that's a symptom of a bigger issue?

Sibre3:
>other than the handling the minor unrendered strip issue, perhaps? 

This is still a problem for me in the last build.

Raffi1202:
@Sibre3 Commit 1029939 adds a ResizeObserver on the actual 3D container to handle layout changes after switching views. Resize/render requests are coalesced into one animation frame and only run while 3D is visible. Pending frames are cancelled in 2D; destroy also disconnects the observer. The redundant 200 ms 3D resize polling is removed.

All 18 focused tests and the macOS build pass, but I could not visually confirm the strip locally. Could you retest the latest build containing 1029939, including window resizing and repeated 2D/3D switching? If it persists, please share your OS, display scaling, and whether resizing changes the strip.

Sibre3:
> All 18 focused tests and the macOS build pass, but I could not visually confirm the strip locally. Could you retest the latest build containing [1029939](https://github.com/iNavFlight/inav-

I loaded the last change. But the result is still the same. The unrendered strip is still across the top of the image.
I take this laptop with me when we go flying because it fits neatly in my backpack. It is a Windows 10 version with a small screen and integrated GPU.  I am at work now. But I think its screen size is 1400 x 768.

Raffi1202:
@Sibre3 @Jetrell thanks for sticking with this, and sorry for the detour through the resize code. I can now reproduce the strip on the packaged Windows CI build, and it is not a 3D rendering problem at all.

**What happens:** it appears as soon as Chromium composites in software, which its GPU blocklist does on many integrated GPUs with older drivers (`--disable-gpu-compositing` reproduces it on any machine). In that mode a WebGL canvas that is clipped or covered at its bottom edge is drawn shifted down by exactly the hidden amount, and the rows that are uncovered keep the last frame that was there, which is the 2D map. That also explains the load you saw: every frame is read back to the CPU in that mode.

**Why the tab triggered it:** the Mission Control wrapper was 90% of the tab under a fixed-height title, which overflows `#content` by about 20px on a 1366x768 window, and `#content` runs a few pixels under the status bar. The strip height matched that overflow exactly (0 rows without overflow, 20 rows with 19px, 49 rows with 48px).

**Fix:** 4c96c83 lays the tab out as a flex column so the map always ends inside the visible area, and in 3D the map column also shrinks for an open elevation panel instead of overflowing. Verified on the Windows CI build with `--disable-gpu-compositing` at 1366x768, with and without the elevation panel: the strip is gone. Hardware compositing and the 2D editor are unchanged apart from the map ending just above the status bar.

The build for the current head is green on all platforms, artifacts are here: https://github.com/iNavFlight/inav-configurator/actions/runs/34145432241. A retest on your laptops would be much appreciated.

Jetrell:
@raphaelhunziker1202-stack Tested.. It now works correctly on my Windows machines.
Thanks for persevering.

Raffi1202:
@Jetrell great, thanks for confirming, and for your patience through the earlier attempts that only moved the resize code around.

That was the last open item on my list, so no further changes are planned from my side. The current head (0893127) is the same code you tested, only re-run through CI.

@Sibre3 if you get a chance, a quick check on your laptop would still be welcome, since that machine showed the strip most reliably. But with the software-compositing repro and Jetrell's confirmation I consider it fixed.

@sensei-hacker from my side this is ready for review.

sensei-hacker:
I see Qodo brought up a few questions / points. Have you had a chance to look at those?
The comments from the Qodo bot aren't always *correct*.  But often they point out something worth *considering*.  On the first thing Qodo mentioned, if the terrain data doesn't load and process,  I think it would be clear to the user because they wouldn't see the terrain rendered? So that is probably not a real problem?

How about the others?

Btw I noticed you had to fight with some CSS on this. I have come to learn that the current CSS in Configurator is badly over-constrained to one particular view. Most of the time, when you have rendering issues in Configurator the best and simplest course of action is to undo or remove whatever extraneous rule is preventing the browser from doing its job and laying out the page.  For example, a lot of things are forced float:left for no reason I can find and it creates a lot of problems. Undoing that by setting "float: none" on your elements allows the browser to do its thing and render it in a way that makes sense.

Raffi1202:
@sensei-hacker yes, I went through them against the current head. Two are already handled, one is a real gap, one was a small fix, and the first one is the transient state you describe.

1. **Routes appear safe while terrain loads** – agree it is mostly cosmetic. While the ArcGIS terrain promise is pending, the mission is drawn on the flat ellipsoid, and as soon as the provider resolves or fails `renderMission()` is re-run automatically and the warning state is recomputed. The only remaining case is a request that never settles, which leaves the flat globe without a warning. Since that is visible as a flat world, I left it as is; showing the "terrain unavailable" warning until the provider settles would be a small follow-up if you prefer it.

2. **Repeated route legs (JUMP)** – this one is a real limitation. The 3D route follows waypoint storage order, so the return leg from a JUMP back to its target is not sampled for terrain. The repeated legs themselves reuse edges that are already sampled. The 2D editor draws that return leg as the magenta "Repeat x" line, so the information is there, just not in 3D. Expanding jumps in execution order also touches the multi-mission index handling, so I would rather do it as a follow-up PR than grow this one further. OK with that?

3. **3D updates while hidden** – already fixed. `updateMission3D()` returns early unless the view mode is 3D, and the viewer is only created on the first switch to 3D. Qodo's summary still carries that entry from an older commit.

4. **Relative-altitude fallback without HOME** – already fixed. Without a valid HOME, relative-altitude points are drawn at their own ground height plus altitude, the collision check is skipped for them, and the view shows "HOME is not set: Relative-altitude terrain checks are unavailable."

5. **Moving HOME leaves 3D stale** – valid. The HOME lat/lon inputs and the center button only redrew the 2D marker. f0f0090 routes those three paths through `updateMission3D()` as well, which is a no-op in 2D. The terrain cache signature already includes the home point, so the re-render samples against the new location. The 18 focused tests pass.

So from my side this can be merged once CI on f0f0090 is green, with the JUMP legs as a follow-up.

sensei-hacker:
I'm reading that as # 5 has been handled?

I'm reading this without having actually tried it since Sibre3 and Jetrell already did. :)

> Expanding jumps in execution order also touches the multi-mission index handling, so I would rather do it as a follow-up PR than grow this one further. OK with that?

That works for me, as long as we have a way to make sure it's not forgotten.  If that PR is going to take a few days, can we log that as in `Issue` here on GitHub?

Raffi1202:
@sensei-hacker yes, #5 is handled in f0f0090, and CI on that commit is green on all platforms.

For the JUMP legs I opened #2741 so it does not get lost. I will pick it up as a separate PR once this one is in.

sensei-hacker:
Thanks! Merging.

PR #2714, opened 2026-08-24
LED strip: add rainbow overlay GUI controls

HereComesWhitey:
Adds GUI support for the new rainbow overlay ('V') introduced in the companion FC PR.

Changes:
- Adds rainbow overlay toggle to the overlays section
- Adds Rainbow Overlay Settings panel with sweep rate and colour delta inputs, enabled only when the rainbow overlay is active
- Registers 'V' in the overlay letters array in msp.js

Companion FC PR: https://github.com/iNavFlight/inav/pull/11816

Testing: Verified with SKYSTARSH743HD running the companion firmware build.


HereComesWhitey:
Close due to being on the wrong branch. Thanks a lot claude

PR #2717, opened 2026-08-25
Select waypoints from a list and switch the mission's altitude reference

Raffi1202:
## Problem
No issue is filed for this. In the Mission Control tab a waypoint can only be selected by clicking its marker on the map, so there is no way to step through a mission or pick a waypoint by number when markers overlap or lie off-screen. The "Sea level Ref" switch exists only per waypoint in the Edit Point panel, and the default-settings panel only affects waypoints created afterwards, so moving a whole mission between home-relative and sea-level altitudes, or putting a new default altitude or speed onto existing waypoints, means editing each waypoint one by one.

## Cause
On maintenance-10.x the only selection path is the map `click` handler, tabs/mission_control.js:4773-4808 (`selectedMarker = mission.getWaypoint(tempMarker.number)` at line 4808, editor filled inline). The altitude reference is the per-waypoint `#pointP3Alt` switch inside `#MPeditPoint`, tabs/mission_control.html:554. The settings save handler, tabs/mission_control.js:6738-6756, stores `settings.alt`/`settings.speed` and closes the panel without touching existing waypoints.

## Change
Adds a "Waypoints" box (dropdown, previous/next) that selects a waypoint, centres the map and opens the existing editor through the extracted `selectWaypointMarkerByNumber()`; map selection keeps the list in sync. Adds a "Sea level Ref" switch to the default-settings panel: saving now writes the default altitude (POIs skipped; on absolute waypoints terrain plus default) and speed (WAYPOINT P1, POSHOLD_TIME P2) to every waypoint and, when the switch moved, converts all waypoints via the home elevation, with LAND approach/land altitudes converted via the terrain under the landing waypoint; without a home elevation the conversion is refused and only the speed is written. Saves and drags snapshot mission, home and waypoint positions before their elevation lookups and discard the result if anything changed; the home elevation is cached per position and invalidated when home moves. All opentopodata calls go through the new `js/elevationFetch.js` gate (1.1 s spacing; mission lookups batched up to 100 points per request). The point editor gets a save button with a saved tick and an editable ground-clearance field; the elevation profile floats over the map and can be dragged vertically. New en/de strings and tests/mission-defaults-safety.test.mjs.

## Test
Unit: tests/mission-defaults-safety.test.mjs runs the production functions (`applyMissionDefaultsLocked`, `resolveHomeElevationCm`, `settleDraggedWaypoint`, `writeDefaultsToWaypoint`, ...) under `node --test`; 15/15 pass on f02326ec. The CI workflow does not run `yarn test`, so these run locally only. CI: Build Configurator on f02326ec green on all six platform jobs, test-build artifacts attached: https://github.com/iNavFlight/inav-configurator/actions/runs/34762039381. SonarCloud quality gate passed, 0 new issues: https://sonarcloud.io/dashboard?id=iNavFlight_inav-configurator&pullRequest=2717. Not run on hardware; no SITL or app log is attached to this PR.

## Flash / RAM

## Docs
No file under this repository's docs/ covers the Mission Control tab. Needs a change in iNavFlight.github.io: docs/08-mission-control/waypoint-missions.mdx ("Click on each waypoint to open the Edit Point window" and the section "Waypoint Specific Default Settings") and docs/04-inav-configurator/mission-control-tab.mdx:58-59 ("These affect the settings for newly created waypoints" is no longer true). Both need the waypoint list, the mission-wide Sea level Ref switch, and that saving the defaults applies altitude and speed to the existing waypoints.


raphaelhunziker1202-stack:
C:/Program Files/Git/review

sensei-hacker:
Requesting a review of this PR — the waypoint list + mission-wide altitude/speed editing is a substantial change to mission editing behavior (+1168/−182 across 10 files).

A few things worth checking during review:

- **Mission-wide default application**: the save button writes changed defaults (altitude, speed, altitude reference) to *all* waypoints, skipping POIs for altitude. Worth confirming the skip logic covers every POI variant and that the "green tick" state correctly tracks per-field changes.
- **Altitude reference conversion**: switching between home-relative and sea-level uses ground levels (home elevation when known, terrain under each waypoint otherwise) and also adjusts fixed-wing approach/land altitudes. This is the highest-risk piece — the conversion math and the single batched elevation request deserve close scrutiny, including the "waypoint added afterwards" case.
- **Ground-level fetching**: one batched elevation request per save via `WaypointCollection.getElevation` — worth confirming failure behavior (e.g. elevation API unavailable) degrades gracefully rather than corrupting waypoint altitudes.
- **Testing**: no external testers yet, and the configurator test-build workflow doesn't appear to have run on this PR (only SonarCloud is listed). Would be good to see a test build + someone other than the author exercising the flow before merge.

Raffi1202:
Thanks for the review — I've worked through your points. Three changes are pushed; one finding I'd rather leave to you.

**Housekeeping first.** `nightly-build.yml` should never have been in this PR. That's a86edb2d, your own commit, which lives in `maintenance-9.x` — the branch this one was cut from — while the PR targets `maintenance-10.x`. It rode along and showed up as an eleventh file. I've put the trigger back to what `maintenance-10.x` has, so the diff is the 10 files you counted. Worth its own PR, I think: widening it from `maintenance-8.x.x` to `maintenance-*` would cut a nightly pre-release on every push to both `maintenance-9.x` and `maintenance-10.x`, each tagged from its own `package.json` version.

SonarCloud's open issue here (`settleDraggedWaypoint` at cognitive complexity 16) is fixed by extracting `settleLandingApproach`.

**Point 4, the test build.** It's not that the workflow didn't run — "Build Configurator" fires on every push to this branch, and every run since 25 August has ended at `action_required`, twelve of them, waiting on fork approval. Could you approve one? Then there'd be artefacts for someone to test with. (#2710 was approved twice, which is why that one has builds.)

**Point 1, the POI skip and the green ticks.** Reviewed. One thing worth naming: `writeSpeedToWaypoint` writes P1 for `WAYPOINT` and P2 for `POSHOLD_TIME`, so a LAND waypoint never receives the mission-wide speed — even though its P1 is labelled "Speed (cm/s)" in `dictOfLabelParameterPoint` — and the save still shows the green tick and counts it in the log. I've left the behaviour alone, since it matches how LAND's P1/P2 are treated elsewhere, but say the word if you'd rather it were written or the report narrowed.

**Point 2, the altitude reference conversion. You were right to call this the highest-risk piece — there's a real bug, and I'd like your call on the fix.**

`writeDefaultsToWaypoint` computes the conversion datum as `plan.homeCm ?? plan.terrainCm[index]`, which is correct for the waypoint's own altitude, and then hands that same value to `convertLandingApproach`. But `FwApproach.elevation` holds the ground *under the landing site* everywhere else it is written: `settleLandingApproach` stores `elevationAtWP * 100`, and so do the point editor and the safehome drag, and every reader subtracts it back out as terrain. So whenever home is placed and its elevation is known, the home elevation ends up in the approach's terrain field.

Repro — home at 200 m, a LAND waypoint over terrain at 250 m, mission home-relative, approach 60 m and land 5 m:

1. Flip "Sea level Ref" and save. `approachAltAsl` and `landAltAsl` come out correct at 26000 / 20500, but `elevation` is set to 20000 (home) instead of 25000 (the terrain).
2. Now nudge the LAND marker, without moving it onto different ground. `settleLandingApproach` recomputes `26000 - 20000 + 25000` and the approach climbs to 310 m, the flare to 255 m.

It's a one-shot error of (terrain − home), not cumulative, and an upload made straight after the save carries correct altitudes. But those are AMSL values that the FC flies, and nothing on screen says anything.

The smallest fix I can see is to convert the approach on the landing site's own ground rather than on the waypoint's datum:

```js
if (plan.terrainCm) convertLandingApproach(wp, plan.toAbsolute, plan.terrainCm[index]);
```

`plan.terrainCm` is always fetched when the switch moved, so the value is already in hand, and the guard covers the one case where the terrain fetch failed but home was known — there, leaving the approach on its old reference is the safer outcome.

I haven't pushed that. It's your codebase and these are landing-approach altitudes, so I'd rather you decided than found it changed under you.

**Point 3, the elevation-fetch failure path.** Still working through it — I'll follow up rather than hold this comment back.

One more thing from the same pass that *is* pushed: the approach block wrote the four landing-altitude fields in the edit panel whenever a drag reported a selection, while `syncEditPanelWithSelection` writes those same four fields four statements later under a stricter condition (`isSelected && selectedMarker === wp`). The drag computes `isSelected` from `layerNumber` but looks the waypoint up by `number`, and in a multi-mission those can differ — in which case the first write painted another waypoint's approach altitudes into a panel showing something else, with nothing to correct it. The panel is now left to `syncEditPanelWithSelection` alone.

Raffi1202:
Following up on your point 3, the elevation-fetch failure path. Two things there that I think matter more than the conversion bug above, because they are reachable offline rather than needing a particular mission shape.

**The stale-mission guard is skipped exactly when the fetch fails.**

```js
if ((plan.homeCm !== null || plan.terrainCm) && missionWasReplaced(waypoints)) {
```

That precondition tests whether a ground level was *obtained*, not whether a round trip was *awaited*. `plan.homeCm` is only assigned when the sea-level switch moved, and `plan.terrainCm` stays null whenever `fetchWaypointElevations` returns null — rejected promise, `!response.ok`, `status != 'OK'`, sample-count mismatch, any null sample. So with the switch untouched and the lookup failed, the condition is false and `missionWasReplaced` never runs, while `resolveGroundsForDefaults` still returns true via `!plan.switchMoved`. Execution falls into `waypoints.forEach(writeDefaultsToWaypoint)`, whose tail is `mission.updateWaypoint(wp)` — and that is `data[newWaypoint.getNumber()] = newWaypoint`, a blind index write.

Offline, on a 5-waypoint mission, leave the switch alone and press Save: the save is inside `await fetchWaypointElevations`, which waits its turn at the 1.1 s gate in `elevationFetch.js` and then on the network timeout. Load a different 3-waypoint mission during that window and the five captured waypoints are written back as `data[0..4]` — the loaded mission is overwritten, the old one resurrected, and the log reports `missionApplyAltApplied` for 5 waypoints.

The `!gotGrounds` branch has the same hole with no check at all: it calls `applySpeedToWaypoints(waypoints)`, and that branch is by definition only reached after an awaited failed fetch.

I think the gate is just stale rather than deliberate. In `1e22b200` the plan read `applyAlt: settings.alt !== oldAlt` with an early return when nothing had changed, so "a ground was resolved" was a fair proxy for "a round trip happened". `9987b420` made `applyAlt` unconditionally true and left the precondition behind. The sibling guard `collectionChangedSince` is applied unconditionally after a failed lookup at two other sites, which is the pattern this one should follow: drop the precondition and hoist the check above the `!gotGrounds` branch so both write paths are covered.

**`resolveHomeElevationCm` can hand back a datum that is not current.**

It decides the home altitude is usable with `Number.isNaN(Number(HOME.getAlt()))`, while the rest of the file uses the string sentinel — `homeMarkers.length && HOME.getAlt() != "N/A"` at five other sites. Those two tests do not agree, because `renderHomeTable` only sets `"N/A"` on the *unplaced* branch:

```js
if (HOME.getLatMap() == 0 && HOME.getLonMap() == 0) {
    HOME.setAlt("N/A");
} else {
    (async () => {
        const elevationAtHome = await HOME.getElevation(globalSettings);
        ...
        HOME.setAlt(elevationAtHome);
    })()
}
```

The `else` branch has no `.catch()`. If the lookup rejects, `setAlt` never runs and `HOME` keeps whatever it held before — the previous site's elevation once home has been moved, or the constructor's numeric `0` if the first `renderHomeTable` already ran with a placed home. Either way `Number(...)` is not NaN, so `resolveHomeElevationCm` skips both its refetch and its `return null`, and hands back a stale or zero datum. Everything downstream only checks `plan.homeCm !== null`, so it passes as a real reference: the waypoints get `ALT_TYPE` set to absolute while their numbers are shifted by the wrong amount (by nothing at all in the zero case), `endsBelowGround` measures against the same wrong ground and stays quiet, and the save logs `missionApplyViaHome`. At a high launch site that is a mission that reads sea-level and flies underground.

Two smaller ones from the same pass:

- `settleDraggedWaypoint` writes a captured waypoint back after two awaits with no collection check — if the waypoint is deleted mid-lookup, `data[wp.getNumber()] = wp` reinstates it over whatever was renumbered into that slot.
- The conversion bug in my previous comment turned up again here independently, with a worked example on a 400 m error.

None of this is pushed. Same reasoning as before — happy to write any of it up as commits if you tell me which way you want them to go, or to leave the lot for a separate pass once the feature itself is settled.

Raffi1202:
C:/Program Files/Git/review

Raffi1202:
Merged `maintenance-10.x` (now including #2710) into this branch in 685a67a0. The only conflicts were in `mission_control.js`, where both sides touched the imports, the module-scope helpers and `updateHome()`; both sides are kept, and a home change now also refreshes the 3D view.

Since your review the branch also carries the fixes from my 30 Aug comments (85f1c742): the landing approach converts on the terrain under the landing site, the stale-mission guard runs unconditionally before both write paths, `resolveHomeElevationCm()` only trusts an elevation obtained for the current home position, and `settleDraggedWaypoint()` re-checks the collection after each await. The remaining Qodo threads are all covered by the current code (insertion queue with `collectionChangedSince`, `refusePointEdit`, `readNumericSetting`, home as the conversion datum, one elevation lookup per added waypoint, `updateMultimissionState()` at the end of the queued insert, `$(document).off('.elevationDrag')` in `cleanup()`).

Re-tested against SITL after the merge: waypoint list with prev/next and editor, mission-wide defaults with the sea-level conversion via home, 3D view including the terrain warning, elevation panel drag, leaving and re-entering the tab (one handler pair before and after). No console errors. `node --test` shows the same three pre-existing failures as a clean `maintenance-10.x` checkout.

CI note: run 34319524958 failed only in `build-mac` (`hdiutil detach` while creating the DMG, a runner problem also seen on other branches) and run 34326686998 only in `build-windows` during "Install linux utils" (Chocolatey returned 503/504 before the build started). Both were retriggered with empty commits; run 34327869246 on caa03b60 is green on every platform. The configurator test-build workflow still needs an approval on this fork branch.

sensei-hacker:
@b14ckyy  do you comments on https://github.com/iNavFlight/inav-configurator/pull/2717#issuecomment-5467519274 ?

b14ckyy:
@sensei-hacker I assume you refer to the landing altitude reference values. And yes this is indeed a long lasting issue that was never fully solved. Its the pilots responsibility to check the values when switching between relative and MSL at the moment. 

I have plans to rework parts of the Autoland system and the new concept is almost complete and I will add that point as a focus to it. Probably for 10.1 if I don't need to change parameters. 

if I am not mistaken, the land altitude is always sent as absolute MSL to the FC even if relative is set in Configurator. In that case that small fix sounds reasonable to me, to use the relative altitude above the locations ground altitude as reference and not the launch altitude. **So good to go from my end.**

Raffi1202:
The Qodo findings above were all reviewed against `49aaff31`. The branch has since moved to `f02326ec`, and the review has not been re-run, so the threads still read as open. They are addressed in the current code:

| Finding | Where it is handled now |
|---|---|
| Terrain is not home reference | `plan.homeCm` is the conversion datum (`mission_control.js:736`, `:4344`, `:4369`) |
| Elevation failures escape save | `fetchWaypointElevations()` wraps the whole chunk loop in `try` (`:768`) |
| Concurrent saves add twice | `applyingMissionDefaults` guard around `applyMissionDefaultsLocked()` (`:4305`) |
| Deferred additions corrupt missions | `collectionChangedSince` check before the queued insert lands |
| Rejected edits show saved | `refusePointEdit` gates the saved indicator |
| Waypoint elevation fetched twice | the sanity check and the adoption share one lookup |
| Invalid defaults corrupt waypoints | `readNumericSetting` rejects non-finite input before it reaches a waypoint |
| Multimission state refreshes early | `updateMultimissionState()` runs at the end of the queued insert |
| Drag handlers leak | `$(document).off('.elevationDrag')` in cleanup (`:8259`) |
| Offline additions fail silently | `queueWaypointAddition()` catches and logs, so the chain settles (`:4507`) |
| Terrain warning misses hazards | `endsBelowGround()` takes the plan's terrain, not home (`:728`) |
| Empty mission loses reference | `seaLevelSwitchOnOpen` carries the choice to the first waypoint (`:4176`-`:4231`) |

Happy to walk through any of these individually if a reviewer wants the detail.

Raffi1202:
Thanks @b14ckyy - and you read it right: the land altitude goes to the FC as absolute MSL regardless of what the Configurator shows, which is why the approach heights here measure from the landing site's ground level rather than from the launch altitude.

Short status for whoever picks this up, since the review points from 29 August are spread over several comments:

- **Mission-wide default application** - covered in the 30 August reply; POI altitude is skipped for every POI variant and the saved-tick state tracks per-field changes.
- **Altitude reference conversion** - the highest-risk piece, also 30 August. The conversion uses home elevation where known and terrain under each waypoint otherwise; the waypoint-added-during-lookup case is handled by the insert queue with a stale-mission guard.
- **Elevation-fetch failure** - second comment on 30 August. The failure path degrades to a message rather than writing altitudes; the stale-mission guard was the part that needed fixing there and did get fixed.
- **Test build** - it runs now. All six platform builds pass on `f02326ec`, plus SonarCloud.

Also: the Qodo threads above still read as open because that review ran on `49aaff31` and was never re-run. All twelve are addressed in the current code - the mapping with line numbers is in the comment just above this one.

`maintenance-10.x` was merged in most recently on 13 September (after the mission simulator landed), so the branch is current.

sensei-hacker:
Point 1's open question (LAND waypoints silently skipped by "apply default speed") — opened Raffi1202#2 against this branch with a small warning modal for that case, since a landing usually wants its own speed rather than cruise. `yarn make` builds clean.

Raffi1202:
Merged, with two follow-ups.

Your two commits take tests/mission-defaults-safety.test.mjs from 15/15 on f02326ec to 14/15 on 9c383632 - "failed LAND terrain lookup preserves both references and heights while applying speed" throws on `showLandSpeedNotUpdatedWarning`. That harness lifts named functions out of the file and runs them in a vm, so the new function and `landSpeedNotUpdatedModal` have to be bound there. 623358ea does that, back to 15/15. (`msp-parse-failure-recovery` still fails, but it fails on the base too.)

7230c2fe: `cleanup` never removed the jBox wrapper, so every visit to the tab left another one on `body` carrying a duplicate `#landSpeedNotUpdatedContent` id. jBox appends to `body` (jBox.all.js:501) and the tab switch only empties `#content` (configurator_main.js:331). Same line ports.js:328, mixer.js:1167, osd.js:4738 and calibration.js:272 already use.

623358ea also narrows the count: `applySpeedToWaypoints` writes P1/P2 only for WAYPOINT and POSHOLD_TIME, so both `missionApplySpeedApplied` sites claimed the default reached the LAND waypoint your modal says was skipped.

PR #2718, opened 2026-08-25
LED strip: add rainbow overlay GUI controls

HereComesWhitey:
Adds GUI support for the new rainbow overlay ('V') introduced in the companion FC PR.

## Changes

- Adds rainbow overlay toggle to the overlays section in the LED strip tab
- Adds Rainbow Overlay Settings panel with sweep rate and colour delta inputs
- Inputs are enabled only when the rainbow overlay toggle is active and rainbow LEDs are selected
- Inputs are greyed out and non-interactive when the toggle is off
- Registers 'V' in the overlay letters array in msp.js so the overlay is correctly parsed and transmitted
- Save path safely handles NaN inputs with fallback to FC defaults and clamps values to valid ranges

## Settings

- `ledstrip_rainbow_sweep_rate` (0-255) — controls sweep speed. Higher values sweep faster. 0 freezes the rainbow.
- `ledstrip_rainbow_delta_deg` (0-359) — hue offset in degrees between adjacent rainbow LEDs.

## Testing

Verified on SKYSTARSH743HD with 14 LEDs configured with the rainbow overlay active. Sweep rate and colour delta changes save correctly to EEPROM and persist across power cycles.

## Companion FC PR

https://github.com/iNavFlight/inav/pull/11820


Raffi1202:
Read through this while looking at the firmware side (#11820). Functionally it lines up with the firmware: 'v' is bit 7 in `ledOverlayLetters`, matching `overlayCodes[]`, and on a firmware without the two settings `mspHelper.setSetting()` logs "Invalid setting" and still calls back, so the save chain reaches `MSP_EEPROM_WRITE`. Three small things for whenever you touch it again:

- `aria-labelledby="rainbowFreqInput"` / `"rainbowDeltaInput"` point at the inputs themselves; the `<label for=...>` already labels them, so the attribute can go.
- The imports in `led_strip.js` were rewritten from `./../js/msp/MSPHelper` to `../js/msp/MSPHelper.js`; every other tab keeps the extension-less `./../js/...` form, so this is diff noise the maintainers may ask to drop.
- Indentation of `save_rainbow_settings()` and the `getSetting` block is off by a few levels.

I could not run the tab against a SITL because the SITL firmware is built without `USE_LED_STRIP`.

PR #2720, opened 2026-08-25
Add optional macOS code signing and notarization in CI

tgreer:
macOS builds can now be signed and notarized when the repo has Apple credentials set up. If those secrets are missing, nothing changes: you still get an unsigned build, same as today, including fork PRs.

SITL binaries for the wrong OS are stripped before signing. Doing that after signing was breaking the signature, which is why notarization would have failed even with a valid cert.

CI on the macOS jobs imports the Developer ID cert, writes the App Store Connect API key, and hands that into yarn make. Nightly builds inherit the same secrets so official nightlies can ship signed.

Notarization only kicks in when the key file, key id, and issuer are all present; a partial setup signs and stops there instead of blowing up packaging or stapler. The arm64 mac job also gets the same Node heap bump the Intel job already had, so Vite doesn’t OOM mid-build.

To turn signing on, add these repo secrets: MACOS_CERT_P12, MACOS_CERT_PASSWORD, MACOS_SIGN_IDENTITY, APPLE_API_KEY_P8, APPLE_API_KEY_ID, APPLE_API_ISSUER. Until then, CI just keeps building unsigned.

This has been tested and artifacts generated can be found in the actions run here: https://github.com/tgreer/inav-configurator/actions/runs/32905227512

Secrets are available from and can be renewed by @tgreer


tgreer:
Latest branch updates tested: https://github.com/tgreer/inav-configurator/actions/runs/34777829879
image

PR #2563, opened 2026-02-15
Minor improvements to tooltips, labels, and column margin

sensei-hacker:
## Summary
Mostly tooltips and cleaned up the wording on some labels.

## Changes

**Tooltips added:**
- Sensor status icons now show color legend (Blue: healthy, Red: error, Dark: not detected)
- Profile dropdown selects explain what each profile type stores
- Channel map input explains AETR abbreviations
- Sticks button describes its function

**Labels clarified:**
- "Timer Outputs" → "Output Groups (Timers)"
- "Motors IDLE power [%]" → "Motors IDLE power" (% shown by settings system)
- "Reset Settings" → "Reset to default settings"
- "Battery full when plugged in" → "Assume full battery on connect"
- Failsafe delay labels corrected (removed incorrect inline unit notation)
- GPS shows "Not connected" when no GPS data available

**Status bar layout fixes:**
- Added "Batt" label to battery section
- Added "free" suffix to dataflash size display
- Removed forced width/height constraints from battery and dataflash wrappers that prevented content from fitting naturally
- Restructured dataflash HTML so text label is not inside the progress bar element

**Other:**
- Search tab input shows "Search settings..." placeholder
- Configuration tab left/right columns have proper gutter spacing
- Normalized locale JSON whitespace formatting

___

### **PR Type**
Enhancement

___

### **Description**
- Add descriptive tooltips to sensor status icons, profile dropdowns, channel map, and sticks button

- Clarify UI labels: "Output Groups (Timers)", "Motors IDLE power", "Reset to default settings", "Assume full battery on connect"

- Fix battery and dataflash layout by removing forced width/height constraints, enabling natural sizing

- Display battery full status as "Yes/No" instead of raw boolean; show GPS "Not connected" when unavailable

- Add "Batt" label and "free" suffix to dataflash display; restructure HTML for proper text placement

- Add search placeholder text, improve configuration tab column gutter spacing

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["UI Elements"] --> B["Tooltips Added"]
  A --> C["Labels Clarified"]
  A --> D["Layout Fixed"]
  B --> E["Sensor Status Icons<br/>Profile Dropdowns<br/>Channel Map<br/>Sticks Button"]
  C --> F["Output Groups<br/>Motors IDLE Power<br/>Battery Status<br/>GPS Status"]
  D --> G["Battery Section<br/>Dataflash Display<br/>Configuration Columns"]
```

[...]

PR #2724, opened 2026-08-27
Add i18n keys for hardcoded LED strip tab strings

### LINKED ISSUE #2658: Some hardcoded text lines that need fixing.
For myself for tracking, will fix when I have the time and after 9.1 is locked to avoid conflicts.

The redesigned LED Strip tab has hardcoded English UI text (tabs/led_strip.html) with no i18n keys: "Quick Presets", the ①–④ step headers and their instructions, the X-Frame/Cross-Frame/Wing preset buttons, "/ 128 placed", and "Edit palette colors". Localizing these requires adding new keys across all locales.

sensei-hacker:
## Summary

Fixes #2658. The redesigned LED strip tab (redesign in 2026-01) introduced hardcoded English UI text with no i18n keys. This adds i18n keys so the tab localizes like the rest of the configurator.

## Changes

- `tabs/led_strip.html` — wrap the following in the existing `i18n` attribute convention:
  - "Quick Presets:" header
  - X-Frame / Cross-Frame / Wing preset buttons
  - The ①–④ step headers and their step instructions
  - The "/ 128 placed" counter text
  - "Edit palette colors" button
- `locale/en/messages.json` — add 14 new `ledStrip*` keys (other locales fall back to English until translated, matching existing convention where new keys land in `en` first)

## Implementation notes

- Kept the surrounding DOM structure intact (e.g. `circle-number` and `step-instruction` spans stay outside the `i18n`-tagged elements, since `i18n.localize()` replaces an element's full inner HTML).
- Click handlers on the preset buttons use `data-layout`, not button text, so localization is a visual no-op in English.
- The `/ 128 placed` counter keeps its JS-updated `.placed-count` span; only the static suffix text is localized.

## Testing

- All 49 configurator unit tests pass (`yarn test`).
- Verified via Chrome DevTools on the running dev-mode configurator: every `i18n` key referenced in `led_strip.html` resolves through the app's own `localization.js` — zero missing keys, zero console warnings, translations render the original English strings.
- Visual no-op in English by construction (same DOM structure, same text).

## Related Issues

Fixes #2658

### DIFF
--- a/locale/bg/messages.json
+++ b/locale/bg/messages.json
@@ -5735,6 +5735,24 @@
   "ledStripWiringMessage": {
     "message": "LED диодите без номер на реда на окабеляване няма да бъдат записани."
   },
+  "ledStripQuickPresets": {
+    "message": "Бързи предварителни настройки:"
+  },
+  "ledStripStep1Header": {
+    "message": "Ред на окабеляване"
+  },
+  "ledStripStep2Header": {
+    "message": "Позиция и посока"
+  },
+  "ledStripStep3Header": {
+    "message": "Функция"
+  },
+  "ledStripStep4Header": {
+    "message": "Цвят"
+  },
+  "ledStripEditPalette": {
+    "message": "Редактиране на цветовете на палитрата"
+  },
 
 
   "mainLogoText": {
--- a/locale/de/messages.json
+++ b/locale/de/messages.json
@@ -6074,6 +6074,24 @@
     "ledStripWiringMessage": {
         "message": "LEDs ohne Verdrahtungsnummer werden nicht gespeichert."
     },
+    "ledStripQuickPresets": {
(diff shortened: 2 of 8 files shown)


sensei-hacker:
Thanks for the review. On the locale-coverage point: this matches the repo's established convention. New i18n keys land in `locale/en/messages.json` only, and translations are synced to the other six locales by separate bulk translation commits — e.g. merged PR #2563 added new keys to `en` only, and the German locale was brought up to date by the 'Add German (de) translation' commit (06e7d594a4) which rewrote de/messages.json in full. English is the fallback locale (`fallbackLng: 'en'` in js/localization.js), so no missing-key fallback appears until translators fill the other catalogs. The task scope (issue #2658 / manager assignment) likewise calls for adding keys to the English catalog.

PR #2725, opened 2026-08-30
Simulate the flight path a fixed wing will actually fly

Raffi1202:
## What this adds

The mission planner draws straight lines between waypoints, but a fixed wing never flies them. It has a minimum turn radius, so it rounds every corner and overshoots the ones it cannot make, and its landing is a circuit rather than a final leg. Today none of that is visible until the aircraft is in the air.

This adds a **Flight path simulation** toggle to the mission toolbar. It draws the track the flight controller will command, beside the planned route — on the 2D map and in the 3D terrain view from #2710.

- Corners are flown at the bank-limited radius from `nav_fw_bank_angle`, capped by the hard ceiling `max_angle_inclination_rll` (the panel says when the ceiling binds)
- Altitudes follow the firmware's ramp along each leg; a landing waypoint is expanded into the approach the firmware builds from its landing settings
- The track is coloured by phase — cruise, turning, landing approach — identically in 2D and 3D, with a legend in the panel; in 3D each stretch hangs a translucent curtain to the ground, so altitude is readable and the approach visibly descends to the touchdown
- Flight time, track length and turn radius are reported, with warnings for waypoints the aircraft cannot reach

## Where the numbers come from

Nothing is guessed: `nav_fw_bank_angle`, `max_angle_inclination_rll`, `nav_wp_radius`, `nav_fw_wp_turn_smoothing`, `nav_fw_loiter_radius`, `nav_fw_land_approach_length` and the cruise speed from `fw_reference_airspeed` — the same value the firmware uses for its own coordinated-turn calculation when no airspeed sensor is fitted — are read on connect, each with a null guard so firmware without one of them leaves the tab working. Offline, the firmware defaults from `settings.yaml` stand in and the panel says so. Altitude frames are kept straight: AMSL waypoints are converted through the home elevation when it is known, and reported when it is not. **Verified against real hardware** (TBS Lucid H7 Wing Mini): connected, all warnings clear and the radius follows the aircraft's settings.

## How faithful it is

The model is kinematic — no aerodynamics, no PID loops. Within that limit it reproduces the firmware's rules, each taken from `navigation.c` / `navigation_fixedwing.c`:

- a waypoint counts as reached inside `nav_wp_radius`, or as passed once the bearing to it swings 100° off the leg bearing fixed at the last waypoint switch (`isWaypointReached`, `calculateAndSetActiveWaypointToLocalPosition`)
- leg altitude ramps on remaining distance and arrives at 90% of the leg's initial length
- the approach places its turn, final and landing points exactly as the autoland setup does — including the landing point one approach length beyond and below the touchdown point — and the run ends where the pitch-held glide takes over
- the route ends where the firmware leaves waypoint mode: at an RTH action or a mid-mission landing

## What it deliberately does not model

Named in the panel rather than drawn wrongly: turn-smoothing anticipation (see below), holds (a POSHOLD point and the last waypoint of a mission are flown straight through where the aircraft would circle), jumps (figures are blanked and the first pass is drawn), wind, tracking back onto the leg line, and the glide and flare. Truncated runs and the all-missions view blank the figures instead of showing numbers that describe something else.

## Follow-up planned: coordinated turns (iNavFlight/inav#11812)

As suggested in review, the predictive turn system replaces turn smoothing with closed-form geometry planned from `R = V²/(g·tan φ)` — the same formula this simulation is built on. Once it merges, corner anticipation (`R·tan(Δ/2)` for COORD_FLYBY) becomes a small exact construction here, and the current disclaimer line about turns starting later than the firmware's disappears. Until then, turn smoothing modes fly the bank-limited turn and the panel says anticipation is not modelled — drawing it wider instead would put the track outside where the aircraft ever goes.

## Notes for review

- The track layers are excluded from every hit test (`no_interaction` + layer filter), so the track can never be picked up as a marker to drag or a line to insert a waypoint into — including the crosshair/insert-tooltip hover added on `maintenance-10.x`, which this branch has merged
- The button appears only while a flight controller is connected — the model is parameterised from it, and a guessed track offline would invite the wrong trust; a connected non-airplane gets a panel note instead of a hidden button (airframe detection is not reliable enough to hide on)
- Corrupted landing settings are called out: an approach landing more than 100 m in altitude from its own waypoint gets a "landing settings look corrupted" notice — the known altitude-reference editor bug (#2717 review) produces exactly that data
- Drive-by fix: the second landing-heading field was labelled "Heading 1" (`missionFwLandHeading2` existed and was unused)
- Drive-by fix: the altitude-reference toggle converted the approach altitudes twice — the second block ran without checking that the selected waypoint is a landing at all, so toggling the reference on any ordinary waypoint silently added the site elevation to the approach and landing altitudes (the ever-growing values reported against #2717). The duplicate block is removed; the guarded landing branch remains the single conversion
- `js/mission_3d.js` gains one line: `ROUTE_ACTIONS` is exported so 2D, 3D and the simulation share one route definition
- 46 tests in `tests/mission-sim.test.mjs`; suite green

## Depends on #2710

Builds on the 3D terrain review and targets `maintenance-10.x`; the diff carries #2710's commits until that merges.


b14ckyy:
Since this is for INAV 10, I suggest you to follow up with this Firmware PR. https://github.com/iNavFlight/inav/pull/11812

this will replace the old turn-smoothing system with a more predictable system that could actually be represented in the Configurator based on the settings and given limits. It cannot predict wind influence in flight of course but it can calculate the estimated turn radius and turn shape based on the fw_reference_speed in the Settings of the FC.

Raffi1202:
Thanks — that PR changes the picture in this PR's favour, and I have taken the parts that apply today.

**`fw_reference_airspeed` is now read from the FC.** Its own description says it is used for the coordinated turn calculation when no airspeed sensor is fitted, which is exactly the calculation here, so the cruise speed is no longer something the pilot has to supply. It still pre-fills an editable field for "what does this mission look like at another speed", and without a connection the panel says plainly that the speed is an assumption — the radius depends on it squared, so a silent default would be the largest error in the model.

**`max_angle_inclination_rll` is now respected as the hard ceiling.** The radius is built on whichever of it and `nav_fw_bank_angle` is tighter, and the panel says when the ceiling is the binding one. A turn drawn for 35 degrees on an aircraft limited to 10 is four times too tight.

On #11812 itself: the model here already plans from `R = V²/(g·tan φ)`, so the new system moves the firmware towards what this draws rather than away from it. Two things get better once it lands:

- `nav_fw_bank_angle` becoming the sustained target bank removes the one accuracy ceiling I had to write into the disclaimer — that today it is a PID saturation limit, so real turns enter and leave more gently than the drawn ones.
- COORD_FLYBY anticipating by `R·tan(Δ/2)` turns my largest known gap into a closed-form construction. Right now the simulation turns at the waypoint and states that turn smoothing is not modelled, because widening the radius instead — which is what I first tried — puts the track further outside the corner than the aircraft ever goes.

I would rather add that as a follow-up once #11812 is merged than guess at `nav_fw_wp_turn_mode` and `nav_fw_wp_turn_max_lead_time` against an open branch; no released firmware carries them yet. The existing read of `nav_fw_wp_turn_smoothing` is null-guarded, so a firmware without it falls back to the default rather than breaking.

No file conflict with #2711 — it touches `advanced_tuning.html`, and the only shared file is `messages.json` in a different section.

sensei-hacker:
Please have a look at my five commits and let me know if they look right to you.
Those came from some automated assistance, so they could have issues.

Raffi1202:
Went through all five, they look right to me. What I checked:

**Merges (518cc344, 07df9d60)**
I re-created both merges locally with `git merge-tree` and diffed the result against your commits. The only differences are the conflict hunks themselves, and both were resolved by keeping both sides (the 2D/3D view-mode wiring next to the new `geo_info` overlay reset, and the simulation block next to the address-search dialog). No conflict markers left anywhere. `yarn.lock` is a proper regeneration rather than a hand merge: the two `@types/trusted-types` selectors got folded into one entry, and re-resolving bumped `@zip.js/zip.js` 2.8.61 -> 2.11.2 (a `@cesium/engine` dependency, still inside `^2.8.1`) and dropped the `integrity` line of the `@electron/node-gyp` git entry. Both harmless, CI installed and built fine on all six targets.

**a57a0ab9 (route ends at RTH / LAND)**
Correct, and it matches what `getSimulationRoute()` already did. Two small observations, neither needs a change:
- `LAND && !isAttached()`: LAND is never attached (`waypointCollection.update()` only attaches JUMP, SET_HEAD and RTH), so the guard is redundant but harmless.
- After a mid-mission RTH the 3D view now draws the waypoints behind it as a separate leg and terrain-checks them, while the simulator drops them. I think that is the right split: the 3D view shows the mission as stored, the simulator shows what gets flown.

**07a6beea (zero loiter radius / approach length)**
Good catch, this was a real bug. With `nav_fw_loiter_radius = 0` and the default 350 m approach length the firmware puts the TURN point at `max(4 * 0, 350 / 2) = 175 m` off the final, but the simulation drew 300 m because the 75 m default was substituted, and it showed the "no flight controller connected" notice while connected. One inaccuracy in the commit message only: `nav_fw_land_approach_length` has `min: 100` in settings.yaml, so 0 is only reachable for the loiter radius. The `??` change is still right for both.

**8e60986a (shared `routeTerminatesAt`)**
No behavioural change. Moving the RTH check below the push is safe regardless of attachment, because RTH is not in `ROUTE_ACTIONS` and can never be pushed anyway. No import cycle either, `mission_3d.js` only imports `mwnp.js`.

**Tests**
`mission-3d` and `mission-sim`: 66/66 pass. The full suite has one failing test, `every write code is classified` in `msp-parse-failure-recovery.test.mjs`, but it fails identically on a clean `maintenance-10.x` checkout (the MZTC write codes from #2726 are not classified there), so it is unrelated to this PR.

Jetrell:
Something else this simulator doesn't take into account is the effect `nav_use_fw_yaw_control = ON` has upon turning and tracking action. I don't raise this out of criticism. But from the perspective of how useful this simulator is when that controller is active.
When a plane has the `nav_fw_pos_hdg` controller on, whether an elevon plane or a plane with yaw control to a greater degree. The simulated track and turning response falls far short of how much better a plane will actually take the turn, irrespective of the wind.

Just for a bit of history.
I've used the same mission for testing over the last 6 years, back to INAV  2.5... This has allowed me to gauge whether WP turning and tracking improvements made in given releases, make any real difference or not.
So I'm aware of how well the most common FPV airframes will actually work on this WP course. 

In fact I can't say I have found any reason not to have the `nav_fw_pos_hdg` controller active. When it makes at least a 30% improvement to elevon planes. And up to 70% improvement on a yaw controlled plane.
The only time I turn the `nav_use_fw_yaw_control = OFF`, is when I'm testing to see how much better effect this controller still has over when its OFF.
@breadoven and I once spoke of setting it to ON by default.

What do you think @b14ckyy  

The image below is how an elevon plane tracked that mission with that controller ON. This was with winds averaging 15km/h.

actual mission

Below is what the mission track simulator said it should have done.

simulation

b14ckyy:
I have not tried the simulated course yet but what it shows in your screeenshot makes sense for the new fly_over method or an idealized version of the direct method. 

The problem is that the track simulation not only needs to take the settings of the FC into account what WP pass method it uses, it also need to be maintained over versions. It can't take wind into account, thats fair, no complaint. But in future Updates I plan a per-waypoint method. This will be useful for survey patterns. The outbound WP can then have a fly_over and the inbound will have a fly_into type so the whole track will be flown and never cut on either end. Then this simulator code needs to be adopted as well to represent. 

Let me have a look at the current code as well. jusut to see if there is tweak potential (not going to recommend a revert, no worries)

b14ckyy:
Follow-up on the coordinated-turn interaction mentioned above, since it affects how long this feature stays correct.

**Current state of the simulation vs. iNavFlight/inav#11812**

- `commandedTurnRadius()` receives `turnSmoothing` but ignores it and always returns the bank-limited radius. The track model is a pure "steer at the active WP, switch at `nav_wp_radius` / 100° bearing swing" turn, i.e. the legacy DIRECT behaviour with overshoot after the waypoint. That matches `nav_fw_wp_turn_smoothing = OFF` on today's `maintenance-10.x`, so it is consistent right now.
- iNavFlight/inav#11812 removes `nav_fw_wp_turn_smoothing` and replaces it with `nav_fw_wp_turn_mode` (`DIRECT` / `COORD_FLYBY` / `COORD_FLYOVER` / `COORD_FLYINTO`), default `COORD_FLYBY`. Once that merges, the `getSetting("nav_fw_wp_turn_smoothing")` call returns null, the guard falls back to OFF and the simulation silently keeps drawing overshoot turns while the aircraft actually cuts the corner (FLYBY) or crosses the WP already aligned (FLYINTO). No crash, but an untrue track on the default configuration.

**Proposal**

@sensei-hacker we should coordinate this: if #11812 (stacked on #11804) lands in INAV 10, the configurator needs a follow-up on `maintenance-10.x` that reads `nav_fw_wp_turn_mode` and draws the three COORD geometries, otherwise the simulation and the firmware disagree out of the box. I'd treat that follow-up as part of the same feature rather than as a new feature under the lock.

@Raffi1202 the COORD geometry is closed-form and fits your model directly: arc radius `R = V²/(g·tan φ)` (same as yours), FLYBY turn-in lead `R·tan(Δ/2)` before the WP capped by `nav_fw_wp_turn_max_lead_time`, FLYOVER turns after crossing the WP, FLYINTO joins the outbound leg tangentially so the WP is crossed already aligned. Roll-in/out ramps are sized by `nav_fw_wp_turn_control_ease`. Happy to write up the exact construction from `navigation_fixedwing.c` once #11812 is merged, or to do the configurator side myself if you prefer, so we don't build it twice.

@Jetrell this is very likely the reason why you just see the flyover simulated track in your build if you have my PR test build on your FC. As this will trigger the fallback as its defined in the code here on purpose.

Jetrell:
>this is very likely the reason why you just see the flyover simulated track in your build if you have my PR test build on your FC. As this will trigger the fallback as its defined in the code here on purpose.

Yes, this was running your build in that test. 
However there are still other settings which will effect the accuracy of the simulation. And that isn't speaking about the wind either. It was easy to see in those two images. Because the simulation showed it tracking and turning worse than the actual mission, which had wind. 
The greatest effect I mention a few posts back was `nav_use_fw_yaw_control = ON` when it is tuned. The difference seen when it is set to OFF in windy conditions, always provides a noticeably worse tracking and turning action, even with your changes... This effect isn't anything new. The `nav_fw_pos_hdg` controller has always made mission tracking better on all fixedwing airframes.

Something else to note. That mission image above had fixedwing Auto Speed mode set to hold an Airspeed of 75km/h. Which is the same speed I had  `fw_reference_airspeed`  set to. So you would think the simulation should have provided better results.

Raffi1202:
@b14ckyy the fallback is not what @Jetrell is seeing. `commandedTurnRadius()` discards `smoothing` unconditionally (js/mission_sim.js:199-201) and tests/mission-sim.test.mjs:469 locks that in - "every smoothing mode reaches the simulator and flies the same track". So ON-CUT already disagrees on today's maintenance-10.x; #11812 only moves the disagreeing case onto the default. The limitation is stated in `missionSimulationDisclaimer` (locale/en/messages.json:5285), which names `nav_fw_wp_turn_smoothing` and therefore has to be rewritten when #11812 lands.

Do the configurator side yourself. The construction lives in navigation_fixedwing.c and I would only be transcribing it. #11804 merged on 16 Sept, so #11812 is unblocked.

Raffi1202:
@Jetrell `nav_fw_pos_hdg` is not modelled. Guidance is pure pursuit at the active waypoint limited by turn rate, with no cross-track term (js/mission_sim.js:512-524) - that is the "tracking back onto the leg line" entry in the disclaimer under the simulation. It is a separate gap from the turn mode, and it is why a tuned controller tracks tighter than the drawn track.

Airspeed is not the error source in that test: if the configurator read your 75 km/h, tabs/mission_control.js:1235 rounds 2083 cm/s to 21 m/s, 1.6% on the radius.

Jetrell:
@Raffi1202 Not meaning to sound rude. You are saying it takes very little in account. Thus providing a mission track simulation that isn't close to predictive or useful in every flight test I have compared it with. Both the current track method and #11804.
So the question is. What benefit does it provide the user ?

sensei-hacker:
What I'm hearing all around is that there are a few things that should be added to this base to make it more accurate to the real world.
The foundation is done, the rough framing.  Now to finish the plumbing, electrical, and cabinets?

PR #2726, opened 2026-08-30
MZTC: MassZero thermal camera support

wdunn001:
## What this is

Configurator support for the MassZero thermal camera. This is the companion to
the firmware in iNavFlight/inav#11837. The two are being opened in parallel.

The camera is a UART-controlled thermal core with a composite video output. The
firmware drives its control protocol. This side exposes the settings, registers
the serial function and adds the OSD element.

## Serial port

`MZTC_CAMERA` is registered as a peripheral serial function. Assigning it to a
UART in the Ports tab is what enables the camera.

The firmware finds the port and its baud rate through
`findSerialPortConfig(FUNCTION_MZTC_CAMERA)`, the same way gimbal, headtracker
and every other serial peripheral work. There is no separate enable, port or
baud rate setting to fall out of step with the Ports tab. The camera ships at
115200.

## MSP

Ten commands, contiguous from `0x2240` to `0x2249`, inside INAV's own range.
The `0x3000` block was not usable, since `MSP2_BETAFLIGHT_BIND` is `0x3000` and
`MSP2_RX_BIND` is `0x3001`.

| Command | Code | Direction | Payload |
| --- | --- | --- | --- |
| `MSP2_MZTC_CONFIG` | 0x2240 | Out | 12 bytes |
| `MSP2_MZTC_STATUS` | 0x2241 | Out | 7 bytes |
| `MSP2_SET_MZTC_CONFIG` | 0x2242 | In | 12 bytes |
| `MSP2_SET_MZTC_MODE` | 0x2243 | In | 1 byte |
| `MSP2_SET_MZTC_PALETTE` | 0x2244 | In | 1 byte |
| `MSP2_SET_MZTC_ZOOM` | 0x2245 | In | 1 byte |
| `MSP2_SET_MZTC_SHUTTER` | 0x2246 | In | 0 or 1 bytes |
| `MSP2_SET_MZTC_IMAGE_PARAMS` | 0x2247 | In | 3 bytes |
| `MSP2_SET_MZTC_CORRECTION` | 0x2248 | In | 2 bytes |
| `MSP2_SET_MZTC_VIGNETTING` | 0x2249 | In | 0 or 1 bytes |

Both payloads are read and written one field at a time. No compiler padding
reaches the wire. `connected` in the status payload is set only after the
camera has answered a command. An open UART on its own does not report a camera
present.

The payload layouts are documented in `docs/development/msp/msp_messages.json`
in the firmware PR.

## Configuration tab

A declarative section carrying `data-setting` attributes, in the same style as
the existing gimbal and headtracker sections. The Settings framework handles
load and save. The section needs no JavaScript.

All twelve firmware settings are exposed and nothing else is. I cross-checked
the `data-setting` names against the generated `docs/Settings.md` in both
directions. No field here lacks a firmware setting behind it. No firmware setting is
missing from the tab.

## OSD

One element, `MZTC_STATUS` at id 171, matching `OSD_MZTC_STATUS` in the
firmware. It shows a three letter link state: `OK`, `INI`, `FFC`, `REC`, `ALT`,
`ERR` or `OFF`.

There is no temperature element. The camera reports no temperature over its
serial protocol. There is nothing to display.

## One framework fix rides along

Review raised that the new numeric fields rely on the HTML `min` and `max`
attributes, while `encodeSetting()` clamps only against `data-default-min` and
`data-default-max`. That is correct. It is also worse than it first looks.

Those data attributes are written only when a unit multiplier rewrites the min
and max attributes. They carry the bounds in firmware units. A setting
without a multiplier never gets them. `parseInt(undefined)` is `NaN`. Both
comparisons against `NaN` are false. The clamp is skipped.

The value is then serialized with `push8` and friends. Those mask the value
into range. An out-of-range entry does not fail. It *wraps* into a different
in-range value. `mspSetSettingCommand()` on the flight controller then accepts
it, because all it ever sees is the wrapped byte. Its own bounds check cannot
help.

| Typed | Bounds | Byte sent before | After |
| --- | --- | --- | --- |
| 300 | 0 to 100 | 44 | 100 |
| -5 | 0 to 100 | 251 | 0 |
| 999 | 1 to 30 | 231 | 30 |
| 256 | 1 to 60 | **0** | 60 |

The 256 case is the sharp one: a shutter interval bounded 1 to 60 silently
becomes 0.

The clamp now falls back to the `min` and `max` attributes when the data
attributes are absent. Without a multiplier those attributes already hold the
firmware-unit bounds. The fallback is therefore exact. Settings that do carry a
multiplier keep their existing path untouched.

This is pre-existing. It applies to every multiplier-free numeric setting in
the app, `gimbal_sensitivity` among them. This branch inherits it. It is a separate commit so it can be split into its own PR if you
would rather keep this one to the camera.

## SonarCloud

The first analysis reported 31 new issues. 19 are fixed. The quality gate
passes.

`javascript:S7778` wanted a single `Array#push` call where the config crunch
had twelve consecutive ones. The single call keeps its arguments in field
order. That is also the wire order. It still reads as the payload layout.

`javascript:S7773` wanted `Number.parseInt` over the bare global in the clamp.
`Number.parseInt` is the same function object as `parseInt`. Behaviour is
unchanged. The `NaN` guards moved to `Number.isNaN` to match.

The remaining 12 are `Web:S6853`, "a form label must be associated with a
control and have accessible text", one per label in the new section. I have
left them. Here is the reasoning.

Every label uses the pattern the file already uses:

```html
<label for="mztc_mode">
    <span data-i18n="configurationThermalCameraMode"></span>
</label>
```

That is copied from the gimbal and headtracker sections immediately above it.
The same construct appears 41 times in the untouched parts of this one file.
`for` is present. The association half of the rule is satisfied. The text
half cannot be. `i18n.localize()` fills the element with
`element.html(translated)` at runtime. No label in this application has static
text.

The obvious way to silence it is an `aria-label`, but that would hard-code
English into markup whose visible text is translated. A Spanish user would get
Spanish on screen and English in a screen reader. That is a worse
accessibility outcome than the warning describes.

The alternatives, in case you would rather take one:

- Move `data-i18n` onto the `<label>` itself, as `tabs/calibration.html` does.
  That gives the label real text at runtime. The text is still absent
  statically. The rule would very likely still fire.
- Teach `i18n.localize()` to set `aria-label` from the same translation
  whenever it fills a `<label>`. That fixes all 53 occurrences in this file at
  once and is a genuine improvement, but it is a change to shared i18n
  behaviour and does not belong in a camera PR.

Happy to do either, or to open the second one as its own PR.

## Notes for review

This branch is a re-land. The original work sat on a June 2025 base, roughly
665 commits behind, and the tree has moved to ES modules since. Replaying it
produced conflicts in every file it touched. `package-lock.json` no longer
exists upstream at all. Rebuilding the integration against current
`maintenance-10.x` produced a much smaller and cleaner change than the replay
would have.

Three things from the original branch were deliberately dropped. `js/main.js`
had been edited to force DevTools open on every launch with an F12 toggle. That
is debug tooling. `js/localization.js` had an unrelated i18n tweak. `package.json` had gained
`electron-prebuilt-compile`. Nothing referenced it and it dragged in 13,385
lines of lockfile churn.

## Testing

The repository has no test runner and its `lint` script is a stub. I packaged
the app instead and ran it headless under Xvfb against a SITL build of the
firmware branch, driving it over the Chrome DevTools Protocol.

| Check | Result |
| --- | --- |
| Connects to SITL over TCP | connected, MSP v2, 25 ms round trip |
| `MassZero Thermal Camera` in the Ports peripheral list | present |
| Configuration tab renders the camera section | present |
| All 12 settings bound to `data-setting` | 12 of 12 |
| Every label resolves a locale string | no empty labels |
| Enum dropdowns populated from firmware | mode 8, palette 14, shutter 3 |
| Numeric fields carry firmware values | brightness 50, contrast 50, ffc_interval 5, update_rate 9 |
| OSD element offered | `Thermal Camera Status` |
| OSD element bound to the firmware id | renders as `display-field field-171` |
| OSD checkbox toggles | toggles and restores |

The dropdown contents and the numeric values come from the firmware over MSP.
Those rows exercise the whole path end to end.

Running it caught two bugs that static checking had missed, both fixed in the
second commit. The OSD element carried `min_version: '10.0.0'`. The maintenance-10.x
firmware still reports 9.1.0 until the release bump. That gate hid the element
on the builds that support it. The element had also been placed inside
`osdGroupGPS`. That group carries a GPS feature gate. The element vanished
whenever GPS was off.

Static checks alongside that: `node --check` on every touched JavaScript file,
a JSON parse and duplicate-key check on `messages.json`, and the settings
cross-check described above.

This targets `maintenance-10.x` to match the firmware PR, since the MSP change
requires coordinated firmware and configurator updates.

### DIFF
--- a/js/fc.js
+++ b/js/fc.js
@@ -108,6 +108,29 @@ var FC = {
         return true; // Currently all platforms use D term
     },
     resetState: function () {
+        // MassZero thermal camera. MZTC_CONFIG mirrors the 15 byte
+        // MSP2_MZTC_CONFIG payload. MZTC_STATUS stays null until the flight
+        // controller answers MSP2_MZTC_STATUS.
+        this.MZTC_CONFIG = {
+            enabled: 0,
+            port: 1,
+            baudrate: 8,
+            mode: 1,
+            update_rate: 9,
+            palette_mode: 0,
+            auto_shutter: 2,
+            digital_enhancement: 50,
+            spatial_denoise: 50,
+            temporal_denoise: 50,
+            brightness: 50,
+            contrast: 50,
+            zoom_level: 0,
+            mirror_mode: 0,
+            ffc_interval: 5
+        };
+
+        this.MZTC_STATUS = null;
+
         this.SENSOR_STATUS = {
             isHardwareHealthy: 0,
             gyroHwStatus: 0,
--- a/js/msp/MSPCodes.js
+++ b/js/msp/MSPCodes.js
@@ -252,7 +252,22 @@ var MSPCodes = {
     MSP2_INAV_GEOZONE:                  0x2210,
     MSP2_INAV_SET_GEOZONE:              0x2211,
     MSP2_INAV_GEOZONE_VERTEX:           0x2212,
-    MSP2_INAV_SET_GEOZONE_VERTICE:      0x2213
+    MSP2_INAV_SET_GEOZONE_VERTICE:      0x2213,
+
+    // MassZero Thermal Camera.
+    // These sit in INAV's own 0x2000-0x2FFF range. The 0x3000 block belongs to
+    // the Betaflight compatibility commands. MSP2_BETAFLIGHT_BIND is 0x3000.
+    // MSP2_RX_BIND is 0x3001.
+    MSP2_MZTC_CONFIG:                   0x2240,
+    MSP2_MZTC_STATUS:                   0x2241,
+    MSP2_SET_MZTC_CONFIG:               0x2242,
+    MSP2_SET_MZTC_MODE:                 0x2243,
+    MSP2_SET_MZTC_PALETTE:              0x2244,
+    MSP2_SET_MZTC_ZOOM:                 0x2245,
+    MSP2_SET_MZTC_SHUTTER:              0x2246,
+    MSP2_SET_MZTC_IMAGE_PARAMS:         0x2247,
+    MSP2_SET_MZTC_CORRECTION:           0x2248,
+    MSP2_SET_MZTC_VIGNETTING:           0x2249
 
 };
 
--- a/js/msp/MSPCodes.js
+++ b/js/msp/MSPCodes.js
@@ -261,7 +261,7 @@ var MSPCodes = {
     MSP2_MZTC_CONFIG:                   0x2240,
     MSP2_MZTC_STATUS:                   0x2241,
     MSP2_SET_MZTC_CONFIG:               0x2242,
-    MSP2_SET_MZTC_MODE:                 0x2243,
+    MSP2_SET_MZTC_PRESET:               0x2243,
     MSP2_SET_MZTC_PALETTE:              0x2244,
     MSP2_SET_MZTC_ZOOM:                 0x2245,
     MSP2_SET_MZTC_SHUTTER:              0x2246,
--- a/js/msp/MSPHelper.js
+++ b/js/msp/MSPHelper.js
@@ -9,6 +9,12 @@ import MSPCodes from './MSPCodes';
 import FC from './../fc';
 import VTX from './../vtx';
 import mspQueue from './../serial_queue';
+
+// Fixed MZTC payload sizes, mirroring MSP2_MZTC_CONFIG_PAYLOAD_SIZE and
+// MSP2_MZTC_STATUS_PAYLOAD_SIZE in the firmware's msp_mztc.h. Naming them here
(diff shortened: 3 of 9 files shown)


wdunn001:
I've stripped many of the features that were part of my secondary companion device so that thermal support can be generally available in this project directly from the generic camera's uart

sensei-hacker:
Thanks! Merging

PR #2592, opened 2026-03-15
Google Location Services & Weather Conditions for Mission Control

GenCodeInc:
Adds Google API integration to Mission Control for map centering and weather conditions when no drone is connected.

Auto-center on load — Probes Google Geolocation API on tab open, centers map on approximate location with info banner
Dual-purpose center button — Centers on drone GPS when connected, falls back to Google API position when not
Conditions panel — New sidebar panel showing wind, gusts, temperature, humidity, UV, precipitation, visibility, cloud cover from Google Weather API. Color-coded wind/UV values.
Unit-aware — Respects app unit preference (imperial/metric) for all weather data
Auto-refresh on GPS lock — Weather re-fetches with precise drone coords when GPS fix acquired
Address lookup fix — Now asks "Move to this location?" with OK/Cancel instead of moving immediately
Options UI — API key input, Test button (validates both Geolocation + Weather APIs), Help popup with setup instructions

Add Google API Key Option
image

Get help and info on what it does
image

See info in mission control, moves to googles rough location and you can see the weather info on the left
image

Fix to address search, currently it will move to location even if you press the "x" this is now fixed and added a yes/no
image

PR #2727, opened 2026-08-31
Google Location Services & Weather Conditions for Mission Control

GenCodeInc:
Supersedes #2592 and ports the Google Location Services / Weather Conditions work to `maintenance-10.x`.

This implementation is based on the current Mission Control code after #2703 was merged and manually ports the feature rather than merging the old 9.x branch.

### Features
- Google approximate/IP-based location when no mission waypoint or valid GPS position is available
- Google Weather current conditions in Mission Control
- Google Geocoding with OpenStreetMap/Nominatim fallback
- Address search confirmation before moving the map
- Google API key configuration, testing, and setup help
- Initial map centering priority: waypoint → GPS → approximate Google location
- Center-on-Drone behavior remains GPS → approximate location
- Localized location/weather/address-search UI

### Review items addressed
- Fixed the address-search Cancel path so Cancel does not move the map
- `showApiLocationBanner()` is always called with the required arguments
- Corrected wind-direction arrows
- Uses current 10.x behavior rather than the old `reloadActiveTab()` implementation
- Mission Control location handlers are namespaced and cleaned up to prevent duplicate API requests
- Pending Google location, weather, and address requests are aborted during tab cleanup
- Weather rendering tolerates missing optional fields
- `conditionsFetched` reflects the actual render result

### Validation
- 28/28 transpiler tests passed
- 62/62 Node tests passed
- JavaScript syntax checks passed
- Locale JSON validation passed
- Electron Forge packaging completed successfully
- Local Configurator smoke test completed successfully, including live Weather Conditions display
- Existing #2703 grid, keyboard, marker validation, and remaining-waypoint-capacity behavior preserved

### DIFF
--- a/js/configurator_main.js
+++ b/js/configurator_main.js
@@ -2,6 +2,7 @@ import '../src/css/styles.css'
 
 import $ from 'jquery';
 import 'jquery-ui-dist/jquery-ui';
+import jBox from 'jbox';
 import * as THREE from 'three'
 
 import GUI, { TABS } from './gui';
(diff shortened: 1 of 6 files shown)


GenCodeInc:
@sensei-hacker 
Thanks. I addressed the API-key normalization issue along with the Sonar findings.

For waypoint-only weather, the current behavior intentionally sources conditions from current GPS or approximate location, consistent with the original implementation. Using a mission waypoint as a weather source would expand the feature behavior, so I’ve left that unchanged for now.

I also agree that failed Weather API calls should not retry continuously. Adding retries would need a bounded cooldown/attempt policy to avoid repeated API calls, so I’ve left that unchanged rather than introducing an unbounded retry path.

MartinovEm:
To me it looks like a nice addition to Mission Control. One thought from a pilot's side, if you're open to it (and if sensei agrees, of course): current conditions are a good glance, but what really decides go/no-go is the next few hours - wind and gusts, ideally at the height we actually fly, plus rain and thunderstorms. Open-Meteo gives exactly that (hourly wind at 10/80/120/180 m, gusts, precipitation probability, thunderstorm codes, plain JSON, no API key) and would fit this panel nicely with an "hours ahead" and a height selector; Google's hourly forecast with gusts would work too, same key. Just an idea - I'll follow this with interest.

GenCodeInc:
@sensei-hacker We should be all set now

sensei-hacker:
> To me it looks like a nice addition to Mission Control. One thought from a pilot's side, if you're open to it (and if sensei agrees, of course): 

Sounds like a nice future addition to me.  Merging this one for now.

PR #2693, opened 2026-07-25
Browser based configurator

ramkrishna757575:
This pull request introduces a comprehensive browser-port dependency audit (`BROWSER_PORT_AUDIT.md`) and initiates the migration of the application to support browser environments. It documents the current Electron/Node dependencies, proposes browser-compatible replacements, and details the initial implementation of Web Serial support. Additionally, the main entry point in `index.html` is switched to a browser-specific bootstrap script.

**Browser-port audit and migration:**

* Added `BROWSER_PORT_AUDIT.md`, providing a detailed inventory of Electron/Node dependencies, browser-compatible alternatives, and migration status for serial I/O, filesystem, USB/DFU, dialogs, child processes, networking, and build/runtime boundaries. This document serves as a roadmap and checklist for the browser migration effort.
* Describes the new Web Serial transport (`js/connection/connectionWebSerial.js`), its integration, and the platform abstraction layer that enables the app to run in browsers with minimal code changes.
* Outlines the safe boundary for implementation, ensuring the MSP protocol layer and its consumers remain unchanged during the migration.
* Documents the current status and next steps, including remaining work for TCP/UDP, SITL, DFU, and file persistence.

**Build/runtime entry point update:**

* Changed the main script in `index.html` from `js/configurator_main.js` to `js/browser-entry.js`, ensuring the browser build uses the new platform abstraction and initialization logic.

PR #2448, opened 2025-12-01
Configurator as Web APP (PWA)

Scavanger:
*** Work in Progess ***

Preview, ~~only partially~~ working now: https://inav-configurator.pages.dev/
Now fully functional (only Bugfixing now) inkl. Demo mode and SITL 

Working:
- In principle, runs on the web (HTML/CSS/assets/...)
- Flashing via DFU
- MSP connection via WebSerial
- Some tabs are working

To do:
~~- Add more bridge code to replace elctron api calls (Dialgos, load/save file, ...)~~
~~- Remove update~~
- More from https://github.com/sensei-hacker/inav-configurator/blob/7d1b2bceca1cae9830fdccecddde4640d2e8498a/WEB_MIGRATION.md 

~~-Not~~- working, :
SITL, inkl Simulator imterface. But needs a local Websocket-to-Socket Proxy, because Applications in the browser are only allowed to comunicate via Websockets. 

Things to consider:
- The CORS policy is causing problems with the firmware flasher. I have now solved this by using a proxy on a Cloudflare worker, which allows the app to download the assets.
Any other suggestions?
- Where should all this be hosted later? Just on GitHub pages? => Deploy pipeline?
Github pages will not work, because SITL webassembly needs a special HTTP-Header, wich is impossible to set here. Cloudflare pages as alternative?

PR #2729, opened 2026-08-31
Browser/PWA configurator: Web Serial + WebUSB DFU + in-browser WASM SITL + installable PWA

sensei-hacker:
## Summary

Assembles a "best of" version of the browser-based (non-Electron) configurator from two overlapping efforts:

- **PR #2693** — Web Serial transport as a clean transport-boundary swap (`connectionSerialTransport.js`), used as the architectural base. The MSP layer and every existing caller are untouched.
- **PR #2722** (successor of #2448) — WebUSB DFU flashing, in-browser WASM SITL, and PWA installability, ported into #2693's architecture rather than taken wholesale, since #2693's own storage/file-dialog abstractions (already polyfilling `electronAPI` behind a single boundary in `js/browser/platform.js`) turned out to be the better pattern of the two.

End result: the same codebase now builds either as the existing Electron desktop app (`yarn make`) or as an installable web app (`yarn web:build`) via a new `vite.web.config.js`

## Changes

- **Web Serial transport** (`js/connection/connectionWebSerial.js`, `connectionSerialTransport.js`) — swaps the serial transport at the connection boundary only.
- **Browser platform shim** (`js/browser/platform.js`) — polyfills `electronAPI` (storage via `localStorage`, file open/save via `<input type=file>` and blob downloads) for everything outside Electron.
- **WebUSB DFU flashing** (`js/web/dfu.js`) — permission-request flow so the browser can reach devices in DFU mode; `js/protocols/stm32usbdfu.js` needed no changes.
- **In-browser WASM SITL** (`js/web/SITL-Webassembly.js`, `js/connection/connectionExt.js`) — runs the INAV SITL binary compiled to WebAssembly directly in the browser tab; capped to 3 consecutive auto-restarts after a crash instead of retrying forever.
- **PWA installability** (`vite.web.config.js` + `vite-plugin-pwa`) — web app manifest, service worker, offline precaching. Large binary assets (`.wasm`, `.gltf`, `.glb`) are emitted as real files rather than inlined as base64, to avoid blowing the service worker's precache size limit.

## Testing

- `yarn test` — 75/77 pass; the 2 failures (`tests/cli-tab-msp-polling.test.mjs`) are pre-existing on the base branch, confirmed unrelated via `git stash`.
- `yarn web:build` — clean build; verified `dist-web/manifest.webmanifest`, `dist-web/sw.js` + Workbox runtime, and the WASM binary + large `.gltf` models are emitted as real separate files (not inlined).
- `yarn package` (Electron) — builds and packages successfully; confirms the shared code touched by this PR (`connectionExt.js`, `SITL-Webassembly.js`) doesn't break the desktop bundle.
- Live-tested the built web app against real flight-controller hardware: CLI tab works over Web Serial, and WebUSB DFU firmware flashing completed successfully on a real FC.
- One unrelated pre-existing console error observed during live testing (a CORS-blocked `HEAD` request to a GitHub docs URL in `js/globalUpdates.js`, used only to pick a docs-link fallback) — confirmed via `git diff` that this file is untouched by this PR; out of scope here.

## Code Review

Reviewed with the `inav-code-review` agent. One release-blocking issue (unconditional firmware-download URL change breaking desktop flashing) and several correctness/robustness issues (double-invoked connect callback, unbounded crash-restart loop, log data loss on tab close, dead migration code) were found and fixed before this PR was opened.

## Known follow-ups (not blocking this PR)

- `inavflight.github.io/firmware/{version}/{file}` needs to actually be populated by the release process for browser-build firmware flashing to work end-to-end; that's separate release-pipeline work.
- The committed WASM SITL binary is pinned to firmware 9.0.0; there's no build step here that regenerates it from current firmware source, so it will drift from `package.json`'s reported version over time.


sensei-hacker:
Follow-up commit: selecting "choose port" now triggers connect immediately (deferred one tick so the picker isn't opened while the dropdown's own native popup is still closing).

Also confirmed: Firefox's native "Allow" stays disabled here even with a real, correctly-permissioned device connected — same code/device works in Chromium. That's rendered by the browser itself, not something a page can control; matches this PR's existing "use a Chromium-based browser" message.

PR #2730, opened 2026-09-02
fix(connection): guard against a legitimately-zero connectionId

sensei-hacker:
## Summary

Follow-up to #2716. `Connection.disconnect()` and every transport's `disconnectImplementation()`/`sendImplementation()` gate on `if (this._connectionId)`, which silently no-ops (or, after #2716's refactor, feeds `null` into `completeClose()`/`completeSend()`) whenever the id is falsy - not just when there's genuinely no connection, but also for a legitimate id of `0`.

Each transport's id counter here starts at 1, so this can't actually fire on this branch today. But #2716 just duplicated the same truthiness check into three more `completeSend()`/`completeClose()` call sites, and other configurator branches already have a transport where `0` is a real, live id (a WASM-hosted SITL port) - so the pattern is a latent bug waiting for the next transport that doesn't start counting at 1.

- Adds a shared `hasConnectionId()` helper on the base `Connection` class and uses it everywhere instead of testing `_connectionId` for truthiness.
- Changes the "never connected" sentinel from `0` to `null`, so a real id of `0` stays distinguishable from that initial state.

## Test plan

- [ ] `node -c` syntax check on the four touched files (done locally)
- [ ] Connect/disconnect over serial and confirm no regression
- [ ] Connect/disconnect over TCP (SITL) and confirm no regression

https://claude.ai/code/session_01231ETgELxNn2b4Ue4DhtUt

### DIFF
--- a/js/connection/connection.js
+++ b/js/connection/connection.js
@@ -11,8 +11,8 @@ const ConnectionType = {
 
 class Connection {
 
-    constructor() {       
-        this._connectionId   = 0;
+    constructor() {
+        this._connectionId   = null;
         this._openRequested  = false;
         this._openCanceled   = false;
         this._bitrate        = 0;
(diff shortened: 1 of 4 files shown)


sensei-hacker:
Closing - the motivating case (a transport where 0 is a legitimate connectionId) doesn't exist on this branch. All four transports here (serial/tcp/udp id counters starting at 1, BLE's fixed 0xff) make 0 structurally unreachable as a real id; it's used here specifically as this PR's own 'port is gone' sentinel (see tests/connection-send-queue.test.mjs), which my change broke. This fix is only relevant on branches carrying a transport that legitimately uses 0 (e.g. a WASM-hosted SITL port), not here.

PR #2736, opened 2026-09-04
MSP over MAVLink Tunnel support

xznhj8129:
Talk MSP to a flight controller over a MAVLink link, by wrapping MSP frames in MAVLink `TUNNEL` messages. This is the Configurator half of the firmware support merged in iNavFlight/inav#11718.

Without this, a vehicle on a MAVLink radio link has to be unplugged from its telemetry setup and connected over USB to change anything. With it, the Configurator connects through the same link the GCS uses. 

Note that it may be very slow due to the very nature of Mavlink and tunneling.

## Using it

A **MAVLink Tunnel** toggle appears next to Wireless mode, with an optional **SysID** to pin one vehicle when several share the link (0 = first heartbeat seen).

The Configurator waits for a `HEARTBEAT` before starting the MSP handshake, so it knows which system to address, then chunks outgoing MSP frames to the 128-byte `TUNNEL` payload and reassembles the replies addressed back to it.

## Requirements and limitations

- Needs firmware with `USE_MAVLINK_MSP_TUNNEL` (`maintenance-10.x` and later; enabled for any target with MAVLink telemetry that does not opt out via `DISABLE_MAVLINK_MSP_TUNNEL`).
- MAVLink 2 only — `TUNNEL` is message 385.
- The **CLI tab is unavailable** while tunnelling, and LTM decoding is disabled on the port. Neither survives the framing, so both are switched off rather than left to fail confusingly.
- Large configuration pages are noticeably slower over a radio link than over USB. The link's telemetry bandwidth, not the tunnel framing, is the limit.

## Three commits

**`Fix UDP connection lifecycle and udpClose IPC handler`** — two pre-existing bugs, unrelated to MAVLink, found while testing the tunnel over UDP:

- The dgram socket was created once at module import and `close()` called `socket.disconnect()`, which throws on a socket that was only bound, so a UDP connection could not be reopened after being closed. The socket is now created per connection and closed properly, and sends fail cleanly once it is gone instead of resolving nothing.
- `js/main/preload.js` already calls `udpClose()` through `ipcRenderer.invoke()`, but the main process registered it with `ipcMain.on()`, so the promise in `connectionUdp.js` never settled. Registered with `ipcMain.handle()` now.

Separated so they can be reviewed — or reverted — on their own.

**`Add MSP-over-MAVLink tunnel transport`** — the tunnel itself: `js/mavlink/mspTunnel.js` plus the connection plumbing, a transport hook in `js/msp.js`, and the UI.

**`Offer baud rates above 115200 when connecting`** — the connection baud dropdown stopped at 115200, so an ELRS MAVLink link at 460800 could not be opened at all, making the tunnel unreachable over the most common radio it would be used with. Adds 230400, 460800 and 921600.

## Testing

**On hardware:** an H743 over an ELRS MAVLink link at 460800 and MLRS link at 115200. The tunnel connects and the vehicle is configurable through it.

Against SITL built from `maintenance-10.x` (`8d0e6825`), with UART2 as a MAVLink telemetry port and no MSP on it:

| | Result |
|---|---|
| Plain MSP on the MSP port | Connects, handshake completes |
| **Control: MAVLink port, tunnel off** | `No configuration received within 10 seconds` — correctly fails |
| MAVLink port, tunnel on | Heartbeat detected on SYSID 1, full handshake, FC identified |
| Modes and Configuration tabs over the tunnel | Both load, 63 populated inputs — multi-chunk frames past the 128-byte payload work |
| CLI tab while tunnelling | Refused with an explanatory message |
| Disconnect, then reconnect without the tunnel | Clean; tunnel state resets |

The control row isolates the tunnel as the cause: same port, same firmware, tunnel off is dead and tunnel on is a working MSP session. Status bar read `MSP version: 2`, `MSP round trip: 44 ms`, **`Packet error: 0`** — no framing corruption.

Protocol details were cross-checked against `src/main/fc/fc_mavlink.c`: payload type `0x8001`, 128-byte chunking matching `MAVLINK_MSG_TUNNEL_FIELD_PAYLOAD_LEN`, the FC accepting `target_component == 0`, and the MAVLink 2 trailing-zero truncation of the FC's zero-padded reply.

### DIFF
--- a/index.html
+++ b/index.html
@@ -44,6 +44,10 @@
                 </div>
                 <div id="portsinput">
                     <div class="portsinput__row">
+                        <div id="mavlink-sysid-option" class="portsinput__top-element portsinput__top-element--port-override" style="display: none;">
+                            <label for="wireless-mavlink-sysid" i18n="wirelessMavlinkSysidLabel"></label>
+                            <input id="wireless-mavlink-sysid" type="number" min="0" max="255" value="0" />
+                        </div>
                         <div id="port-override-option" class="portsinput__top-element portsinput__top-element--port-override">
                             <label id="port-override-label" for="port-override" i18n="mainPortOverrideLabel"></label>
                             <input id="port-override" type="text" value="/dev/rfcomm0" />
@@ -77,6 +81,12 @@
                                 <span class="" data-i18n="wirelessModeSwitch"></span>
                             </label>
                             <input id="wireless-mode" class="togglesmall" type="checkbox" />
+                        </div>
+                        <div id="mavlink-tunnel-option" class="portsinput__top-element portsinput__top-element--inline" style="display: none;">
+                            <label for="wireless-mavlink-tunnel">
+                                <span data-i18n="wirelessMavlinkTunnelSwitch"></span>
+                            </label>
+                            <input id="wireless-mavlink-tunnel" class="togglesmall" type="checkbox" />
                         </div>
                     </div>
                 </div>
--- a/index.html
+++ b/index.html
@@ -62,6 +62,9 @@
                         <div class="dropdown dropdown-dark portsinput__top-element">
                             <!--suppress HtmlFormInputWithoutLabel -->
                             <select class="dropdown-select" id="baud" title="Baud Rate">
+                                <option value="921600">921600</option>
+                                <option value="460800">460800</option>
+                                <option value="230400">230400</option>
                                 <option value="115200" selected="selected">115200</option>
                                 <option value="57600">57600</option>
                                 <option value="38400">38400</option>
--- a/js/configurator_main.js
+++ b/js/configurator_main.js
@@ -150,6 +150,10 @@ $(function() {
                 }
 
                 if (GUI.allowedTabs.indexOf(tab) < 0) {
+                    if (CONFIGURATOR.mavlinkTunnelActive && tab === 'cli') {
+                        GUI.log(i18n.getMessage('tabSwitchMavlinkTunnelUnavailable'));
+                        return;
+                    }
                     GUI.log(i18n.getMessage('tabSwitchUpgradeRequired', [tabName]));
                     return;
                 }
--- a/js/connection/connection.js
+++ b/js/connection/connection.js
@@ -24,6 +24,7 @@ class Connection {
         this._onReceiveListeners      = [];
         this._onReceiveErrorListeners = [];
         this._type = null;
+        this._timeoutOverride = null;
         
         if (this.constructor === Connection) {
             throw new TypeError("Abstract class, cannot be instanced.");
@@ -260,11 +261,19 @@ class Connection {
         this._transmitting = false;
     }
 
+    setTimeoutOverride(timeoutMs) {
+        this._timeoutOverride = timeoutMs;
+    }
+
     /**
      * Default timeout values
      * @returns {number} [ms]
      */
     getTimeout() {
+        if (this._timeoutOverride !== null) {
+            return this._timeoutOverride;
+        }
+
         if (this._bitrate >= 57600) {
             return 3000;
         } if (this._bitrate >= 19200) {
--- a/js/data_storage.js
+++ b/js/data_storage.js
@@ -8,6 +8,7 @@ var CONFIGURATOR = {
     'connectionValidCliOnly': false,
     'cliActive': false,
     'cliValid': false,
+    'mavlinkTunnelActive': false,
     'connection': false
 };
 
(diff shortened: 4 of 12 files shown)


sensei-hacker:
What do you think about those Qodo comments?
Also I see there is a merge conflict.

xznhj8129:
> What do you think about those Qodo comments? Also I see there is a merge conflict.

The first one might be the cause of some dropouts on loading lots of settings, i''ll check it
The second i honestly don't see how it's possible for a user to do

sensei-hacker:
I took a look at this with an AI-based tool I have, following up on the Qodo review above. It may well be wrong on any of the points below — please treat this as a set of questions to check rather than a verdict, and correct me if I've misread something.

## Frame parser (`mspTunnel.js`) — swallowing valid frames on a bad candidate?

Qodo's first finding looked right to me when I traced through `ingest()`: once enough bytes arrive to match a candidate's *declared* length, the whole candidate gets sliced off the buffer before its CRC is checked. If the candidate turns out to be bogus (corrupted header, or a stray `0xFE`/`0xFD` byte that happens to land where real frame data is), the CRC check fails afterward but by then the candidate's bytes — including any real HEARTBEAT/TUNNEL frame they overlapped — are already gone, since `_findFrameStart()` only resumes from whatever's left in the buffer. On the lossy radio links this feature targets, that seemed like it could delay heartbeat discovery or cause MSP timeouts more than necessary.

I put together a small patch that only removes the CRC-checked candidate from the buffer *after* it passes, and on failure drops just the one magic byte instead of the whole candidate, so a real frame starting partway through isn't discarded with it:

```diff
-            const frame = this._buffer.slice(0, frameLength);
-            this._buffer = this._buffer.slice(frameLength);
-
-            if (crcExtra === null) {
-                continue;
-            }
-
-            if (!this._crcMatches(frame, magic, payloadLength, checksumOffset, crcExtra)) {
-                continue;
-            }
+            const frame = this._buffer.slice(0, frameLength);
+
+            if (crcExtra !== null && !this._crcMatches(frame, magic, payloadLength, checksumOffset, crcExtra)) {
+                // Not really a HEARTBEAT/TUNNEL frame - the magic byte may just
+                // coincide with unrelated data. Drop only that byte and rescan,
+                // instead of the whole candidate, so a real frame starting
+                // inside it isn't discarded too.
+                this._buffer = this._buffer.slice(1);
+                continue;
+            }
+
+            this._buffer = this._buffer.slice(frameLength);
+
+            if (crcExtra === null) {
+                continue;
+            }
```

I ran this against a hand-built stream (a bogus/corrupt candidate immediately followed by a real heartbeat) and it now correctly recovers the heartbeat instead of losing it. Does this match your understanding of the code, and would it be useful as-is, or is there context I'm missing about why the original ordering was chosen? Happy to push this to your branch directly if that's easier than a diff in a comment — just let me know.

## A few other things I noticed while in there (unverified beyond reading, so please double check)

- **CRC only covers HEARTBEAT/TUNNEL** — every other message ID on the link (and this is explicitly meant to share a port that also carries normal telemetry) skips CRC validation entirely and just trusts the declared length byte to know how far to advance. Was that a deliberate scope decision (only HEARTBEAT/TUNNEL matter, so why validate anything else), or worth extending later?
- **`onValidFirmware()` vs. `isMavlinkTunnelEnabled()`** (`serial_backend.js`) — this reads the live checkbox state rather than `CONFIGURATOR.mavlinkTunnelActive`, which looked like it was snapshotted once at connect time. Could the two disagree if someone toggles the tunnel checkbox while the handshake is still in flight?
- **`onInvalidFirmwareVariant()`/`onInvalidFirmwareVersion()`** still set `GUI.allowedTabs = ['cli']` unconditionally for recovery — since CLI can't function over the tunnel, could a tunnel-mode connection end up stuck on a CLI tab that doesn't actually work?
- **`wrapMspFrame()`** throws synchronously if called before a target is discovered, and `msp.js`'s `send_message()` doesn't appear to catch that. It looks like today's flow always gates on `targetDiscovered` first, so this may never trigger — just flagging it in case a future caller doesn't preserve that ordering.
- **`udp.js`'s `connect()`** calls `this._socket.close()` without awaiting it before creating the replacement socket, while the standalone `close()` method does await it. Could this race if a reconnect happens quickly?

Really appreciate the work here — the MAVLink framing and CRC implementation looked solid where I could check it against the spec, and the UDP socket-lifecycle fix looks like a genuine improvement. These are just things I'd want to understand before feeling confident about the parser's robustness on a lossy link, which seems to be exactly the scenario this is built for.

PR #2737, opened 2026-09-05
fix(cli): stop syncing raw FC output into command box during autocomplete build

kolabuzlu:
### What
Entering the CLI tab with autocomplete enabled showed random/garbled characters in the command entry box for ~15-20 seconds after connecting, which then settled down once the tab became usable (v9.1.1).

### Why
`cliTab.read()` had a leftover unconditional `setPrompt(this.cliBuffer)` call that mirrors the FC's incoming line buffer into the command textarea on every serial read. When CliAutoComplete was added (21a41698), a second, gated `setPrompt()` call was introduced above it to only run this "native fallback" echo when autocomplete is *not* enabled - but the original unconditional call was left in place below it, so it kept firing regardless. That meant every chunk of the autocomplete builder's hidden `help`/`dump`/`get` traffic (used to enumerate every FC setting on CLI entry) was written straight into the user's command box for as long as the cache build took.

### Fix
Remove the redundant unconditional call; keep only the one already gated by `!CliAutoComplete.isEnabled()`, matching how `writeLineToOutput()` already suppresses builder traffic.

### Testing
Reproduced by opening the CLI tab against an FC with autocomplete enabled; confirmed the command box previously filled with garbage during "Building AutoComplete Cache..." and stays clean with this fix

### DIFF
--- a/tabs/cli.js
+++ b/tabs/cli.js
@@ -518,12 +518,15 @@ cliTab.read = function (readInfo) {
         }
     }
 
-    // fallback to native autocomplete
+    // Mirror the FC's line buffer into the command textarea. This is only correct as a
+    // "fallback to native autocomplete" when nothing else is driving the textarea: while the
+    // autocomplete cache builder is running (silently sending `help`/`dump`/`get` right after
+    // CLI entry, see CliAutoComplete.builderStart), this would splash fragments of that hidden
+    // output into the command box instead of the user's own input, so it must stay gated behind
+    // the same isEnabled() check used above rather than always running unconditionally.
     if (!CliAutoComplete.isEnabled()) {
         setPrompt(removePromptHash(this.cliBuffer));
     }
-
-    setPrompt(removePromptHash(this.cliBuffer));
 
     if (cliTab.promptCallback && this.cliBuffer.endsWith('# ')) {
         const cb = cliTab.promptCallback;


sensei-hacker:
Thanks for that.  Can you check against the *current* code, please?  There have been at least three and probably four relevant changes since last year.

The proper branch is maintenance-10.x

kolabuzlu:
> Thanks for that. Can you check against the _current_ code, please? There have been at least three and probably four relevant changes since last year.
> 
> The proper branch is maintenance-10.x

Checked against maintenance-10.x — this is already fixed there by e484b1c6 (error414, PR #2681) and cd9f11a1 (Ray Morris), both from July. They handle it more thoroughly than my patch (also cover MSP status polls leaking into the CLI stream, which I hadn't caught). Since there's nothing left to fix on that branch, I'll close this PR unless you'd rather see those two commits backported to master/9.1.x, since that's the release line the bug was actually reported against.

sensei-hacker:
Awesome, thanks. Six months ago I would have merged it against maintenance-9.x

And I probably should have made the earlier fixes there. 9.1.2 will be only be required but fixes that are also in 10.x, so I probably won't merge this since it would create a conflict with merging up to 10.x.

kolabuzlu:
> Awesome, thanks. Six months ago I would have merged it against maintenance-9.x
> 
> And I probably should have made the earlier fixes there. 9.1.2 will be only be required but fixes that are also in 10.x, so I probably won't merge this since it would create a conflict with merging up to 10.x.

Makes sense, thanks for explaining the 9.1.2 policy. Since you mentioned you probably should have backported the earlier fixes to maintenance-9.x — I'm happy to do that: I tested cherry-picking e484b1c6 and cd9f11a1 onto maintenance-9.x and it applies cleanly (one trivial conflict in a test file, easy to resolve), so it wouldn't create the merge conflict you're worried about later. Want me to open that as a separate PR? If so I'll close this one.

sensei-hacker:
> I'm happy to do that: I tested cherry-picking https://github.com/iNavFlight/inav-configurator/commit/e484b1c62b4e7b43be934c031dad618a3c08feb3 and https://github.com/iNavFlight/inav-configurator/commit/cd9f11a12d084c82a93fd40024696d06abf78403 onto maintenance-9.x and it applies cleanly (one trivial conflict in a test file, easy to resolve), so it wouldn't create the merge conflict you're worried about later. Want me to open that as a separate PR? If so I'll close this one.

That would be awesome, thanks. Milestone will be 9.1.2

PR #2738, opened 2026-09-05
Add Serbian language

flightmode-academy:
Adds Serbian (Latin) translation of iNav Configurator.

All 2,263 strings translated, plus language_sr. Placeholders ($1, $2)
and HTML tags checked against the English source - no mismatches.
Both JSON files validate.

Three changes:
- new locale/sr/messages.json
- 'sr' added to availableLanguages in js/localization.js
- language_sr added to locale/en/messages.json

Terminology follows the Serbian translations of Betaflight Configurator
and Gyroflow, which I also did:

crowdin.com/project/betaflight-configurator
(3,209 strings, Latin and Cyrillic, both at 100%, accepted into the
official project)

github.com/gyroflow/gyroflow/blob/master/resources/translations/sr_CS.ts
(merged into their repository)

Serbian isn't currently among the languages iNav supports. It covers
around 20 million people across Serbia, Bosnia, Montenegro and Croatia,
and there is no FPV training or documentation in the language apart
from what I write. I run a free academy at flightmode.website and I'll
keep this translation current across releases.

Happy to adjust anything - file layout, language code, or terminology.


flightmode-academy:
Targeting master on purpose — this adds a new language rather than
changing existing behaviour, so it belongs in the next release rather
than in a maintenance branch. Happy to retarget if you prefer it
elsewhere.

For context: 2,264 strings, complete, no untranslated entries. Serbian
is also merged into Betaflight Configurator (PR #5470) and into Gyroflow,
so the terminology matches across all three programs — a pilot who learns
a term in one expects the same word in the others.

sensei-hacker:
> so it belongs in the next release 

Yes, it does belong in a release. I'm moving it to maintenance-10.x so it can be in a 10.x release.

We have a policy that a PR should mark which major version it uses as base / intended destination.

As mentioned in the automatic warning, 10.x is the next major release and "If master is the correct target for this change **because it should never be included in any release** ..."

It looks like there are some minor merge conflicts with 10.

sensei-hacker:
Since I don't know Serbian myself, I had my AI-assisted tool spot-check 50 randomly sampled strings from `locale/sr/messages.json` against the English source, focused on proper nouns/initialisms (GPS, MSP, OSD, CLI, board names, etc. — it found zero errors there, all correctly preserved) and general translation quality.

It may well be wrong on some of these. I don't expect the LLM to actually be reliable, but these kinds of things might be worth checking, just in case the AI happens to be right on some of them. Please treat them as questions to check rather than a verdict, and correct me if I've misread something. Overall it came back fairly positive: 44/50 sampled strings looked accurate and natural, and the "Pitoova cev" adaptation for pitot tube looks like the right standard aviation term, not an error.

A few smaller things it flagged as possibly worth a look:

1. **`w_z_baro_p`** — EN "Vertical position barometer weight" → SR "Težina barometer-a za vertikalnu poziciju". Is "barometer-a" meant to be "barometra" (genitive of "barometar")? Looks like it might be splicing the English spelling with a Serbian case ending.

2. **`osdElement_GLIDESLOPE`** — EN "Glideslope" → SR "Putanja planiranja". Since "planiranje" most commonly reads as "planning" in standard Serbian, could this read ambiguously to most users? Wondering if something like "Glisada" or "Ugao spuštanja" might be clearer, if that fits your intent — you'd know best.

3. **`osdElement_RX_POWER_DOWNLINK`** ("Rx Power Downlink" → "Rx snaga povratne veze") vs. **`osdElement_LQ_DOWNLINK`** ("RX Downlink Quality %" → "Kvalitet RX silazne veze %") — these render "Downlink" two different ways ("povratne veze" vs. "silazne veze") in the same file. Intentional distinction, or worth aligning?

4. **`portsFunction_RX_SERIAL`** — EN "Serial RX" is left as "Serial RX" (untranslated), while the neighboring `portsFunction_GIMBAL`/`portsFunction_HEADTRACKER` translate "Serial" as "Serijski". Was RX_SERIAL meant to follow the same pattern?

5. **`landMinAltVspdHelp`** — this one I'm less sure about: three phrases (`Final approach altitude`, `Slow down altitude`, `Initial landing speed`) appear in English inside `<strong>` tags mid-sentence. I couldn't find standalone keys with those exact strings elsewhere in the file, so I'm not sure if these are meant to reference other translated UI labels or if they're literal strings generated outside the i18n system (e.g. from settings.yaml) — in which case leaving them in English would be correct as-is. Do you know which it is?

That's from a sample of 50 strings. It may reveal some patterns throughout the file?

Separately, and not something to act on now: the sr file has 225 keys not yet present that exist in the en source (out of the sample's scope, just flagging for awareness on translation completeness going forward).

Thanks for the translation — really appreciate you bringing Serbian coverage to the project!

flightmode-academy:
Thanks for the review — most of these are right.

**1. `w_z_baro_p`** — correct, that is an error. "barometer-a" mixes the English
spelling with a Serbian case ending. Changing it to `Težina barometra za
vertikalnu poziciju`.

**2. `osdElement_GLIDESLOPE`** — needs changing, but not to "Glisada" or "Ugao
spuštanja". The help string reads "Amount of horizontal distance traveled per
unit of lost altitude", so this is a ratio, not an approach path or an angle.
"Glisada" in Serbian is the ILS approach path.

The wrong word in mine is "putanja". "Planiranje" is correct for gliding and is
what I already use in `GLIDE_TIME` and `GLIDE_RANGE`. Changing it to `Odnos
planiranja`.

**3. `RX_POWER_DOWNLINK` vs `LQ_DOWNLINK`** — not intentional. The file already
has `LQ_UPLINK` as "uzlazne veze" and `LQ_DOWNLINK` as "silazne veze".
`RX_POWER_DOWNLINK` is the odd one out. Changing it to `Rx snaga silazne veze`.

**4. `portsFunction_RX_SERIAL`** — deliberate. "Serial RX" is the label in the
Ports tab of Betaflight Configurator, where the Serbian translation is already
merged and also leaves it untranslated. Translating it here would make the two
configurators disagree on the same screen.

`GIMBAL` and `HEADTRACKER` read as descriptions rather than labels, so "Serijski
gimbal" works there.

If you prefer consistency within this file over consistency across the two
tools, I will change it.

**5. `landMinAltVspdHelp`** — correct, those are not external strings. They are
labels of fields translated elsewhere in this same file: `fwLandApproachLength`
is already "Dužina završnog prilaza". I will translate them to match.

**6. The 225 keys** — against the English file I used the difference is one key
(`language_sr`). I moved this PR from `master` to `maintenance-10.x`, and my
snapshot is from early September. Which branch did you compare against? Either
way I will rebase on the target branch and fill in what is missing.

One more, not on your list: "final approach" is "završni prilaz" in
`fwLandApproachLength` but "finalni prilaz" in its help text and in
`fwLandFinalApproachPitch2throttle`. Aligning those too.

sensei-hacker:
Awesome, thanks! This will be targeting maintenance-10.x

"master" is a lagging copy of 9.x.y, so that when someone doesn't know which version they are targeting (or doesn't pay attention), it doesn't drag 500 INAV 10 commits into an INAV 9.x.y fix.

> If you prefer consistency within this file over consistency across the two
tools, I will change it.

I'll leave it up to as to what makes sense. Whichever is "best".
Consistency with Betaflight isn't a particular goal for me, but it might make future updates easier if you plan to maintain both from time to time. Easier to maintain one translation of a string than two, eh?

I figure if INAV was just like Betaflight, one of them would be redundant and doesn't need to exist. :)  Each has certain things they do better, so I prefer to do anything the *best* way, whatever that means.

flightmode-academy:
Thanks — that explains the branch better than the name does.

On `portsFunction_RX_SERIAL`, taking your "whichever is best" seriously rather
than reaching for consistency as the answer: I am keeping "Serial RX", and the
reason is the string itself rather than Betaflight.

"Serial RX" is the name of the firmware feature. It is what appears in the CLI,
in the docs, and in every video and forum post a beginner will find while
setting up a receiver. A Serbian pilot searching for help types "Serial RX", so
translating it breaks the link between what is on their screen and everything
written about it. `GIMBAL` and `HEADTRACKER` carry none of that weight — nobody
searches by those names — which is why translating those and not this one is a
real distinction rather than an oversight.

That it also matches the Serbian Betaflight translation is a convenience for me
later, not the reason. Though your point stands: I maintain both, so one string
is one place to fix instead of two.

A correction to my own point 5, which I got wrong. I checked the file before
changing anything: "Final approach altitude", "Slow down altitude" and "Initial
landing speed" do not exist as keys in `messages.json`. They appear only inside
other help strings, and the labels themselves come from the firmware — so they
render in English in the UI, and the help text matches what the user sees. Your
second reading was the right one, so I left them as they are.

Everything else is pushed: 253 strings for what came with maintenance-10.x
(DroneCAN, Map Generator, mission path simulation, Google location and weather,
thermal camera, ADS-B, grid survey), plus the four fixes from your list.

sensei-hacker:
Excellent!  I through another 100 random samples at an LLM since I don't know Serbian. It said that sample of 100 looks really good. :)  It pointed out this "inconsistency" I should ask you about:

failsafeOffDelayHelp: "…letelica ne disarm-uje" — half-English hybrid?
osdElement_ONTIME_FLYTIME_HELP: "dok nije armed" — "armed" left in English?
configurationAutoDisarmDelay: "razoružavanja" — fully translated?

Again, I have no idea, so I'm just asking for what you think.

flightmode-academy:
Good catch — that's a real one.

Serbian can either take the English term or translate it, and the file did both. I've settled it the way the Serbian Betaflight translation does, so the two configurators agree:

- **arm** → *armovati / armovanje / armovan* — the English term with Serbian endings
- **disarm** → *razoružati / razoružavanje / razoružan* — a native verb
- **ARM**, **PREARM** and the other mode names stay in English, as they appear in the Modes tab

So two of your three were wrong: `failsafeOffDelayHelp` and `osdElement_ONTIME_FLYTIME_HELP`. `configurationAutoDisarmDelay` was already right.

`osdElement_ONTIME_FLYTIME_HELP` had a second problem the sample didn't catch — it quoted "On Time" and "Fly Time", which appear nowhere in the UI. The element is listed as "Vreme rada / Vreme leta" and the OSD draws only a symbol and the time. The help quotes the Serbian names now.

And I need to correct myself. On the 18th I wrote that the landing labels — "Slow down altitude", "Initial landing speed", "Final approach altitude" — come from the firmware and render in English. That was wrong. Those fields are `data-i18n` labels from `messages.json`, and the phrases I quoted don't exist in the UI in any language; they only occur inside the English help text. They are translated as descriptions now, not as quoted labels.

Since that class of mistake clearly wasn't limited to three strings, I went through the whole file against the English source rather than patch them: **372 strings changed** — help quoting labels that don't match, leftover English, and terms that had two or three different renderings (home, gyro, propeller, file, backup, cruise).

I also caught the file up to today's `maintenance-10.x`: **83 strings added, 2 dropped**. That covers the alignment wizard, the waypoint panel, coordinated turns, the rainbow overlay and the SRXL2 ESC calibration — I didn't want "REMOVE THE PROPELLER" showing up in English.

Two notes for you, unrelated to Serbian:

- `missionSimulationDisclaimer` still describes `nav_fw_wp_turn_smoothing ON / ON-CUT`, which #2711 replaced with `nav_fw_wp_turn_mode`.
- `srxl2CalibrateRefusedNoPort` says "set the motor protocol to SRXL2", but that field is labelled "ESC protocol" — which is what the neighbouring `srxl2ProtocolNotSet` calls it.

`en/messages.json` is four added lines and `js/localization.js` is one; everything else is `locale/sr/messages.json`.

PR #2740, opened 2026-09-08
Fix INA226 settings flicker during configuration loading

FlyingRC-Official:
INA226 fields added in #2707 briefly appear on the Configuration tab even when neither meter uses INA226: `Settings.configureInputs()` calls `.show()` as each setting loads, overriding the initial inline `display: none`, before the final visibility update hides the fields again.

Use `hidden` for the INA226 rows with a narrowly scoped CSS rule that survives the loader's `.show()`. The existing metadata-based selection logic reveals the appropriate fields once settings have loaded. ADC calibration controls retain their original immediate display and visibility behavior.

Validation:
- Reviewed the cumulative diff: only configuration HTML, JS and CSS changed (12 additions, 5 deletions).
- Seven browser regression scenarios passed using the actual HTML/CSS, jQuery, settings loader and visibility functions with delayed simulated MSP responses: all four voltage/current combinations, missing INA settings, a read failure and tab re-entry. ADC controls remained visible from the first sample and INA226 controls did not appear prematurely. Selection changes and parameter saves also passed.
- macOS arm64 DMG and ZIP rebuilt successfully; `git diff --check` passed.
- The full test suite run before the final ADC display correction had 28 transpiler tests passing and 108 main tests passing, with the same three failures observed on the unmodified base (two CLI polling positive controls and MZTC write-code classification). The final correction was checked with the browser regression and packaging; the full suite was not rerun.

Browser regression uses simulated responses; physical flight-controller verification of the final build remains pending.


Jetrell:
Thanks for the fix @FlyingRC-Official
Tested and it works as expected.

Issue #2741, opened 2026-09-08
Mission Control 3D view: JUMP return legs are not drawn or terrain-checked


Raffi1202:
Follow-up from #2710 (3D terrain review in Mission Control).

**Current behaviour**
The 3D route is built from the waypoints in storage order (`getMission3DPoints()` / `getMission3DRouteSegments()` in `js/mission_3d.js`). Attached actions such as JUMP are dropped, so:

- the return leg from the waypoint carrying the JUMP back to its target waypoint is not drawn in 3D and not sampled for terrain collisions;
- repeated legs are not represented as such (they reuse edges that are already sampled, so no terrain information is lost there).

The 2D editor already draws the return leg as the magenta "Repeat x" line, so the information exists, just not in the 3D view.

**Expected behaviour**
Expand the mission in execution order like the existing traversal in `js/waypointCollection.js` (jump target and repeat count), so the JUMP return leg is drawn and terrain-checked in 3D. Infinite repeats (`p2 == -1`) should be handled by sampling each edge once. Multi-mission index handling (`multiMissionWPNum + p1`) needs to be respected the same way `repaintLine4Waypoints()` does it in 2D.

**Notes**
- Raised by the Qodo review on #2710, confirmed as a real gap.
- Unit tests live in `tests/mission-3d.test.mjs`.

PR #2742, opened 2026-09-08
Draw and terrain-check JUMP return legs in the 3D mission view

Raffi1202:
## Problem
Follow-up to #2710, closes #2741. In the Mission Control 3D view the route is built from the waypoints in storage order and attached actions are dropped, so the leg a JUMP adds from the waypoint it is attached to into its target waypoint is neither drawn nor sampled for terrain collisions. The 2D editor already shows that leg as the magenta "Repeat x" line.

## Cause
`js/mission_3d.js:38` on maintenance-10.x skips every attached waypoint, so a JUMP never reaches the point list, and `getMission3DRouteSegments()` (`js/mission_3d.js:88`) chains the remaining route points in storage order. `tabs/mission_control.js:1759` samples only those segments.

## Change
`getMission3DFlightLegs()` (new) walks the waypoints in execution order: a JUMP is taken P2 times, an infinite one once, a zero-repeat one never; P1 is resolved relative to its sub-mission start; jumps whose target is missing, itself or not a route waypoint, or that follow an RTH/LAND, are ignored; a step limit bounds malformed loops; each leg is recorded the first time it is flown. `getMission3DFlightSegments()` replaces `getMission3DRouteSegments()` and gives each jump leg its own segment carrying the repeat count. In `tabs/mission_control.js`, `sampleRouteTerrain()` samples those segments, `renderRouteTerrain()` draws clear jump legs in the 2D editor's `#e935d6` with a mid-leg "Repeat x N" / "Repeat x infinite" label while colliding runs stay red and raise the warning, and the terrain cache signature includes the legs. `getMission3DPoints()` is split into per-point helpers and points carry `waypointNumber`.

## Test
Unit tests in `tests/mission-3d.test.mjs` (27 tests, 13 of them for the flight-leg walk: backward, forward, infinite and zero-repeat jumps, sub-mission offsets, invalid targets, a jump after RTH, the step limit) pass with `node --test tests/mission-3d.test.mjs`; CI does not run the test suite. CI build green on all six platforms for cd6862c: https://github.com/iNavFlight/inav-configurator/actions/runs/34761993579. SonarCloud quality gate passed: https://sonarcloud.io/dashboard?id=iNavFlight_inav-configurator&pullRequest=2742. Not checked in the app with an attached screenshot or log; the Cesium rendering in `tabs/mission_control.js` is covered by the build only.

## Flash / RAM

## Docs
No documentation change needed: neither `docs/` nor `README.md` describes the 3D mission view, and the JUMP semantics the 3D route now follows (P1 target, P2 repeats, -1 infinite) are the ones the 2D editor already implements.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,7 @@ jobs:
               with:
                 name: pr-number
                 path: pr_number.txt
+                if-no-files-found: error
 
     build-linux:
         runs-on: ubuntu-24.04
@@ -69,16 +70,19 @@ jobs:
               with:
                 name: ${{ env.BUILD_NAME }}_DEB
                 path: ./out/make/deb/x64/*.deb
+                if-no-files-found: error
             - name: Upload Linux rpm
               uses: actions/upload-artifact@v4
               with:
                 name: ${{ env.BUILD_NAME }}_RPM
                 path: ./out/make/rpm/x64/*.rpm
+                if-no-files-found: error
             - name: Upload Linux zip
               uses: actions/upload-artifact@v4
               with:
                 name: ${{ env.BUILD_NAME }}_ZIP
                 path: ./out/make/zip/linux/x64/*.zip
+                if-no-files-found: error
 
     build-linux-aarch64:
         runs-on: ubuntu-24.04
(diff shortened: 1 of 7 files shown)


Raffi1202:
Both Qodo points were right, and cfcaac9 addresses them at the root: the route is now built in execution order instead of storage order plus a shortcut.

1. **Forward JUMP** – `getMission3DFlightLegs()` walks the mission the way the firmware flies it, so the waypoints a forward jump skips get no legs and no terrain check. They still show as markers, which makes the skip visible in 3D.
2. **Zero repeats** – a JUMP with P2 = 0 is never taken, so it adds no leg. The 2D editor keeps its "Repeat x0" line as a hint of the setting; the 3D view only shows what is flown.

Finite repeats are taken as often as configured, an infinite one once, and every leg is recorded the first time it is flown, so repeats add no duplicate samples. Verified in the dev app with a backward x2 jump, the same jump at x0, and a forward jump that skips a waypoint. 31 focused tests pass.

Raffi1202:
@sensei-hacker this is the follow-up for #2741 we talked about on #2710. CI is green on all platforms and SonarCloud is clean on the current head (eb9f911).

One design choice I would like your view on: the 3D route now follows execution order, so waypoints a forward JUMP skips keep their markers but get no route legs, and a JUMP with 0 repeats adds no leg at all. The 2D editor still draws all storage-order lines plus its "Repeat x" line, so the two views deliberately differ there. Does that match how you would expect the 3D view to behave, or would you rather keep it visually identical to 2D?

Issue #2723, opened 2026-08-27
If manual mode is enabled as an available mode in the mode selector, but not currently selected by the switches, it still prevents the acro mode indicator from being shown


Patronics:
I've identified this issue to be caused by manual mode being continuously on when the configurator is active, as controlled by the logic here:
https://github.com/iNavFlight/inav-configurator/blob/515307c27c61cf193aa6c9f32966c3f858eb3c36/tabs/auxiliary.js#L403-L409

This leads to misleading display of the status for the acro mode, seeming to indicate that acro mode is unable to be activated even when no other modes are active.

See the following screenshots for an example of the behavior of the acro mode when there's a range defined for manual mode, as opposed to when manual mode has no assigned range.

<img width="1591" height="1052" alt="Image" src="https://github.com/user-attachments/assets/bcf69319-ac45-4cda-b57b-f9d72a26e6d7" />

<img width="1587" height="1052" alt="Image" src="https://github.com/user-attachments/assets/cf54906f-8fb7-40a5-ad01-5e8abff90b1f" />

sensei-hacker:
The only way I can reproduce this is to flip the aircraft upside down so that the runtime calibration never finishes. In which case it is in fact in manual mode -- Acro doesn't engage until the runtime calibration completes, normally within a few seconds of powering the board.
As long as my board is sitting relatively flat, Acro works fine for me?

Patronics:
Hmm, I'll need to do a bit more testing then. Either way this is misleading behavior as in that case it shouldn't falsely indicate that Acro mode is active unless there's a configured (yet currently inactive) option for manual mode.

PR #2751, opened 2026-09-10
Base the ACRO indicator on the selected modes

Raffi1202:
## Problem

A range assigned to MANUAL makes the Modes tab show ACRO as unavailable even when the switch is nowhere near that range, so the indicator suggests ACRO cannot be activated while no other mode is active. Without a range on MANUAL the indicator is correct. Fixes #2723.

## Cause

`update_ui()` tests `FC.isModeBitSet()` first and clears `acroEnabled` for every mode in `acroFail` - `tabs/auxiliary.js:416-421` on maintenance-10.x. The channel-range loop sits in the `else` branch (lines 425-438), so it never runs for a mode the flight controller reports as active, and the FC reports MANUAL as active whenever the Configurator is connected.

## Change

The range loop now runs before the mode-bit check, so `inRange` is known in both branches. A mode in the new `acroFailOnlyWhenSelected` set - currently only MANUAL - clears `acroEnabled` only when its range is selected; the other `acroFail` entries keep their behaviour. Ranges on a channel outside the receiver's aux channels, or with a marker position that does not parse as a number, are skipped instead of counting as selected. The per-mode row classes are untouched and the two mode lists are now `Set`s.

## Test

Not run in the app for this revision; the reproduction, with screenshots, is the reporter's in #2723. Cause verified by reading `tabs/auxiliary.js:416-438`, and `node --check tabs/auxiliary.js` passes on the head commit. CI is green on `d833634`: [six platform builds](https://github.com/iNavFlight/inav-configurator/actions/runs/34615547870) and [SonarCloud](https://sonarcloud.io/dashboard?id=iNavFlight_inav-configurator&pullRequest=2751), 0 new issues. The one Qodo correctness finding, ranges on an unavailable channel counting as selected, is what `d833634` fixes. No automated test covers `update_ui()`.

## Flash / RAM

## Docs

None. No Markdown file on maintenance-10.x mentions the Modes tab or the ACRO indicator.

### DIFF
--- a/tabs/auxiliary.js
+++ b/tabs/auxiliary.js
@@ -395,10 +395,10 @@ auxiliaryTab.initialize = function (callback) {
         function update_ui() {
             let hasUsedMode = false;
             let acroEnabled = true;
-            let acroFail = ["ANGLE", "HORIZON", "MANUAL", "ANGLE HOLD", "NAV RTH", "NAV POSHOLD", "NAV CRUISE", "NAV COURSE HOLD", "NAV WP", "GCS NAV"];
+            let acroFail = new Set(["ANGLE", "HORIZON", "MANUAL", "ANGLE HOLD", "NAV RTH", "NAV POSHOLD", "NAV CRUISE", "NAV COURSE HOLD", "NAV WP", "GCS NAV"]);
             // The flight controller reports these modes as active for as long as the Configurator
             // is connected, so only the selected channel range can make them block ACRO.
-            let acroFailOnlyWhenSelected = ["MANUAL"];
+            let acroFailOnlyWhenSelected = new Set(["MANUAL"]);
 
             var auxChannelCount = FC.RC.active_channels - 4;
 
@@ -436,14 +436,14 @@ auxiliaryTab.initialize = function (callback) {
                     // The flight controller can activate the mode
                     $('.mode .name').eq(modeElement.data('index')).data('modeElement').addClass('on').removeClass('inRange').removeClass('off');
 
-                    if (jQuery.inArray(modeElement.data('modeName'), acroFail) !== -1 &&
-                        (inRange || jQuery.inArray(modeElement.data('modeName'), acroFailOnlyWhenSelected) === -1)) {
+                    if (acroFail.has(modeElement.data('modeName')) &&
+                        (inRange || !acroFailOnlyWhenSelected.has(modeElement.data('modeName')))) {
                         acroEnabled = false;
                     }
                 } else if (inRange) {
                     $('.mode .name').eq(modeElement.data('index')).data('modeElement').removeClass('on').addClass('inRange').removeClass('off');
 
-                    if (jQuery.inArray(modeElement.data('modeName'), acroFail) !== -1) {
+                    if (acroFail.has(modeElement.data('modeName'))) {
                         acroEnabled = false;
                     }
                 } else {
--- a/tabs/auxiliary.js
+++ b/tabs/auxiliary.js
@@ -396,6 +396,9 @@ auxiliaryTab.initialize = function (callback) {
             let hasUsedMode = false;
             let acroEnabled = true;
             let acroFail = ["ANGLE", "HORIZON", "MANUAL", "ANGLE HOLD", "NAV RTH", "NAV POSHOLD", "NAV CRUISE", "NAV COURSE HOLD", "NAV WP", "GCS NAV"];
+            // The flight controller reports these modes as active for as long as the Configurator
+            // is connected, so only the selected channel range can make them block ACRO.
+            let acroFailOnlyWhenSelected = ["MANUAL"];
 
             var auxChannelCount = FC.RC.active_channels - 4;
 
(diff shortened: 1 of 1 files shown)


sensei-hacker:
The only way I can reproduce this is the flip the aircraft upside down so that the runtime calibration never finishes. In which case it is in fact in manual mode -- Acro doesn't engage until the runtime calibration completes, normally within a few seconds of powering the board.
As long as my board is sitting relatively flat, Acro works fine for me?

Raffi1202:
Not orientation. An uncalibrated compass latches it: fc_core.c:784-789 forces MANUAL_MODE on a fixed wing while `!ARMING_FLAG(ARMED) && areSensorsCalibrating()`, and compass.c:448-460 keeps COMPASS_CALIBRATED disabled for as long as magZero is all zero and magGain is all 1024. Both screenshots in #2723 are a LUCID_H7_WING whose status bar reads ARMING_DISABLED_SENSORS_CALIBRATING and ARMING_DISABLED_COMPASS_NOT_CALIBRATED. That is read off the code, I have not flashed it.

The indicator bug itself does not depend on why the bit is set: auxiliary.js:410-413 skips range-less modes with a `continue` before the mode-bit test at :416, so identical FC state lights ACRO with no MANUAL range and greys it with one.

Two things are wrong with the PR as it stands. Its stated cause - "the flight controller reports these modes as active for as long as the Configurator is connected" - has no basis; fc_core.c:784-793 has no MSP dependency, and that sentence goes out with the rework. And :418 still adds `on` to the MANUAL row while `acroEnabled` stays true, so MANUAL and ACRO end up lit together. Moving the acroFail test above the `continue` is the opposite fix and matches what the FC is actually doing.

Which direction do you want, and should this target maintenance-9.x? :410/:416/:420 are identical there and the reporter is on Configurator 9.1.1.

sensei-hacker:
Yes, having the aircraft upside-down was my quick way of preventing the runtime calibration from succeeding. 
As you stated, in the user screenshot the message clearly states sensors are not calibrated, so Acro will not engage. Regardless of the reason that calibration hasn't completed.

sensei-hacker:
*I looked at this with an AI-based tool I have. It may well be wrong on any of the points below — please treat it as a set of questions to check rather than a verdict, and correct me if I've misread something.*

@Raffi1202 — I agree with your comment, and I think the direction you outline ("matches what the FC is actually doing") is the right one. A couple of confirmations plus a concrete proposal below.

**The MANUAL bit is genuine, not an MSP/connection artifact.** As you noted, `fc_core.c` forces `MANUAL_MODE` while `!ARMED && areSensorsCalibrating()`, and `packBoxModeFlags()` (`fc_msp_box.c:416`) reports `FLIGHT_MODE(MANUAL_MODE)` over `MSP_ACTIVEBOXES` → `FC.CONFIG.mode` → `isModeBitSet()`. So Configurator's `on` for MANUAL reflects real FC state, and the premise in the current diff's comment ("the flight controller reports these modes as active for as long as the Configurator is connected") isn't correct — I'd suggest dropping it.

**Net effect to aim for** (unarmed + still calibrating): MANUAL = light blue `on`, the requested-but-denied mode (e.g. ANGLE) = dark blue `inRange`, and the ACRO indicator = gray `off`. That's the honest picture, and it only needs the mode bit to be authoritative.

**Proposed change** — keeps the `Set` conversion and the in-range channel/marker guards (both look good; they address the Qodo "unassigned range" point), and makes three edits:

1. Drop `acroFailOnlyWhenSelected` and its comment.
2. Evaluate `isModeBitSet` once up front and apply `on` + ACRO blocking before the unused-mode `continue`, so a range-less-but-forced mode still lights up and still blocks ACRO.
3. Make the `inRange`/`off` branch conditional on `!modeActive`.

```js
for (var i = 0; i < LOCAL_AUX_CONFIG.length; i++) {
    var modeElement = $('#mode-' + i);
    let inRange = false;
    const modeActive = FC.isModeBitSet(modeElement.data('origId'));

    // The FC can engage a mode even when no range is configured — e.g. MANUAL
    // is forced while unarmed and sensors are still calibrating (fc_core.c).
    // Apply the "on" state before the unused-mode skip.
    if (modeActive) {
        $('.mode .name').eq(modeElement.data('index')).data('modeElement').addClass('on').removeClass('inRange').removeClass('off');

        if (acroFail.has(modeElement.data('modeName'))) {
            acroEnabled = false;
        }
    }

    if (modeElement.find(' .range').length == 0) {
        if (!modeActive) {
            modeElement.removeClass('off').removeClass('on');
        }
        continue;
    }

    // Check to see if the mode is in range
    var modeRanges = modeElement.find(' .range');
    for (let r = 0; r < modeRanges.length; r++) {
        const channel = Number.parseInt($(modeRanges[r]).find('.channel').val(), 10);
        if (!Number.isInteger(channel) || channel < 0 || channel >= auxChannelCount) {
            continue;
        }
        var rangeLow = $(modeRanges[r]).find('.lowerLimitValue').html();
        var rangeHigh = $(modeRanges[r]).find('.upperLimitValue').html();
        var markerPosition = $(modeRanges[r]).find('.marker')[0].style.left;
        markerPosition = Number.parseFloat(markerPosition);
        if (!Number.isFinite(markerPosition)) {
            continue;
        }

        rangeLow = (rangeLow - 900) / (2100 - 900) * 100;
        rangeHigh = (rangeHigh - 900) / (2100 - 900) * 100;

        if ((markerPosition >= rangeLow) && (markerPosition <= rangeHigh)) {
            inRange = true;
        }
    }

    if (!modeActive) {
        if (inRange) {
            $('.mode .name').eq(modeElement.data('index')).data('modeElement').removeClass('on').addClass('inRange').removeClass('off');

            if (acroFail.has(modeElement.data('modeName'))) {
                acroEnabled = false;
            }
        } else {
            $('.mode .name').eq(modeElement.data('index')).data('modeElement').removeClass('on').removeClass('inRange').addClass('off');
        }
    }
    hasUsedMode = true;
}
```

**Targeting:** +1 for `maintenance-9.x` — the relevant lines match there, and the reporter is on 9.1.1.

Does this match what you had in mind?

Issue #2639, opened 2026-05-25
OSD elements orphaned when underlying hardware set to NONE; configurator hides them but firmware keeps rendering


jeremykoerber:
## Summary

When `pitot_hardware` is changed from a non-NONE value (e.g. VIRTUAL) to NONE while a pitot-dependent OSD element (Air Speed) is enabled, the element becomes orphaned: the configurator hides it from both the toggle list and the preview canvas, but the firmware continues rendering it on the live OSD output. There is no GUI path to disable the element once it's hidden.

## Environment

- Configurator 9.0.1
- Firmware 9.0.1 / COREWINGF405WINGV2 (commit d44f2cf6)
- DJI O4 Lite Air Unit + DJI Goggles 3, `osd_video_system = DJI_NATIVE`
- Fixed wing, no physical pitot

## Reproduction

1. `set pitot_hardware = VIRTUAL`, save
2. OSD tab: enable Air Speed, save
3. `set pitot_hardware = NONE`, save
4. OSD tab: Air Speed is no longer present in the toggle list (not filterable, not findable) and is not drawn in the preview canvas
5. Live OSD on the goggles continues to render Air Speed, displaying `0` with the warning indicator

## Possible fix ideas

1. Render guard in firmware: skip OSD elements whose required hardware is NONE, irrespective of enable state.
2. Configurator: auto-disable hardware-dependent elements on hardware transition to NONE.
3. Configurator: keep the element visible but greyed out when its hardware is NONE, with the toggle still operable.

## Scope

Likely not pitot-specific. Likely affects any OSD element with a disable-able hardware dependency (ESC telemetry, rangefinder, current sensor, etc.).

b14ckyy:
Long lasting issue and definitely a QoL thing to look into. 

Probably the easiest way to approach this from the configurator side to include a check, that disables OSD elements that are hidden in the UI already. No need to put the same change in the firmware and waste code there.

b14ckyy:
fixed in #2779

PR #2759, opened 2026-09-10
Disable OSD elements the tab can no longer show

Raffi1202:
## Problem

With `pitot_hardware = VIRTUAL` and the Air Speed element enabled, setting `pitot_hardware = NONE` makes the Configurator drop Air Speed from the element list and from the preview, while the flight controller keeps drawing it on the goggles as `0` with the warning indicator, and no GUI path is left to switch it off. Reported on Configurator 9.0.1 / firmware 9.0.1, and not pitot-specific: any element behind a switchable hardware or feature gate can end up the same way. Fixes #2639.

## Cause

`OSD.is_item_displayed()` hides an element as soon as its group or item `enabled()` gate returns false — tabs/osd.js:2465 and tabs/osd.js:2471 on maintenance-10.x — so it leaves the list and the preview. The save handler at tabs/osd.js:3707 only calls `Settings.saveInputs(save_to_eeprom)` and never writes the layout, so the element's visible bit stays set on the FC.

## Change

`is_item_displayed()` is split: `is_item_supported()` takes the three checks that do not depend on hardware (id unknown to the FC, field disabled in the build, `min_version` too new), and `is_item_displayed()` keeps the gate checks on top. `get_unreachable_items()` returns the elements of the selected layout that are still visible and hidden only because a gate closed, skipping any id that is still reachable through another open group. On save those elements get their visible bit cleared and are written back with `MSP2_INAV_OSD_SET_LAYOUT_ITEM` on the layout captured before the writes start, keeping their x/y, and `GUI.log` names them. A refused or failed write aborts the save before EEPROM and logs `osdLayoutSaveFailed`.

## Test

Not run on a flight controller or in the app for this revision — no screenshot or log — and no test was added to `tests/*.test.mjs`. Cause verified by reading tabs/osd.js:2465 and tabs/osd.js:3707 on maintenance-10.x. CI is green for head `c9c24a6`: six platform builds plus the SonarCloud gate with 0 new issues — https://github.com/iNavFlight/inav-configurator/actions/runs/34618043694. Those jobs package the app; they do not exercise this path.

## Flash / RAM

## Docs

None needed. The repository's markdown covers fonts, terrain files and developer patterns; no document describes OSD element visibility or the save path.

### DIFF
--- a/locale/en/messages.json
+++ b/locale/en/messages.json
@@ -6788,6 +6788,9 @@
     "osdClearLayout": {
         "message": "Layout has been cleared"
     },
+    "osdLayoutSaveFailed": {
+        "message": "The flight controller did not accept all OSD layout changes. Settings were not saved to permanent storage. Check the connection and reload the OSD tab before trying again."
+    },
     "osdElementsWithoutHardwareDisabled": {
         "message": "Switched off elements without hardware or feature: $1. They keep their position and come back switched off when the hardware is enabled again."
     },
--- a/locale/en/messages.json
+++ b/locale/en/messages.json
@@ -6788,6 +6788,9 @@
     "osdClearLayout": {
         "message": "Layout has been cleared"
     },
+    "osdElementsWithoutHardwareDisabled": {
+        "message": "Switched off elements without hardware or feature: $1. They keep their position and come back switched off when the hardware is enabled again."
+    },
     "failedToOpenSerialPort": {
         "message": "<span style=\"color: red\">Failed</span> to open serial port"
     },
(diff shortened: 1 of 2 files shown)


sensei-hacker:
If I take the time to set up an element on my OSD, I'm not sure I want it automatically removed "permanently" just because values aren't available *at the moment*. 
I'm thinking perhaps the page should show elements which are either 1) available to enable or 2) already enabled. So pilots can choose to turn them off if they want to. 

For discussion purposes, here's what that would look like:
https://github.com/iNavFlight/inav-configurator/pull/2779

PR #2770, opened 2026-09-14
Spektrum Smart ESC (SRXL2) support in Ports and Outputs

MrScothh:
Configurator side of iNavFlight/inav#11947, which adds Spektrum Smart ESC (SRXL2)
motor output. Draft for the same reason: not yet run against a physical ESC.

Without the firmware PR none of this appears — every control is behind the SRXL2
protocol being selected.

### Ports tab

`Spektrum Smart ESC (SRXL2)` becomes an assignable peripheral function. It is
deliberately not unique: the protocol carries one ESC per bus, so a model with
several motors needs a port each, assigned in motor order.

Its rate selector uses the existing `lockedBaud`, and adds `negotiatedBaud`, which
shows **auto** rather than a number. A locked rate is usually still a real number —
CRSF_SENSOR runs at 420000 and says so — but 115200 is only where an SRXL2 link
starts before the handshake negotiates upwards, so showing it would be a setting
that is not merely locked but wrong. The substitute option carries the stored rate
as its value and changes only the text, so the saved configuration is unchanged.

The lock is now also applied when the tab loads, not only when the function is
changed. It was not before, so a locked selector came up editable until you touched
it — that affects CRSF_SENSOR today, independently of this PR. It is a separate
function from `updateDefaultBaud()` because that one also rewrites the rate and must
not run over a saved configuration.

Saving a port assigned to Spektrum Smart ESC now also sets the motor protocol.
Assigning the port and choosing the protocol were two settings, and doing only the
first leaves a board that does nothing at all: the port is never opened and the
motor never driven. That combination is never wanted, so the protocol follows the
port, logged rather than done quietly. This needs the tab to load the advanced
configuration, which it did not before — writing `MSP_SET_ADVANCED_CONFIG` from a
copy never read would overwrite the gyro denominators and PWM rates with zeroes.

### Outputs tab

An SRXL2 block, shown only when that protocol is selected:

* telemetry on or off, and how often to ask for it — the reply shares the throttle
  wire, so the right answer depends on whether the RPM filter is in use
* a thrust reverse switch and the ESC channel it uses, which is **not** a
  transmitter channel: it selects a slot on the SRXL2 wire and must match the ESC's
  own `Thrust Rev.` parameter
* the throttle-range calibration wizard
* how many SRXL2 ports the board actually opened, and for how many motors

Both those counts come from the firmware rather than being inferred. The obvious
place to look, `MIXER_CONFIG.numberOfMotors`, is a trap: `MSP2_INAV_MIXER`'s last two
bytes are `MAX_SUPPORTED_MOTORS` and `MAX_SUPPORTED_SERVOS`, so that field reads 12
on any board — a single-motor model was told it had twelve motors and needed eleven
more ports.

Reversible motors is hidden when the protocol is SRXL2. A Smart ESC reverses on a
switch and reads the throttle normally, so the centre-zero arrangement does not
apply; the firmware clears the feature for this protocol, so the tab was offering a
control the board would undo. Reverse is the `THRUST REVERSE` mode instead.

### Modes tab

`THRUST REVERSE`, under Fixed Wing. Registered at box id 63, permanent id 72, to
clear AUTO SPEED, TERRAIN AGL HOLD and IN FLIGHT MENU which arrived on
maintenance-10.x after this work started. Serial function id 29, leaving 28 to the
MassZero thermal camera which already claims it here.

### Notes

The wizard, the port counting and the refusal messages were all exercised against a
SITL build of the firmware branch. Two of the fixes above came out of that: the
wizard treated its MSP callback as success and announced "connect the battery" for a
sequence the firmware had refused, and the port count read a field that does not
exist on the parsed port objects.

### Update, 17 September: the review's four findings

**Motor status now lives in shared state.** `MSP2_INAV_ESC_SRXL2_STATUS` had no FC
property and no parser, so every reply reached the unknown-command default while
the tab read phase, refusal reason and port counts straight out of the buffer.
There is now an `FC.SRXL2_STATUS` filled in by `MSPHelper`, and the tab reads it.

**A board without the driver no longer offers the protocol.** Protocol 7 and the
`ESC_SRXL2` port function were listed unconditionally. On firmware built without
`USE_MOTOR_SRXL2`, saving either leaves the board on a protocol nothing drives
and the motors unwritten. That firmware answers the status request as an
unsupported command, which is what Ports and Outputs now ask before building
their menus.

**A dropped request no longer breaks calibration.** `MSP.send_message` calls back
with `false` when a request never arrives, and both callbacks dereferenced
`.data` regardless, so a poller running every 500 ms threw on every tick. Each
use of a reply is checked now; a lost poll is skipped and the next one decides.

**Leaving the tab calls the sequence off.** The poller was a bare `setInterval`
that `cleanup()` neither cleared nor knew about, so changing tabs during
calibration left the firmware holding full throttle waiting for a battery, with
the Abort button gone and a stale closure still writing into removed content. It
is registered with the interval helper the tab lifecycle clears, and `cleanup()`
sends the calibration off on the way out.

The firmware side of this pair has since been run against a real ESC rather than
only against SITL: the driver drives the motor, reverses it while it is turning,
and its calibration moves the ESC's endpoint, all measured. iNavFlight/inav#11947
carries the numbers.


MrScothh:
The two Smart ESC command codes here were one place behind the firmware side.

`MSP2_INAV_ESC_SRXL2_STATUS` and `MSP2_INAV_ESC_SRXL2_CALIBRATE` sat at 0x2232 and 0x2233, while iNavFlight/inav#11947 moved them to 0x2233 and 0x2234 when 0x2232 went to `MSP2_INAV_MAG_UNALIGNED`. Only the number travels on the wire, so with both branches applied the tab asked 0x2232 for the ESC status, the firmware answered it as an unsupported command, and `MSPHelper` read that as `SRXL2_STATUS.supported = false`. The tab then never enabled itself, with nothing on screen to say why. Calibration would have been worse: it would have sent what the firmware now reads as a status request.

Fixed by following the firmware. Same messages, same payloads, two numbers.

Worth noting for anyone reading this later: 0x2232 is no longer free, so this is not just a mismatch between the two sides but a collision with the Alignment Tool's unrotated compass reading.

sensei-hacker:
Thanks for your work on this!
I am assembling INAV 10.0.0RC1 today. After today, there can't be any more new features added until 10.1.
Bug fixes can continue (on a new PR) until December.
So I am going to merge this now and if something else comes up it can be addressed in a new PR.

PR #2771, opened 2026-09-15
Rewrite compass/board alignment wizard around raw unaligned mag data

sensei-hacker:
## Summary
Rewrites the "Alignment Tool" tab's automatic board+compass alignment wizard, which was broken (dead THREE.js matrix code, incorrect yaw math, silently wrote wrong results on boards with a preset alignment already configured). The board-alignment step now does a direct search over the 16 canonical BOARD_ALIGNMENT mounts instead of extracting ambiguous Euler angles from a rotation matrix. The compass-alignment step is rewritten around a new firmware MSP message, `MSP2_INAV_MAG_UNALIGNED` (see companion firmware PR https://github.com/iNavFlight/inav/pull/11708), which reports compass data with zero/gain calibration applied but no alignment rotation — sidestepping the old "which rotation is already baked in depends on preset-vs-custom mode" problem entirely.

On RAM-capable boards, the wizard now hands off to the Calibration tab's own auto-detection (added in the companion firmware PR) after board alignment, with a manual-wizard fallback if detection isn't confident. On RAM-constrained boards it still does the full combined flow in one pass.

## Changes
- Board-alignment step: direct search over the 16 supported mounts (`findBestBoardAlignment`) using two accelerometer readings (flat + ~45° nose-up), replacing a Euler-angle extraction that was ambiguous at 180°/0°.
- Compass-alignment step: derives mounting yaw and flip (upside-down) from `MSP2_INAV_MAG_UNALIGNED` readings taken facing north and east, with a disagreement check if the two estimates don't agree (asks the user to retry).
- Extracted all the pure math (rotation matrices, board-alignment search, compass yaw/flip solver) into `js/boardAlignmentMath.js`, with a 56-case `node --test` suite (`tests/board-alignment-math.test.mjs`) covering known-mount round-trips, the compass yaw/flip solver (including the exact hardware-validated case below), and edge/rejection paths.
- Fixed a real UI bug: the "Save and Reboot" button had two independent click handlers (one `saveChainer.execute()`, one `modal.close()`) relying on jQuery firing order, which sometimes left the modal visibly open across the reboot/reconnect cycle.
- Fixed an MSP poll leak: the wizard's 40ms poll now stops on every completion/error path instead of running until a tab switch.
- Various fixes carried over from code review: `.catch()` on a previously-unhandled promise in the Calibration tab, dead code removal, a `rad2degrees` negative-output edge case, an over-broad jQuery selector, and a mismatched modal illustration.

## Testing
Extensive hardware validation on a live AEDROXH7 across several sessions:
- Board-alignment wizard: verified correct results starting from multiple existing board orientations, including a case that previously round-tripped to a wrong answer.
- Compass wizard: found and fixed two real sign bugs during hardware bring-up (re-derived directly from firmware's `rotationMatrixFromAngles`/`rotationMatrixRotateVector`), then re-validated:
  - Idempotency: repeat runs from scratch converge on the same result.
  - Physical mount changes: rotating the compass 90° and flipping it upside-down both produce correct, independent results.
  - Non-zero FC/board orientation: full wizard runs with the board itself set to yaw 90°, yaw 270°, and upside-down, combined with multiple compass mount orientations at each — all correct.
  - The north/east disagreement error path triggers correctly on a genuine bad-turn test.
- UI: confirmed the Save-and-Reboot modal now closes immediately; confirmed the corrected "face north" modal image renders without a scrollbar.
- `node --test tests/board-alignment-math.test.mjs` — 56/56 passing.
- Manual smoke test in the dev server after the final refactor (module extraction): tab loads, wizard launches, no console errors, verified no accidental writes to the FC.

## Code Review
Reviewed with inav-code-review agent (multiple passes across the branch's history) — no critical issues open; all IMPORTANT/MINOR findings from the most recent pass are fixed or tracked as intentional.

## Related
Depends on firmware PR https://github.com/iNavFlight/inav/pull/11708 for `MSP2_INAV_MAG_UNALIGNED` and the RAM-capability MSP_BOARD_INFO bit.

https://claude.ai/code/session_01BUNPAbSF46z9QMPqTVWjv4


sensei-hacker:
Addressed all four Qodo findings in 7ff440b5ec:

1. **Wizard accepts uncalibrated compasses** — real bug (it actually threw, since `FC.CALIBRATION_DATA` is `null` until requested, rather than silently passing, but broken either way on a RAM-constrained board opened without first visiting Calibration). Fixed: `getMagnetometerCalibrated()` is now null-safe, and the wizard's load chain requests `MSP_CALIBRATION_DATA` itself.
2. **Compass wizard saves an invented alignment** — real bug: `magnetometerUnaligned` inits to `[0,0,0]`, which would read as a perfectly self-consistent heading of 0 if no real MSP response had arrived yet. Fixed: tracks whether a genuine `MSP2_INAV_MAG_UNALIGNED` response was received, shows a new error modal otherwise.
3. **Successful detection can appear to fail** — real, known limitation (firmware doesn't expose a confidence flag over MSP, only the before/after setting diff). Reworded the modal to cover both cases honestly rather than declaring failure; a full fix needs a firmware-side confidence flag, noted in a comment.
4. **Calibration result modal remains over another tab** — real bug, fixed by closing it in `calibrationTab.cleanup()`.

Re-tested on hardware (AEDROXH7): confirmed the wizard still runs correctly with these guards in place, and the Calibration tab's cleanup path no longer throws.

sensei-hacker:
Follow-up simplification: the wizard no longer branches by board RAM class.

Previously, RAM-capable boards stopped after board alignment and were sent to the Calibration tab's separate auto-detect step (with a "manual compass wizard" fallback link) to set compass orientation. Since the wizard already asks the pilot to face north for the board-alignment step, adding the "face east" step costs one extra click and lets the whole flow finish in one pass, on every board — so that's now the only path. Boards with no magnetometer still skip straight to "done" as before.

Changes:
- Removed the RAM-capability branch and the `isRamConstrainedTarget()` check gating it.
- Removed the manual-compass-wizard fallback button/handler and the old board-align-only completion screen (`modal-board-align-done`) — superseded by the existing unified completion screen.
- The calibration-data guard at step 1 (checks `FC.getMagnetometerCalibrated()`) is now unconditional rather than RAM-class-gated, closing a bug where it silently never ran on RAM-capable boards.

Verified end-to-end on live AEDROXH7 hardware (RAM-capable): wizard now goes straight from the 45°-tilt step to "face east" instead of forking to the old board-align-done screen, and reports the correct compass alignment on completion. 62/62 unit tests still pass.

PR #2774, opened 2026-09-16
Treat DSHOT150 as digital, which it has been since 2021

MrScothh:
The Outputs tab picks between the digital and the analog idle guidance with `FC.ADVANCED_CONFIG.motorPwmProtocol >= 5`.

That was DSHOT150's value in `motorPwmProtocolTypes_e` until ONESHOT42 was removed from the firmware in July 2021 (iNavFlight/inav@4713d586), which shifted every protocol after it down by one. DSHOT150 has been 4 ever since, so it currently falls into the analog branch:

- it gets the 10% throttle-idle warning meant for standard PWM and oneshot, instead of the 7% one for digital protocols
- the `.hide-for-shot` fields stay visible when they should be hidden

DSHOT300 and DSHOT600 are unaffected, which is presumably why this went unnoticed for four years.

Fixed by naming the boundary rather than just renumbering it, so the next enum change has the reason sitting next to the number.


sensei-hacker:
Thanks!

sensei-hacker (approved):

PR #2778, opened 2026-09-19
GPS tab: offer the constellations the receiver has, and name the receiver

MrScothh:
Companion to iNavFlight/inav#11979, which is where the receiver's answers start. Without it the fields come back as zero and the tab behaves exactly as it does today, so this can go in on its own.

### What happens

The GPS tab offers the same three constellation switches to every receiver, because until now the only thing it knew about the hardware was one byte, the u-blox generation. A NEO-F10N has no GLONASS: the switch is there anyway, and turning it on made the receiver reject the whole configuration. QZSS and NavIC are never mentioned, though the F10 has both. And the title says GPS whatever is plugged in, while the receiver has been reporting its own name all along.

### The changes

**The tab says which receiver it is.** The module name from the version strings goes next to the title, `GPS - NEO-M10N`. Where a receiver does not report one, as an M8 does not, the title falls back to the generation the hardware version gives, `GPS - u-blox M8`. With the name in the title, the detected-hardware line under the preset no longer has anything left to say and goes: it announced the same receiver a second time, under a preset name rather than a receiver name, and the auto-detect its link performed is already a choice in the preset menu above it.

**A Constellations box.** The four majors, SBAS, QZSS and NavIC in one place, each row present according to what the receiver reported. Two rules, deliberately different:

* a **switch** is withdrawn only when the receiver has said it does not have that constellation. Hiding one on a guess would leave someone unable to turn on something they do have, so an unknown mask shows everything, as today.
* an **indicator**, GPS and QZSS, appears only once the receiver has confirmed it. There is nothing to lose by leaving out a line that only reports.

GPS gets a row of its own, checked and disabled, because it is always on and the tab never said so.

**NavIC.** A real switch now that the firmware has the setting, offered only where the receiver named NavIC.

**The constellation budget.** A receiver tracks only so many at once and says how many. When more than that are selected, a line under the switches says which ones will not be used, before saving as well as after, using the same order the firmware drops them in.

That line describes the firmware in iNavFlight/inav#11965, which is what leaves the extra constellations out and keeps the settings as they are. The limit it reads comes from iNavFlight/inav#11979 instead, so between the two landing the line can appear on a flight controller that does not behave that way: there the receiver answers NAK to the whole configuration and INAV resets the three constellation settings to their defaults without sending them, so nothing is configured at all rather than one constellation being dropped. If that ordering matters, this can wait for #11965.

In practice that line is for u-blox 8: a NEO-M8 tracks three major constellations at once, so selecting all four leaves one out. The M10 platform tracks four and the F10 three, and the F10 has no GLONASS to select in the first place.

**Two small ones while in there:** the update rate field carries the rate its generation is rated for as a tooltip, and the SBAS selector moved into the box with the rest of the constellations, with a help tip saying what the choice does.

### Testing

`npm test`. Two new files: `tests/gps-constellation-availability.test.mjs` covers the mask rules and the budget, `tests/gps-tab-markup.test.mjs` checks that every row the tab code reaches for exists in the markup with the ids and the settings it expects, which is what would have caught the rows hiding behind an inline style.

Checked against three SITL flight controllers running at the same time, each with a different emulated receiver attached, so the three tabs could be compared side by side: an M9 reporting GPS, Galileo, BeiDou and GLONASS; an M10 reporting the same four with SBAS and QZSS; an F10 reporting GPS, Galileo and BeiDou with SBAS, QZSS and NavIC, and no GLONASS.

| | M9 | M10 | F10 |
| --- | --- | --- | --- |
| title | GPS | GPS - NEO-M10N | GPS - NEO-F10N |
| switches offered | GAL, BDS, GLO | GAL, BDS, GLO | GAL, BDS |
| rows shown | GPS | GPS, QZSS | GPS, QZSS, NavIC |
| SBAS selector | hidden | shown | shown |

Against a flight controller without iNavFlight/inav#11979 every mask is zero, every switch is offered and no indicator row appears, which is the current tab.


MrScothh:
All four were worth acting on. Pushed:

**A switch the receiver does not have stayed checked.** Right, and hidden inputs are serialized like any other, so the setting kept going to the flight controller with nobody able to turn it off. A constellation the receiver has just said it does not have is now cleared as well as withdrawn. It costs the preference if someone moves the flight controller to a receiver that does have it, which seemed the lesser of the two: what is saved now matches what is offered.

**The detected-hardware line never went away.** Fixed. The name can arrive after the version, so the line is toggled rather than shown, and it disappears as soon as the receiver names itself.

**An unknown version read as "Manual Settings detected".** Also right, and my doing: I dropped the `!== 'manual'` guard when I rewrote the condition. Put back. An unrecognised version now leaves the line off rather than announcing a preset that is not a detection.

**Fixed font sizes.** The note I added is now sized relative to the text around it. The 12px on the detection line is not new: it is the inline style that was already on that element in the markup, moved into the stylesheet unchanged, and the rest of this stylesheet is in px, so I left it as it was rather than convert one line of it.

The first and the last two now have tests in `tests/gps-tab-markup.test.mjs`, since both are rules inside the tab's closure that nothing else would catch.

MrScothh:
The last one is gone too: the detection line no longer sets a size at all. It inherits, which on the current base is the same 12px the inline style used to force, and it follows the base if that ever moves.

